PR #19, opened 2015-01-02
Add root-keys command

lmars:
The client will need the root keys when performing the initial update, so this change helps the maintainer distribute those keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,11 @@ Recreates the `targets` manifest based on the files in `repository/targets`.
 
 Removes all staged manifests and targets.
 
+#### `tuf root-keys`
+
+Outputs a JSON serialized array of root keys to STDOUT. The resulting JSON
+should be distributed to clients for performing initial updates.
+
 For a list of supported commands, run `tuf help` from the command line.
 
 ### Examples
--- a/tuf/root_keys.go
+++ b/tuf/root_keys.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/flynn/go-docopt"
+	"github.com/flynn/go-tuf"
+)
+
+func init() {
+	register("root-keys", cmdRootKeys, `
+usage: tuf root-keys
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -81,6 +81,11 @@ Recreates the `targets` manifest based on the files in `repository/targets`.
 
 Removes all staged manifests and targets.
 
+#### `tuf root-keys`
+
+Outputs a JSON serialized array of root keys to STDOUT. The resulting JSON
+should be distributed to clients for performing initial updates.
```
titanous: This is for verifying the root manifest (if necessary) during *every* update, not just the initial one, right?

tuf/root_keys.go
```diff
@@ -0,0 +1,28 @@
...
+Outputs a JSON serialized array of root keys to STDOUT.
+
+The resulting JSON should be distributed to clients for performing initial updates.
+`)
+}
+
+func cmdRootKeys(args *docopt.Args, repo *tuf.Repo) error {
+	keys, err := repo.RootKeys()
+	if err != nil {
+		log.Fatal(err)
```
titanous: Can we just return the error?

README.md
lmars: So the Python implementation doesn't check signatures when loading local files, so it just requires a local root.json to be present for the initial update (see [here](https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L235-L246)). I was planning a similar thing, just using a non-signed file to start with (e.g. `tuf.NewClient(local, remote).Init([]*data.Key)`) and then using the local root.json in subsequent updates.

I guess your comment is hinting at also verifying local files. This would require shipping root-keys.json and the client then keeping it in sync with any new root.json it downloads (to support revoking root keys for example).

WDYT?

titanous: Conversation on IRC: https://botbot.me/freenode/flynn/2015-01-02/?msg=28663264&page=2

lmars:
@titanous comments addressed

titanous:
LGTM

PR #24, opened 2015-01-06
Implement client.Update()

lmars:
There are a couple of TODOs in here (i.e. handling invalid signatures, adding more tests) but I'll do them in supplemental PRs, I'm keen to get feedback on the implementation so far.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -96,7 +96,7 @@ func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
 		return err
 	}
 
-	if err := c.decodeRoot(rootJSON); err != nil {
+	if err := c.verifyRoot(rootJSON); err != nil {
 		return err
 	}
 
@@ -158,7 +158,7 @@ func (c *Client) Update() (data.Files, error) {
 		if err != nil {
 			return nil, err
 		}
-		if err := c.decodeRoot(rootJSON); err != nil {
+		if err := c.verifyRoot(rootJSON); err != nil {
 			return nil, err
 		}
 		if err := c.local.SetMeta("root.json", rootJSON); err != nil {
@@ -304,16 +304,17 @@ func (c *Client) downloadMeta(name string, m *data.FileMeta) ([]byte, error) {
 	return buf, nil
 }
 
-// decodeRoot decodes and verifies root metadata.
-func (c *Client) decodeRoot(b json.RawMessage) error {
+// verifyRoot verifies root metadata.
+func (c *Client) verifyRoot(b json.RawMessage) error {
 	var minVer int
 	if c.root != nil {
 		minVer = c.root.Version
 	}
-	if err := signed.Unmarshal(b, &data.Root{}, "root", minVer, c.db); err != nil {
+	s := &data.Signed{}
+	if err := json.Unmarshal(b, s); err != nil {
 		return err
 	}
-	return nil
+	return signed.Verify(s, "root", minVer, c.db)
 }
 
 // decodeSnapshot decodes and verifies snapshot metadata, and returns the new
--- a/client/client.go
+++ b/client/client.go
@@ -1,69 +1,376 @@
 package client
 
 import (
+	"bytes"
+	"encoding/json"
 	"errors"
 	"io"
+	"io/ioutil"
 	"time"
 
-	"github.com/flynn/go-tuf"
 	"github.com/flynn/go-tuf/data"
+	"github.com/flynn/go-tuf/keys"
+	"github.com/flynn/go-tuf/signed"
+	"github.com/flynn/go-tuf/util"
 )
 
 var (
-	ErrNotFound   = errors.New("tuf: file not found")
-	ErrLatest     = errors.New("tuf: the current version is the latest")
-	ErrWrongSize  = errors.New("tuf: unexpected file size")
-	ErrNoRootKeys = errors.New("tuf: no root keys found in local meta store")
+	ErrNotFound       = errors.New("tuf: file not found")
+	ErrLatest         = errors.New("tuf: the current version is the latest")
+	ErrWrongSize      = errors.New("tuf: unexpected file size")
+	ErrNoRootKeys     = errors.New("tuf: no root keys found in local meta store")
+	ErrChecksumFailed = errors.New("tuf: checksum failed")
 )
 
+// LocalStore is local storage for downloaded top-level metadata.
+type LocalStore interface {
+	// GetMeta returns top-level metadata from local storage. The keys are
+	// in the form `ROLE.json`, with ROLE being a valid top-level role.
+	GetMeta() (map[string]json.RawMessage, error)
+
+	// SetMeta persists the given top-level metadata in local storage, the
+	// name taking the same format as the keys returned by GetMeta.
+	SetMeta(name string, meta json.RawMessage) error
+}
+
+// RemoteStore downloads top-level metadata and target files from a remote
+// repository.
 type RemoteStore interface {
-	Get(name string, size int64) (io.ReadCloser, error)
+	// Get downloads the given file from remote storage.
+	//
+	// `file` should be relative to the root of remote repository (e.g.
+	// "root.json" or "targets/path/to/target/file.txt")
+	Get(file string) (io.ReadCloser, error)
 }
 
+// Client provides methods for fetching updates from a remote repository and
+// downloading remote target files.
 type Client struct {
-	repo   *tuf.Repo
+	local  LocalStore
 	remote RemoteStore
+
+	// The following four fields represent decoded metatdata from
+	// local storage
+	root      *data.Root
+	targets   *data.Targets
+	snapshot  *data.Snapshot
+	timestamp *data.Timestamp
+
+	// localMeta is the raw metadata from local storage and is used to
+	// check whether remote metadata is present locally
+	localMeta map[string]json.RawMessage
+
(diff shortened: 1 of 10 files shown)


client/client.go
```diff
@@ -1,71 +1,378 @@
...
 var (
-	ErrNotFound   = errors.New("tuf: file not found")
-	ErrLatest     = errors.New("tuf: the current version is the latest")
-	ErrWrongSize  = errors.New("tuf: unexpected file size")
-	ErrNoRootKeys = errors.New("tuf: no root keys found in local meta store")
+	ErrNotFound       = errors.New("tuf: file not found")
+	ErrLatest         = errors.New("tuf: the current version is the latest")
+	ErrWrongSize      = errors.New("tuf: unexpected file size")
+	ErrNoRootKeys     = errors.New("tuf: no root keys found in local meta store")
+	ErrChecksumFailed = errors.New("tuf: checksum failed")
```
titanous: Some of these should be error structs, open an issue?

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+
+// RemoteStore downloads top-level metadata and target files from a remote
+// repository.
 type RemoteStore interface {
-	Get(name string, size int64) (io.ReadCloser, error)
+	// Get downlaods the given file from remote storage.
+	//
+	// `file` should be relative to the root of remote repository (e.g.
+	// "root.json" or "targets/path/to/target/file.txt")
+	Get(file string) (io.ReadCloser, error)
```
titanous: This should pass the size so that the `Content-Length` can be checked if provided.

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+			return nil, ErrMissingRemoteMetadata{name}
+		}
+		return nil, err
+	}
+	defer r.Close()
+
+	// if m is nil (e.g. when downloading timestamp.json, which has unknown
+	// size), just read the entire stream and return it
+	if m == nil {
+		b, err := ioutil.ReadAll(r)
```
titanous: This should be bounded to some sane maximum.

util/util.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha512"
+	"io"
+
+	"github.com/flynn/go-tuf/data"
+)
+
+type BytesReadCloser struct {
```
titanous: This already exists as `ioutil.NopCloser`

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+		remote: remote,
+	}
+}
+
+// Init initializes a local repository.
+//
+// The latest root.json is fetched from remote storage, verified using rootKeys
+// and then saved in local storage. It is expected that rootKeys were securely
+// distributed with the software being updated.
+func (c *Client) Init(rootKeys []*data.Key) error {
```
titanous: We should probably specify the threshold here?

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+	return buf, nil
+}
 
+// decodeRoot decodes and verifies root metadata.
+func (c *Client) decodeRoot(b json.RawMessage) error {
+	var minVer int
+	if c.root != nil {
+		minVer = c.root.Version
+	}
+	if err := signed.Unmarshal(b, &data.Root{}, "root", minVer, c.db); err != nil {
```
titanous: What is this doing, it doesn't seem to save the data? Perhaps it should be called `verifyRoot` or similar? Also you can just `return` the result of the unmarshal.

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+		return data.FileMeta{}, data.FileMeta{}, err
+	}
+	return snapshot.Meta["root.json"], snapshot.Meta["targets.json"], nil
+}
+
+// decodeTargets decodes and verifies targets metadata, and returns updated
+// targets.
+func (c *Client) decodeTargets(b json.RawMessage) (data.Files, error) {
+	var minVer int
+	currTargets := make(data.Files)
```
titanous:
```go
var currTargets data.Files
```

client/client.go
```diff
@@ -1,71 +1,378 @@
...
+			return nil, err
+		}
+	}
+
+	// Save the snapshot.json now it has been processed successfully
+	if err := c.local.SetMeta("snapshot.json", snapshotJSON); err != nil {
+		return nil, err
+	}
+
+	return updatedTargets, nil
```
titanous: This method doesn't appear to assign downloaded metadata into the relevant fields (`Client.timestamp`, etc.) Should it?

util/util.go
```diff
@@ -0,0 +1,42 @@
...
+func (b BytesReadCloser) Close() error {
+	return nil
+}
+
+func FileMetaEqual(actual data.FileMeta, expected data.FileMeta) bool {
+	if actual.Length != expected.Length {
+		return false
+	}
+	for typ, hash := range expected.Hashes {
+		if !hmac.Equal(actual.Hashes[typ], hash) {
```
heavenlyhash: What's the behavior if for example comparing the `actual` file metadata produced by the currently executing process against the `expected` file metadata deserialized from a different version, and the `expected` has hash types the currently running version hasn't implemented?

titanous:
Overall this looks like a great start.

Here are my observations from a quick read-through of the [Python client](https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py). Some of them may already be addressed:

- Some expires fields need to be ignored when loading local metadata to ensure that the client can update if it only has expired metadata on disk.
- The key database needs to be cleared every time we start with a new root manifest.
- If we become aware that updated metadata exists, we should delete any existing metadata of the same type, even if we were unable to download the new metadata

util/util.go
titanous: Should we compare the intersection of hash types and only error if there are none in common?

heavenlyhash: Yeah, that sounds like a reasonable thing to do.

It seems a little heavyweight to add an error return field to indicate no intersection on available hashes, but maybe it's reasonable -- it is basically doing something like protocol compatibility probing, and that's worthy of error handling.

titanous: We can just change the `bool` to an `error`.

titanous:
On the testing side, please make sure we have a bunch of unhappy path test cases, especially for the attack scenarios that TUF is specifically designed to prevent (there are some good examples in the Python test suite).

client/client.go
lmars: Is there a stdlib method to read at most N bytes? Just checking before I implement it.

titanous: `io.LimitReader`

client/client.go
lmars: So because `Update` always starts from local metadata, there is no need to assign downloaded metadata to the `Client` fields during the update, the downloaded JSON just gets set in local storage once decoded / verified.

The only current use of the client metadata fields is for checking the versions when downloading new metadata, so if the fields seem confusing I could replace them with say `localRootVersion`, `localTargetsVersion` etc.

WDYT?

client/client.go
```diff
@@ -1,71 +1,394 @@
...
+		}
+		c.timestamp = timestamp
+	}
+
+	c.localMeta = meta
 	return nil
 }
 
+// maxMetaSize is the maximum number of bytes that will be downloaded when
+// getting remote metadata without knowning it's length.
```
titanous: s/knowning/knowing/

client/client.go
```diff
@@ -1,71 +1,394 @@
...
+	if err != nil {
+		if err == ErrNotFound {
+			return nil, ErrMissingRemoteMetadata{name}
+		}
+		return nil, err
+	}
+	defer r.Close()
+
+	// if m is nil (e.g. when downloading timestamp.json, which has unknown
+	// size), just read the stream (upto a sane maximum) and return it
```
titanous: s/upto/up to/

client/client.go
```diff
@@ -1,71 +1,394 @@
...
+	if m == nil {
+		stream := io.LimitReader(r, maxMetaSize)
+		b, err := ioutil.ReadAll(stream)
```
titanous: No need to save `stream` to an intermediate variable.

client/client.go
titanous: Yeah, I like that approach better.

lmars:
@titanous @heavenlyhash comments addressed.

> Some expires fields need to be ignored when loading local metadata to ensure that the client can update if it only has expired metadata on disk.

Issue opened: #27

> The key database needs to be cleared every time we start with a new root manifest.

The key database is rebuilt from local root metadata at the start of every update, and when new root metadata is downloaded the update process is immediately restarted.

> If we become aware that updated metadata exists, we should delete any existing metadata of the same type, even if we were unable to download the new metadata

Currently we keep the local metadata so we can check versions of new metadata, so if we delete the local metadata before downloading the new one, we can't check the version, do you not think that is an issue?

titanous:
> Currently we keep the local metadata so we can check versions of new metadata, so if we delete the local metadata before downloading the new one, we can't check the version, do you not think that is an issue?

We definitely need to keep the version info around. For now just open an issue to evaluate it in the future, it's not a big deal.

titanous:
> The key database is rebuilt from local root metadata at the start of every update, and when new root metadata is downloaded the update process is immediately restarted.

Great, please make sure that a test for this is added.

lmars:
@titanous comments addressed, I plan to improve the tests in subsequent PRs

titanous:
LGTM

PR #30, opened 2015-01-08
signed: Don't add duplicate signatures

### LINKED ISSUE #20: Don't allow duplicate signatures
There should be no scenario where we save more than one signature for a given key ID.

lmars:
Fixes #20

### DIFF
--- a/signed/sign.go
+++ b/signed/sign.go
@@ -7,11 +7,19 @@ import (
 )
 
 func Sign(s *data.Signed, k *data.Key) {
+	id := k.ID()
+	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
+	for _, sig := range s.Signatures {
+		if sig.KeyID == id {
+			continue
+		}
+		signatures = append(signatures, sig)
+	}
 	priv := [ed25519.PrivateKeySize]byte{}
 	copy(priv[:], k.Value.Private)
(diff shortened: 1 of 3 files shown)


signed/sign.go
```diff
@@ -7,11 +7,19 @@ import (
 )
 
 func Sign(s *data.Signed, k *data.Key) {
+	id := k.ID()
+	signatures := make([]data.Signature, 0, len(s.Signatures))
```
titanous: Should this be `len(s.Signatures)+1`?

lmars: No, because I don't append the new signature to this slice, this is just the existing signatures

titanous: Looks like you append to it on line 21 below.

lmars: I assumed `append` returns a new slice rather than modifying the given slice? If that is not the case, why do we do `slice = append(slice, ...)` rather than just `append(slice, ...)`?

titanous: If there is no excess capacity, `append` returns a new slice. Otherwise it just uses the existing capacity that is greater than the length. The append below currently returns a new slice, but that is not optimal as it's an unnecessary memory allocation.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #34, opened 2015-01-08
Implement Client.Download

lmars:
An early stage PR to get feedback on the approach and for suggestions on appropriate tests to add.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -47,8 +47,9 @@ type Client struct {
 	localSnapshotVer  int
 	localTimestampVer int
 
-	// localTargets is the list of targets from local storage
-	localTargets data.Files
+	// targets is the list of available targets, either from local storage
+	// or from recently downloaded targets metadata
+	targets data.Files
 
 	// localMeta is the raw metadata from local storage and is used to
 	// check whether remote metadata is present locally
@@ -241,7 +242,7 @@ func (c *Client) getLocalMeta() error {
 			return err
 		}
 		c.localTargetsVer = targets.Version
-		c.localTargets = targets.Targets
+		c.targets = targets.Targets
 	}
 
 	if timestampJSON, ok := meta["timestamp.json"]; ok {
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -379,16 +379,31 @@ func (c *Client) Files() data.Files {
...
+	if !ok {
+		return ErrNotFound{name}
+	}
+	r, err := c.remote.Get("targets/"+name, localMeta.Length)
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+	s := io.TeeReader(io.LimitReader(r, localMeta.Length), dest)
+	actual, err := util.GenerateFileMeta(s)
```
titanous: Might want to add a comment that the copy/hash happens in here.

titanous:
We need a `Targets` accessor method that provides the complete list of available targets.

titanous:
Things we should test:

- Target that is too long doesn't get read fully
- Target that is too short errors
- Target with incorrect hash errors

lmars:
@titanous comments addressed

client/client.go
```diff
@@ -379,16 +379,68 @@ func (c *Client) Files() data.Files {
...
-		TUF downloads and verifies the file and then makes the file available to the
-		software update system.
-	*/
 	return nil
 }
+
+// Targets returns the complete list of available targets.
+func (c *Client) Targets() (data.Files, error) {
+	// populate c.localTargets from local storage
+	if err := c.getLocalMeta(); err != nil {
```
titanous: Check if `localTargets` is set before doing this?

client/client_test.go
```diff
@@ -182,3 +209,81 @@ func (s *ClientSuite) TestNewTargets(c *C) {
...
+}
+
+func (s *ClientSuite) TestDownloadTargetTooLong(c *C) {
+	client := s.updatedClient(c)
+	remoteFile := s.remote["targets/foo.txt"]
+	remoteFile.buf = bytes.NewReader([]byte("foo-ooo"))
+	var dest testDestination
+	c.Assert(client.Download("foo.txt", &dest), IsNil)
+	c.Assert(remoteFile.bytesRead, Equals, 3)
+	c.Assert(dest.deleted, Equals, false)
```
titanous: Should this be true?

lmars: So I guess this is a bit of an edge case. Usually the remote file would have a size corresponding to size of the data, which would result in an `ErrWrongSize` error and the stream being closed immediately (see [this added test](https://github.com/flynn/go-tuf/blob/client-download/client/client_test.go#L248-L256)). Here though the remote size is correct, so we just read the correct number of bytes, get the correct data and all is good, does it then matter that the remote data had more bytes?

lmars:
@titanous comments addressed

titanous:
LGTM

client/client_test.go
heavenlyhash: It's hard to think of a scenario where something that odd happens and it's still correct, so I'd agree with rejecting that state.

titanous: If the length is the same and the content hashes correctly, does it matter?

lmars: FYI this is the same behaviour as [the Python implementation](https://github.com/theupdateframework/tuf/blob/develop/tests/test_download.py#L108-L113)

PR #36, opened 2015-01-09
client: Add downloadMetaUnsafe

lmars:
This makes it more explicit that the method is downloading potentially unsafe data.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -29,9 +29,13 @@ type LocalStore interface {
 type RemoteStore interface {
 	// Get downloads the given file from remote storage.
 	//
-	// `file` should be relative to the root of remote repository (e.g.
-	// "root.json" or "targets/path/to/target/file.txt")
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -31,7 +31,9 @@ type RemoteStore interface {
 	//
 	// `file` should be relative to the root of remote repository (e.g.
 	// "root.json" or "targets/path/to/target/file.txt")
-	Get(file string, size int64) (io.ReadCloser, error)
+	//
+	// ErrNotFound should be returned if the given file does not exist.
+	Get(file string) (io.ReadCloser, int64, error)
```
titanous: This is good, though we should make sure that a sentinel value of `-1` for the size signals that the implementation does not know the length.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #39, opened 2015-01-10
client: Handle expired metadata

### LINKED ISSUE #27: Expired metadata
Downloaded metadata should be rejected if it is expired. Metadata from local storage however should not be rejected due to expiry (only signature checked for consistency).

lmars:
Closes #27

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/flynn/go-tuf"
 	"github.com/flynn/go-tuf/data"
@@ -21,6 +22,7 @@ func Test(t *testing.T) { TestingT(t) }
 type ClientSuite struct {
 	store  tuf.LocalStore
 	repo   *tuf.Repo
+	local  LocalStore
 	remote FakeRemoteStore
 }
 
@@ -86,6 +88,14 @@ func (s *ClientSuite) SetUpTest(c *C) {
 	}
 }
 
+func (s *ClientSuite) syncLocal(c *C) {
+	meta, err := s.store.GetMeta()
+	c.Assert(err, IsNil)
+	for k, v := range meta {
+		c.Assert(s.local.SetMeta(k, v), IsNil)
+	}
+}
+
 func (s *ClientSuite) syncRemote(c *C) {
 	meta, err := s.store.GetMeta()
 	c.Assert(err, IsNil)
@@ -109,7 +119,8 @@ func (s *ClientSuite) rootKeys(c *C) []*data.Key {
 }
 
 func (s *ClientSuite) newClient(c *C) *Client {
-	client := NewClient(MemoryLocalStore(), s.remote)
+	s.local = MemoryLocalStore()
+	client := NewClient(s.local, s.remote)
 	c.Assert(client.Init(s.rootKeys(c), 1), IsNil)
 	return client
 }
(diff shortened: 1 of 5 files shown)


client/client_test.go
```diff
@@ -144,6 +155,15 @@ func (s *ClientSuite) TestInitRootTooLarge(c *C) {
 	c.Assert(client.Init(s.rootKeys(c), 0), Equals, ErrMetaTooLarge{"root.json", maxMetaSize + 1})
 }
 
+func (s *ClientSuite) TestInitRootExpired(c *C) {
+	duration := 100 * time.Millisecond
```
titanous: Any reason to wait 100ms? Why not 10ms?

lmars:
So I just realised that if we download a new `root.json` because the local one is expired, and the new `root.json` has completely different keys, then we will no longer be able to verify the other local metadata in `getLocalMeta`, leaving the client unable to update.

This is an edge case as it is likely that clients will get the new `root.json` through an updated `snapshot.json`, so I'm tempted to open an issue rather than blocking this PR, thoughts?

titanous:
Hmm, if the local metadata fails to validate, shouldn't we download new metadata? Feel free to open an issue.

client/client_test.go
lmars: I had TravisCI in the back of my mind, knowing how slow it can be sometimes. Happy to reduce it, tests only take ~1.3s with this delay though

client/client_test.go
```diff
@@ -204,17 +224,111 @@ func (s *ClientSuite) TestNewTargets(c *C) {
 	c.Assert(files, HasLen, 0)
 }
 
+func (s *ClientSuite) TestLocalExpired(c *C) {
+	client := s.newClient(c)
+	duration := 100 * time.Millisecond
+	syncAndWait := func() {
+		s.syncLocal(c)
+		time.Sleep(duration)
```
titanous: Instead of sleeping, I'd rather abstract the `time.Now()` calculation to be mutable, so for production do:

```go
var timeNow = func() time.Time { time.Now() }
```

And in tests you swap out `timeNow` for a `func()` that is ahead, which allows us to not sleep and make the tests that much faster.

lmars:
>  if the local metadata fails to validate, shouldn't we download new metadata?

Yes that sounds right, but we still need to decode it to get the version out to prevent rollback attacks. I'll open an issue as I don't think it should block this change.

lmars:
@titanous comments addressed

client/client_test.go
```diff
@@ -84,6 +87,27 @@ func (s *ClientSuite) SetUpTest(c *C) {
...
+
+// withMetaExpired sets signed.IsExpired throughout the invocation of f so that
+// any metadata marked to expire at s.expiredTime will be expired (this avoids
+// the need to sleep in the tests).
+func (s *ClientSuite) withMetaExpired(f func()) {
+	signed.IsExpired = func(t time.Time) bool {
+		return t.Unix() == s.expiredTime.Unix()
+	}
+	f()
+	signed.IsExpired = func(time.Time) bool { return false }
```
titanous: For consistency, this should also be set in an `init()` function for the test suite, otherwise tests that run after a call to this could behave differently.

lmars: This new function is not needed for subsequent tests, so I've updated it to just restore the original function: 82bd548

client/client_test.go
```diff
@@ -204,17 +239,117 @@ func (s *ClientSuite) TestNewTargets(c *C) {
...
+	c.Assert(s.repo.GenKeyWithExpires("timestamp", s.expiredTime), IsNil)
+	c.Assert(s.repo.Timestamp(), IsNil)
+	s.syncLocal(c)
+
+	// add far expiring root.json to remote storage
+	c.Assert(s.repo.GenKey("timestamp"), IsNil)
+	s.addRemoteTarget(c, "bar.txt")
+	s.syncRemote(c)
+
+	// check the update downloada the non expired remote root.json and
```
titanous: s/downloada/downloads/

titanous:
LGTM

PR #46, opened 2015-01-10
encrypted: Fix go vet complaint about json tagged field

titanous:
encrypted.go:103: struct field encrypted has json tag but is not exported


lmars:
LGTM

Issue #48, opened 2015-01-11
Prevent slow retrieval attacks


lmars:
Slow retrieval attacks should be prevented by requiring that remote data is fetched in a timely manner.

The Python implementation guards against this attack by reading remote data in small chunks, and signalling an error if reading any given chunk takes longer than a specified time (see [here](https://github.com/theupdateframework/tuf/blob/develop/tuf/download.py#L265-L319)).

lmars:
This is largely implemented in [this commit](https://github.com/flynn/go-tuf/commit/8ec663b3a44888d47728dfc4fe094a192b491ca0) (which is on the `prevent-slow-retrieval-attacks` branch).

However, the implemented `timeoutReader` has the side effect that if the given `[]byte` is larger than `chunkSize` and the underlying reader returns partial data, those partial reads will be buffered until enough chunks have been read.

This may not be an issue, but it needs some thought, and probably needs testing against a real (and slow) HTTP server.

titanous:
It looks like `net/http`'s `Request.Body.Close()` is actually dangerous here: golang/go#9662

mnm678:
Slow retrieval attacks were removed from the TUF spec.

PR #49, opened 2015-01-11
Improve client tests

### LINKED ISSUE #28: Test attack scenarios
The client tests need improving to cover the attack scenarios discussed in [section 1.5.2 of the TUF spec](https://github.com/theupdateframework/tuf/blob/v0.9.9/docs/tuf-spec.txt#L121-L155) (there are some good examples in the Python test suite).

lmars:
This covers all TUF attack scenarios except "Slow Retrieval Attacks" (see #48).

Closes #28.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -43,12 +43,12 @@ type Client struct {
 	local  LocalStore
 	remote RemoteStore
 
-	// The following four fields represent the versions of metatdata from
-	// local storage
-	localRootVer      int
-	localTargetsVer   int
-	localSnapshotVer  int
-	localTimestampVer int
+	// The following four fields represent the versions of metatdata either
+	// from local storage or from recently downloaded metadata
+	rootVer      int
+	targetsVer   int
+	snapshotVer  int
+	timestampVer int
 
 	// targets is the list of available targets, either from local storage
 	// or from recently downloaded targets metadata
@@ -75,6 +75,9 @@ func NewClient(local LocalStore, remote RemoteStore) *Client {
 // and threshold, and then saved in local storage. It is expected that rootKeys
 // were securely distributed with the software being updated.
 func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
+	if len(rootKeys) < threshold {
+		return ErrInsufficientKeys
+	}
 	rootJSON, err := c.downloadMetaUnsafe("root.json")
 	if err != nil {
 		return err
@@ -94,7 +97,7 @@ func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
 		return err
 	}
 
-	if err := c.verifyRoot(rootJSON); err != nil {
+	if err := c.decodeRoot(rootJSON); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


repo.go
```diff
@@ -301,6 +301,40 @@ func (r *Repo) Sign(name string) error {
 	return r.local.SetMeta(name, b)
 }
 
+// getKeys returns signing keys from local storage.
+//
+// Only keys contained in the keys db are returned (i.e. local keys which have
+// been revoked are ommitted), except for the root role in which case all local
```
titanous: s/ommitted/omitted/

client/client.go
```diff
@@ -343,29 +352,24 @@ func (c *Client) downloadMeta(name string, m data.FileMeta) ([]byte, error) {
...
-// verifyRoot verifies root metadata.
-func (c *Client) verifyRoot(b json.RawMessage) (err error) {
-	s := &data.Signed{}
-	if err = json.Unmarshal(b, s); err != nil {
-		return
+// decodeRoot decodes and verifies root metadata.
+func (c *Client) decodeRoot(b json.RawMessage) error {
+	root := &data.Root{}
+	if err := signed.Unmarshal(b, root, "root", c.rootVer, c.db); err != nil {
+		return ErrDownloadFailed{"root.json", err}
```
titanous: Is `ErrDownloadFailed` correct here? Perhaps it should have a comment?

client/client.go
```diff
@@ -373,11 +377,8 @@ func (c *Client) decodeSnapshot(b json.RawMessage) (data.FileMeta, data.FileMeta
 // returns updated targets.
 func (c *Client) decodeTargets(b json.RawMessage) (data.Files, error) {
 	targets := &data.Targets{}
-	if err := signed.Unmarshal(b, targets, "targets", c.localTargetsVer, c.db); err != nil {
-		if err == signed.ErrExpired {
-			err = ErrExpiredMeta{"targets.json"}
-		}
-		return nil, err
+	if err := signed.Unmarshal(b, targets, "targets", c.targetsVer, c.db); err != nil {
+		return nil, ErrDownloadFailed{"targets.json", err}
```
titanous: Ditto on the error.

client/client_test.go
```diff
@@ -397,15 +543,90 @@ func (s *ClientSuite) TestUpdateRemoteExpired(c *C) {
...
+	s.addRemoteTarget(c, "bar.txt")
+	newTargets, ok := s.remote["targets.json"]
+	if !ok {
+		c.Fatal("missing remote targets.json")
+	}
+	s.remote["targets.json"] = oldTargets
+
+	// check update returns ErrWrongSize for targets.json
+	_, err := client.Update()
+	c.Assert(err, DeepEquals, ErrWrongSize{"targets.json", oldTargets.size, newTargets.size})
```
titanous: Can we also do a test where the size is the same?

titanous:
I noticed one code path in the coverage profile that looks like it deserves a test:

```go
// if the latest root.json has been downloaded and it
// is still expired, halt the update with an error
if latestRoot {
        return nil, ErrDownloadFailed{"root.json", err}
}
```

It's probably worth going through the coverage profile and making an issue with a checklist of other code paths that look like they should be tested.

lmars:
@titanous comments addressed.

The code path you found to be untested is actually unreachable with the current code (see [this comment](https://github.com/flynn/go-tuf/blob/improve-client-tests/client/client.go#L124-L126)). I've decided to leave it as a returned error (rather than say `panic("unreached")`) as code changes in the future could make it reachable.

titanous:
LGTM, nice work!

Issue #26, opened 2015-01-07
Interop testing


titanous:
We should have test fixtures generated by the Python implementation to ensure compatibility with our client, as well as a test harness that uses the Python client to test interoperability with repos generated by our code.

PR #51, opened 2015-01-13
Test compatibility with Python generated metadata

lmars:
This is the first half of #26.


util/util.go
```diff
@@ -58,14 +70,41 @@ func FileMetaEqual(actual data.FileMeta, expected data.FileMeta) error {
...
+	m := data.FileMeta{Length: n, Hashes: make(map[string]data.HexBytes, len(hashes))}
+	for hashFunc, h := range hashes {
+		m.Hashes[hashFunc] = h.Sum(nil)
+	}
+	return m, nil
+}
+
+func NormalizeTarget(path string) string {
+	s := filepath.Clean(path)
+	if s[0] == '/' {
```
titanous: It can't hurt to make sure the length is greater than zero before doing this.

data/types.go
```diff
@@ -104,6 +104,14 @@ type FileMeta struct {
 	Custom map[string]interface{} `json:"custom,omitempty"`
 }
 
+func (f FileMeta) HashFuncs() []string {
```
titanous: I'd prefer calling this `HashAlgorithms`

client/client.go
```diff
@@ -447,14 +447,14 @@ func (c *Client) Download(name string, dest Destination) (err error) {
...
-	localMeta, ok := c.targets[name]
+	// return ErrUnknownTarget if the file is not in the local targets.json
+	localMeta, ok := c.targets[util.NormalizeTarget(name)]
 	if !ok {
 		return ErrUnknownTarget{name}
 	}
 
 	// get the data from remote storage
-	r, size, err := c.remote.Get("targets/" + name)
+	r, size, err := c.remote.Get("targets" + name)
```
titanous: Use `path.Join`?

client/client_test.go
```diff
@@ -713,7 +722,7 @@ func (s *ClientSuite) TestDownloadTargetCorruptData(c *C) {
 	remoteFile := s.remote["targets/foo.txt"]
 	remoteFile.buf = bytes.NewReader([]byte("corrupt"))
 	var dest testDestination
-	assertWrongHash(c, client.Download("foo.txt", &dest))
+	assertWrongHash(c, client.Download("/foo.txt", &dest))
```
titanous: Perhaps `Download` should prepend the slash if it doesn't exist?

lmars:
@titanous comments addressed.

`Download` does normalize the given file name, but given the manifest uses "absolute" paths, I think the client should do the same when requesting files.

titanous:
LGTM

PR #53, opened 2015-01-17
Consistent snapshots

### LINKED ISSUE #8: Consistent snapshots
The CLI should default to using consistent snapshots.

lmars:
The idea here is that repositories will default to using consistent snapshots, and can be disabled by running `tuf init --consistent-snapshot=false` before adding any target files.

If target files have already been added, the only way to change whether consistent snapshots are used is to run (the yet unimplemented) `tuf regenerate --consistent-snapshot=false`.

Closes #8.


client/remote_store.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if res.StatusCode == http.StatusNotFound {
+		res.Body.Close()
+		return nil, 0, ErrNotFound{path}
+	} else if res.StatusCode != http.StatusOK {
+		res.Body.Close()
+		return nil, 0, fmt.Errorf("unexpected HTTP response code: %d", res.StatusCode)
+	}
+
+	size, err := strconv.ParseInt(res.Header.Get("Content-Length"), 10, 0)
```
titanous: We can't assume that `Content-Length` will be present, since the response might be chunked.

repo.go
```diff
@@ -470,6 +484,34 @@ func (r *Repo) TimestampWithExpires(expires time.Time) error {
...
+	addHashes := func(name string, meta data.Files) {
+		if m, ok := meta[name]; ok {
+			hashes[name] = m.Hashes
+		}
+	}
+	timestamp, err := r.timestamp()
+	if err != nil {
+		return nil, err
+	}
+	snapshot, err := r.snapshot()
```
titanous: Now that we're doing multiple lookups, does it make sense to memoize these accessors?

tuf/init.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"github.com/flynn/go-docopt"
+	"github.com/flynn/go-tuf"
+)
+
+func init() {
+	register("init", cmdInit, `
+usage: tuf init [--consistent-snapshot=false]
```
titanous: Should this default to `true`? Despite the docs I expect people will call it accidentally/redundantly.

titanous:
Are we doing interop tests for both consistent hashing and not?

client/remote_store.go
lmars: Thats why I return `-1` below to indicate that fact, or is that logic flawed?

repo.go
lmars: So the issue here is the memoized versions getting out of date. For example if data gets signed or thresholds change, the correct invalidation needs to happen. It is possible but I find it less complex to just decode these each time. WDYT?

client/remote_store.go
titanous: Looks right, must have missed that the first time.

tuf/init.go
lmars: The default will be true, the only way it will be false is if the flag is explicitly set to `false`

repo.go
titanous: Okay, that's fine.

lmars:
@titanous interop tests for consistent snapshots added

client/remote_store.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if res.StatusCode == http.StatusNotFound {
+		res.Body.Close()
+		return nil, 0, ErrNotFound{s}
+	} else if res.StatusCode != http.StatusOK {
+		res.Body.Close()
+		return nil, 0, fmt.Errorf("unexpected HTTP response code: %d", res.StatusCode)
```
titanous: Return a [`url.Error`](https://golang.org/pkg/net/url/#Error) here.

titanous:
LGTM pending my last comment.

PR #54, opened 2015-01-19
client: Remove PYTHONPATH from the env before running interop test

titanous:
If a PYTHONPATH already exists in the environment, it will be used instead of the one we specify, breaking the tests.


lmars:
LGTM

PR #55, opened 2015-01-19
data: Change FileMeta.Custom from map[string]interface{} to json.RawMessage

titanous:
This allows consumers of the type to decode into any type instead of being stuck with a `map[string]interface{}`.


lmars:
LGTM (the arguments to `AddTargets` etc. need changing too, but that can be done when support is added to the CLI - #56)

PR #58, opened 2015-01-22
client: Handle updating root with keys that do not match the local meta

titanous:
Closes #39

Version downgrade attack during this scenario is not tested, as it is very unlikely (it would require intentionally signing a metadata file with the wrong version, which makes no sense). The version decode step was verified manually, as it would be very tricky to test (if you have a simple way of testing, suggestions welcome).


lmars:
LGTM

PR #61, opened 2015-01-24
client: Add file local storage

lmars:


titanous:
The downside to doing this here as opposed to in the flynn tree is that we'll have to vendor boltdb twice (we need to add Godeps to this repo, btw).

lmars:
Yes I contemplated adding this to Flynn, but I think it should be straight forward to use a client file store when this library is used standalone.

titanous:
LGTM.

PR #65, opened 2015-01-28
repo: Ensure target paths are always normalized

lmars:
When adding all staged targets, `WalkStagedTargets` calls the `targetsFn` with a path relative to the staged targets dir, so we need to normalize the path.


titanous:
LGTM

PR #66, opened 2015-01-28
tuf-cli: Add Client CLI wrapper

lmars:
This adds three client CLI commands:

* `tuf-cli init` - initialize a local file store using root keys (e.g. from the output of `tuf root-keys`)
* `tuf-cli list` - list available targets and their file size
* `tuf-cli get` - get a target file and write to STDOUT


tuf-cli/get.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	dest := tmpFile{tmp}
+	if err := client.Download(target, &dest); err != nil {
+		return err
+	}
+	defer tmp.Close()
+	if _, err := tmp.Seek(0, os.SEEK_SET); err != nil {
+		return err
+	}
+	_, err = io.Copy(os.Stdout, tmp)
```
titanous: Does it make more sense to download to a filename with the option to specify `-` and send to stdout?

lmars: I usually prefer commands to be filters by default. It can always be piped to a file on the command line if needed, and it is less code. WDYT?

titanous: Sounds good to me. The tempfile needs to be deleted after we're done reading it.

tuf-cli/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"log"
+
+	"github.com/flynn/go-docopt"
+	tuf "github.com/flynn/go-tuf/client"
+)
+
+func main() {
+	log.SetFlags(0)
+
+	usage := `usage: tuf-cli [-h|--help] <command> [<args>...]
```
titanous: Maybe call it `tuf-client`? Could put it in `cmd/tuf-client` and `tuf` in `cmd/tuf`.

lmars:
@titanous comments addressed

titanous:
LGTM

PR #69, opened 2015-02-03
repo,cmd/tuf: Support removing multiple targets at once

lmars:


repo.go
```diff
@@ -415,10 +415,18 @@ func (r *Repo) AddTargetsWithExpires(paths []string, custom map[string]interface
...
+
+func (r *Repo) RemoveTargets(paths []string) error {
+	return r.RemoveTargetsWithExpires(paths, data.DefaultExpires("targets"))
 }
 
 func (r *Repo) RemoveTargetWithExpires(path string, expires time.Time) error {
+	return r.RemoveTargetsWithExpires([]string{path}, expires)
+}
+
+func (r *Repo) RemoveTargetsWithExpires(paths []string, expires time.Time) error {
```
titanous: Should add a comment that a zero-length slice causes the removal of all targets.

titanous: Does it make more sense to have a `SetTargets` method?

lmars: Ok I added a comment, but I'm not sure how a `SetTargets` method relates? Are you proposing something like `tuf set <paths>...`?

titanous: `SetTargets(nil)` makes more sense than `RemoveTargets(nil)` when thought of as an API instead of a direct mapping to a CLI command, however this is fine for now.

titanous:
LGTM

PR #77, opened 2015-03-14
client: Add default targets path

titanous:
There is no situation where an empty targets path makes sense, so it should default to `"targets"`.


lmars:
LGTM

PR #80, opened 2015-03-25
util: Use path instead of filepath in HashedPaths

titanous:
filepath is incorrect here, as we should always use forward slashes, even on Windows.


lmars:
LGTM

PR #81, opened 2015-03-26
repo: Prevent cleaning a new repository

### LINKED ISSUE #78: `tuf clean` can corrupt repo
If I run `tuf clean` before the first `tuf commit`, then `root.json` will be deleted and there is no way to re-create it.

lmars:
After creating keys in a new repository, `repo.Clean()` would remove the staged `root.json`, rendering the keys unusable.

Given `Clean` is only useful to reset the state to what is committed, it only makes sense to call it when something has been committed already, so this change prevents cleaning a new repository.

Fixes #78.

### DIFF
--- a/cmd/tuf/clean.go
+++ b/cmd/tuf/clean.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"fmt"
+	"os"
+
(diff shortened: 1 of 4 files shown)


titanous:
LGTM

PR #84, opened 2015-05-06
Pluggable verifiers

endophage:
There needs to be an update to the keys.Key type to fully support other algorithms. Happy to make those changes if you can provide an opinion. On my fork I just generically treat all keys and signatures as `[]byte` and allow the creators/verifiers to worry about precisely what the correct sizes are and copy the slices into arrays.

Signed-off-by: David Lawrence <david.lawrence@docker.com> (github: endophage)

### DIFF
--- a/signed/verifiers.go
+++ b/signed/verifiers.go
@@ -1,0 +1,72 @@
+package signed
+
+import (
+	"crypto"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+
+	"github.com/flynn/go-tuf/Godeps/_workspace/src/github.com/agl/ed25519"
+)
+
+// Verifier describes the verification interface. Implement this interface
+// to add additional verifiers to go-tuf.
+type Verifier interface {
+	// Verify takes a key, message and signature, all as byte slices,
+	// and determines whether the signature is valid for the given
+	// key and message.
+	Verify(key []byte, msg []byte, sig []byte) error
+}
+
+// Verifiers is used to map algorithm names to Verifier instances.
+var Verifiers = map[string]Verifier{
+	"ed25519": Ed25519Verifier{},
+	//"rsa":     RSAVerifier{},
+}
+
+// RegisterVerifier provides a convenience function for init() functions
+// to register additional verifiers or replace existing ones.
+func RegisterVerifier(name string, v Verifier) {
+	Verifiers[name] = v
+}
+
+// Ed25519Verifier is an implementation of a Verifier that verifies ed25519 signatures
+type Ed25519Verifier struct{}
+
+func (v Ed25519Verifier) Verify(key []byte, msg []byte, sig []byte) error {
+	var sigBytes [ed25519.SignatureSize]byte
+	if len(sig) != len(sigBytes) {
+		return ErrInvalid
+	}
+	copy(sigBytes[:], sig)
+
+	var keyBytes [ed25519.PublicKeySize]byte
+	copy(keyBytes[:], key)
+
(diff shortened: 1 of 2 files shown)


signed/verifiers.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/flynn/go-tuf/Godeps/_workspace/src/github.com/agl/ed25519"
+)
+
+type Verifier interface {
+	Verify(key []byte, msg []byte, sig []byte) error
+}
+
+// Verifiers serves as a map of all verifiers available on the system and
+// can be injected into a verificationService. For testing and configuration
+// purposes, it will not be used by default.
```
titanous: Is the last sentence accurate?

signed/verifiers.go
```diff
@@ -0,0 +1,89 @@
...
+
+// RegisterVerifier provides a convenience function for init() functions
+// to register additional verifiers or replace existing ones.
+func RegisterVerifier(name string, v Verifier) {
+	curr, ok := Verifiers[name]
+	if ok {
+		typOld := reflect.TypeOf(curr)
+		typNew := reflect.TypeOf(v)
+		log.Printf(
+			"Replacing already loaded verifier %s:%s with %s:%s\n",
```
titanous: I'm not a fan of the log lines here and below, what do you think about removing them all?

signed/verifiers.go
endophage: Ah no. It was at one point, then I reverted to a simpler implementation. Will fix.

signed/verifiers.go
endophage: No problem. Docker is mostly using logrus which lets us control log level, so I've been moving all our random bits of logging to that but I didn't want to add that dependency for your fork. I'm happy to just take it all out.

endophage:
One thought, the Ed25519 verifier is covered by existing tests. I'll get tests for the RSA verifier eventually but we can take it out for now and leave additional verifiers to the whim of other users. Let me know if you want to do that. I can also just add a comment for now marking the RSA verifier as "untested, use at your own risk" (it does work though, we used it in a demo)

titanous:
Yeah, add a comment to the effect of "RSA should only be used for compatibility, this verifier is not currently covered by tests, use at your own risk if you really know what you are doing"

signed/verifiers.go
```diff
@@ -0,0 +1,69 @@
...
+	"crypto"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+
+	"github.com/flynn/go-tuf/Godeps/_workspace/src/github.com/agl/ed25519"
+)
+
+// The Verifier interface provides flexibility over exactly how a verifier
+// is implemented vs having a simple function type.
```
titanous: This comment should describe what a `Verifier` does instead of comparing implementation options.

signed/verifiers.go
```diff
@@ -0,0 +1,69 @@
...
+	//"rsa":     RSAVerifier{},
+}
+
+// RegisterVerifier provides a convenience function for init() functions
+// to register additional verifiers or replace existing ones.
+func RegisterVerifier(name string, v Verifier) {
+	Verifiers[name] = v
+}
+
+// Ed25519Verifier is an implementation of a Verifier that verifys ed25519 signatures
```
titanous: s/verifys/verifies

titanous:
Implementation LGTM. Update the comments and squash, and I'll merge. Thanks!

signed/verifiers.go
endophage: Ugh, shoot me now!

PR #89, opened 2015-07-26
Don't reset custom metadata when not specified

lmars:
If I do the following:

```
tuf add --custom '{"foo":"bar"}' foo.txt
tuf add bar.txt
tuf add
```

I expect foo.txt to keep the metadata I give it, but that doesn't currently happen as `tuf add` will set metadata of all staged files to nil.

I have updated the logic to maintain custom metadata if the user hasn't specified any.


titanous:
LGTM.

PR #90, opened 2015-12-01
Ensure we copy boltdb values

titanous:
These slices are backed by a mmap’ed memory area and can cause segfaults if not copied out.

/cc @benbjohnson


jvatic:
LGTM

benbjohnson:
:+1:

PR #91, opened 2015-12-30
S3 remote store implementation

grantr:
This implements an S3 remote store, allowing remote targets and metadata to live in a private S3 bucket instead of an http server.

If a url is given with the `s3` scheme, the `S3RemoteStore` implementation is automatically used. The url must be of the form `s3://<bucket>`.

AWS credentials must be set using the same environment variables and configuration files as awscli, see http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment.

AWS region must be set using the `AWS_REGION` environment variable. Note this is different than awscli's `AWS_DEFAULT_REGION`; this is a quirk of the [aws-sdk-go](https://github.com/aws/aws-sdk-go) library.

I didn't see any tests for `HTTPRemoteStore`, so I didn't write any for this one. I'm not a Go expert so I'll leave that to someone else for now.

I also wasn't able to make godeps behave, so the aws import paths in `S3RemoteStore` are not workspaced. Hopefully someone can help me with that. :smiley:


titanous:
@grantr Sorry for the delay in looking at this.

Do you use this code in production with the CLI?

The reason I ask is because I'm not sure that we want to vendor an S3 client into the repo if possible. Since `RemoteStore` is an interface, you could create a separate package/repo which could then be imported and used with the client. The only downside is that it would not work with the demo/test CLI that we have here.

grantr:
@titanous I don't use this code at all, it's just a PoC. if there are packaging concerns we can close this. No problem :smile:

titanous:
Cool, feel free to reopen if you end up using anything along these lines.

PR #92, opened 2016-02-10
client: Always initialize keys DB from local storage

lmars:
If the local root is expired, an update will download the latest root from remote storage, and we need to be able to verify that new root with the local keys.

We should probably add a test to stop this regressing, but PRing without a test for feedback.

/cc @titanous

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -267,22 +267,21 @@ func (c *Client) getLocalMeta() error {
 		if err := json.Unmarshal(s.Signed, root); err != nil {
 			return err
 		}
-		db := keys.NewDB()
+		c.db = keys.NewDB()
 		for id, k := range root.Keys {
-			if err := db.AddKey(id, k); err != nil {
+			if err := c.db.AddKey(id, k); err != nil {
 				return err
 			}
 		}
 		for name, role := range root.Roles {
-			if err := db.AddRole(name, role); err != nil {
+			if err := c.db.AddRole(name, role); err != nil {
 				return err
 			}
 		}
-		if err := signed.Verify(s, "root", 0, db); err != nil {
+		if err := signed.Verify(s, "root", 0, c.db); err != nil {
 			return err
 		}
 		c.consistentSnapshot = root.ConsistentSnapshot
-		c.db = db
 	} else {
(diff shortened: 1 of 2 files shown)


titanous:
Sounds good, add a atest.

vladimir-v-diaz:
It might be good to also rebuild the locally trusted keys DB once you've fully validated the new root metadata.  When keys are revoked or compromised, root is updated to longer trust these keys, but subsequent root metadata may continue to provide signatures generated by these revoked keys to allow outdated clients to update.  Once outdated clients download the new root, they no longer trust keys that have been revoked.

You might already be doing so, but I wanted to mention it just in case.

I wonder if the TUF specification is unclear on how to set the initial trust on the root keys and how they are updated when a key is revoked or compromised.  While working on the PyPI+TUF proposal, the Python reviewers were unaware of this update procedure.  After realizing the misunderstanding, we updated the Python proposal to say:

Package managers like pip need to ship a file named "root.json" with the installation files that users initially download. This file includes information about the keys trusted for certain roles, as well as the root keys themselves. Any new version of "root.json" that clients may download are verified against the root keys that client's initially trust. If a root key is compromised, but a threshold of keys are still secured, the PyPI administrator MUST push a new release that revokes trust in the compromised keys. If a threshold of root keys are compromised, then "root.json" should be updated out-of-band, however the threshold should be chosen so that this is extremely unlikely. The TUF client library does not require manual intervention if root keys are revoked or added: the update process handles the cases where "root.json" has changed.

lmars:
@vladimir-v-diaz 

> It might be good to also rebuild the locally trusted keys DB once you've fully validated the new root metadata

The library already does this. If you look at [Client.Update](https://github.com/flynn/go-tuf/blob/master/client/client.go#L121-L227), it calls `updateWithLatestRoot` on various conditions (e.g. if the local metadata expired, keys have been revoked etc.) which [downloads and persists the new root manifest](https://github.com/flynn/go-tuf/blob/master/client/client.go#L240-L245).

lmars:
@titanous test updated to cover the issue.

titanous:
LGTM

PR #94, opened 2016-06-18
Remove Godeps

titanous:
Non-main packages should not use Godeps, all of our dependencies are quite stable, and the new vendoring system allows importers to control things properly.


lmars:
LGTM

PR #96, opened 2016-08-09
Refactor key handling

titanous:
This is the first step towards supporting ECDSA keys and signing via PKCS11.

- Remove unused RSA signature verifier.
- Use `golang.org/x/crypto/ed25519` package instead of `github.com/agl/ed25519`.
- Don't mix private and public key structs, use a separate struct
  for private keys.
- Only store public keys in database used for signature
  verification.
- Use `crypto.Signer` interface instead of providing private keys
  directly when signing.
- Don't make as many assumptions about using Ed25519 keys
  everywhere.
- Ignore unknown key types when populating the public key database.

I've tested that keys generated with the previous implementation can be decoded and used by this refactored code (there is no change in the serialized data structures).

### DIFF
--- a/local_store.go
+++ b/local_store.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/flynn/go-tuf/data"
 	"github.com/flynn/go-tuf/encrypted"
+	"github.com/flynn/go-tuf/signed"
 	"github.com/flynn/go-tuf/util"
 )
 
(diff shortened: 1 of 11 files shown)


local_store.go
```diff
@@ -66,15 +67,15 @@ func (m *memoryStore) Commit(map[string]json.RawMessage, bool, map[string]data.H
...
+func (m *memoryStore) GetSigningKeys(role string) ([]signed.Signer, error) {
+	return m.signers[role], nil
 }
 
-func (m *memoryStore) SaveKey(role string, key *data.Key) error {
-	if _, ok := m.keys[role]; !ok {
-		m.keys[role] = make([]*data.Key, 0)
+func (m *memoryStore) SavePrivateKey(role string, key *signed.PrivateKey) error {
+	if _, ok := m.signers[role]; !ok {
+		m.signers[role] = make([]signed.Signer, 0)
```
lmars: Is the explicit `make` necessary?

titanous: Nope, it shouldn't be.

lmars:
LGTM

PR #97, opened 2016-08-12
Split sign and verify packages

titanous:
- Move public key validation logic to `signed` package
- Split `signed` package into `sign` and `verify` packages
- Use `Verifiers` to check key validity instead of hard-coding Ed25519.


lmars:
LGTM

PR #98, opened 2016-08-12
ECDSA verification support

titanous:
- verify: Don't assume signature method, use key type

    It is unsafe to use the signature method field, as it could result in a key confusion attack. Use the key type instead.
    Also, don't assume that we're only working with Ed25519.
-  	verify: Add support for `ecdsa-sha2-p256` signatures

### DIFF
--- a/verify/verify_test.go
+++ b/verify/verify_test.go
@@ -1,6 +1,12 @@
 package verify
 
 import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/sha256"
+	"io"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


verify/verify_test.go
```diff
@@ -18,6 +24,31 @@ type VerifySuite struct{}
...
+}
+
+func (s ecdsaSigner) ID() string {
+	return s.PublicData().ID()
+}
+
+func (ecdsaSigner) Type() string {
+	return data.KeyTypeECDSA_SHA2_P256
+}
+
```
lmars: Do these definitions belong in the test file?

lmars:
LGTM

verify/verify_test.go
titanous: Yeah. This patch intentionally implements only verification in the exposed code. This is in preparation for supporting ECDSA signing keys via PKCS11 in external devices like the Yubikey and HSMs. If you can store keys locally, Ed25519 should be used. We're only adding support for ECDSA because the dedicated hardware devices do not support Ed25519.

lmars: Gotcha.

PR #99, opened 2016-08-26
Support retrying HTTP requests for remote server errors

lmars:
This will be used like:

```go
opts := tuf.HTTPRemoteOptions{
       UserAgent:          fmt.Sprintf("flynn-host/%s %s-%s %s", version.String(), runtime.GOOS, runtime.GOARCH, name),
       ServerErrorRetries: tuf.DefaultServerErrorRetries,
}
```

### DIFF
--- a/client/remote_store.go
+++ b/client/remote_store.go
@@ -8,12 +8,24 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"time"
 )
 
 type HTTPRemoteOptions struct {
 	MetadataPath string
 	TargetsPath  string
 	UserAgent    string
+	Retries      *HTTPRemoteRetries
+}
+
+type HTTPRemoteRetries struct {
+	Delay time.Duration
+	Total time.Duration
+}
+
+var DefaultHTTPRetries = &HTTPRemoteRetries{
+	Delay: time.Second,
+	Total: 10 * time.Second,
 }
 
 func HTTPRemoteStore(baseURL string, opts *HTTPRemoteOptions) (RemoteStore, error) {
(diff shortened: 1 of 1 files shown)


client/remote_store.go
```diff
@@ -51,7 +63,17 @@ func (h *httpRemoteStore) get(s string) (io.ReadCloser, int64, error) {
 	if h.opts.UserAgent != "" {
 		req.Header.Set("User-Agent", h.opts.UserAgent)
 	}
-	res, err := http.DefaultClient.Do(req)
+	var res *http.Response
+	if r := h.opts.ServerErrorRetries; r != nil {
+		for start := time.Now(); time.Since(start) < r.Total; time.Sleep(r.Delay) {
+			res, err = http.DefaultClient.Do(req)
+			if err != nil || res.StatusCode < 500 || res.StatusCode > 599 {
```
titanous: Should we retry if there is an `err`?

lmars: Yes that probably makes sense, I'll update.

lmars:
@titanous comments addressed

client/remote_store.go
```diff
@@ -8,12 +8,24 @@ import (
...
 )
 
 type HTTPRemoteOptions struct {
-	MetadataPath string
-	TargetsPath  string
-	UserAgent    string
+	MetadataPath       string
+	TargetsPath        string
+	UserAgent          string
+	ServerErrorRetries *HTTPRemoteRetries
```
titanous: Maybe just call this `Retries`?

titanous:
LGTM

PR #109, opened 2019-04-25
cmd/tuf: Remove term package usage and import

titanous:
This import is way more trouble than it's worth, so just remove the usage, which is very minor.

### DIFF
--- a/cmd/tuf/main.go
+++ b/cmd/tuf/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


cmd/tuf/main.go
```diff
@@ -127,13 +126,6 @@ func getPassphrase(role string, confirm bool) ([]byte, error) {
 		return []byte(pass), nil
 	}
 
-	state, err := term.SaveState(0)
-	if err != nil {
-		return nil, err
-	}
-	term.DisableEcho(0, state)
-	defer term.RestoreTerminal(0, state)
-
```
lmars: Could switch to the `golang.org/x/crypto/ssh/terminal` package that has a [`ReadPassword`](https://godoc.org/golang.org/x/crypto/ssh/terminal#ReadPassword) function?

lmars: Or the seemingly more convenient [`Terminal.ReadPassword`](https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword) that includes the prompt too.

titanous: Oh, nice. Will do.

titanous:
PTAL

PR #111, opened 2019-11-01
encrypted: fix flake in EncryptedSuite.TestTamperedRoundtrip

ComputerDruid:
If we get really unlucky the encrypted bytes already start with 0x0000
so the tampering fails.

Signed-off-by: Dan Johnson <computerdruid@google.com>


titanous:
Thanks!

PR #123, opened 2020-09-29
Remove support for leading path separators in targets

erickt:
In section 4.5 of the TUF spec, it recommends that target paths do not start with a directory seperator. This removes target resolution automatically stripping the leading "/" from the target path.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -312,9 +312,7 @@ func (c *Client) getLocalMeta() error {
 			return err
 		}
 		c.targetsVer = targets.Version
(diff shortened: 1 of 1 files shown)


trishankatdatadog (approved):

shibumi:
The comments mention that we supported this for backwards compatibility. Are we sure we want to remove that part?

trishankatdatadog:
Closing because the new maintainers are working on separate PRs for catching up compatibility with the 1.x.y spec. Thank you very much for your work, Erick!

PR #124, opened 2020-09-29
Implement the TUF 1.0 workflow

erickt:
This migrates go-tuf over to using the TUF 1.0 workflow for updating repositories.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -133,55 +133,135 @@ func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
 //
 // https://github.com/theupdateframework/tuf/blob/v0.9.9/docs/tuf-spec.txt#L714
 func (c *Client) Update() (data.TargetFiles, error) {
-	return c.update(false)
-}
-
-func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
 	// Always start the update using local metadata
 	if err := c.getLocalMeta(); err != nil {
-		if _, ok := err.(verify.ErrExpired); ok {
-			if !latestRoot {
-				return c.updateWithLatestRoot(nil)
-			}
-			// this should not be reached as if the latest root has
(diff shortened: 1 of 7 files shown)


client/client.go
```diff
@@ -340,6 +373,40 @@ func (c *Client) loadTargets(targets data.TargetFiles) {
...
+		return err
+	}
+	root := &data.Root{}
+	if err := json.Unmarshal(s.Signed, root); err != nil {
+		return err
+	}
+	c.db = verify.NewDB()
+	for id, k := range root.Keys {
+		if err := c.db.AddKey(id, k); err != nil {
+			// FIXME(TUF-0.9) Ignore unknown keyids, which can happen
```
shibumi, line 390: Learning question, because I am new to TUF. In the other PR (https://github.com/theupdateframework/go-tuf/pull/123) we remove support for TUF-0.9, but on the same time we are introducing a change specific to TUF-0.9 in this line. I am a little bit worried about consistence here. Can you explain your thoughts about this? :)

erickt: Thanks for the review! This code was actually moved from the `getLocalMeta` function, and so the comment came along with it. However, now that https://github.com/theupdateframework/taps/blob/master/tap12.md is under consideration, I think the safest thing is that we should be ignoring unknown ids in general, in order to give us the flexibility to change the keyid algorithm in the future. We don't have any test coverage in this area, so I'll factor this bit out into a small patch in order to make this easier to review.

shibumi: Sounds good!

zwass:
@erickt what else needs to be done to move this along? Can I help?

zwass:
@titanous @erickt any chance of moving this along?

trishankatdatadog:
Closing because the new maintainers are working on separate PRs for catching up compatibility with the 1.x.y spec. Thank you very much for your work, Erick!

PR #126, opened 2020-10-19
Add rsassa-pss-sha256 verifier

cce:
Based on RSAVerifier removed in theupdateframework/go-tuf#96 and ECDSA support added in theupdateframework/go-tuf#98

This improves compatibility with reference Python TUF and other TUF implementations. go-tuf already supports the other two verifiers (ECDSA, ED25519) mentioned in the TUF spec, so this completes the set.

Happy to make changes, add tests, etc as per PR feedback.

### DIFF
--- a/data/types.go
+++ b/data/types.go
@@ -11,11 +11,13 @@ import (
 )
 
 const (
-	KeyIDLength              = sha256.Size * 2
-	KeyTypeEd25519           = "ed25519"
-	KeyTypeECDSA_SHA2_P256   = "ecdsa-sha2-nistp256"
-	KeySchemeEd25519         = "ed25519"
-	KeySchemeECDSA_SHA2_P256 = "ecdsa-sha2-nistp256"
+	KeyIDLength                = sha256.Size * 2
+	KeyTypeRSASSA_PSS_SHA256   = "rsassa-pss-sha256"
+	KeyTypeEd25519             = "ed25519"
+	KeyTypeECDSA_SHA2_P256     = "ecdsa-sha2-nistp256"
+	KeySchemeRSASSA_PSS_SHA256 = "rsassa-pss-sha256"
+	KeySchemeEd25519           = "ed25519"
+	KeySchemeECDSA_SHA2_P256   = "ecdsa-sha2-nistp256"
 )
 
 var (
--- a/verify/verifiers.go
+++ b/verify/verifiers.go
@@ -1,9 +1,12 @@
 package verify
 
 import (
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"crypto/rsa"
 	"crypto/sha256"
+	"crypto/x509"
 	"encoding/asn1"
 	"math/big"
 
@@ -25,8 +28,9 @@ type Verifier interface {
 
 // Verifiers is used to map key types to Verifier instances.
 var Verifiers = map[string]Verifier{
-	data.KeySchemeEd25519:         ed25519Verifier{},
-	data.KeySchemeECDSA_SHA2_P256: p256Verifier{},
+	data.KeySchemeEd25519:           ed25519Verifier{},
+	data.KeySchemeECDSA_SHA2_P256:   p256Verifier{},
+	data.KeySchemeRSASSA_PSS_SHA256: rsaVerifier{},
 }
 
 type ed25519Verifier struct{}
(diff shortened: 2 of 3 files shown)


cce:
One idea I had for testing this was to use the actual Python TUF test repos for the `python_interop` tests. The Python [unit/integration test repos](https://github.com/theupdateframework/tuf/tree/develop/tests/repository_data) use a mix of rsassa-pss-sha256 and ed25519, but seem have to been modified to use only ed25519 for the `python_interop` tests in go-tuf.

cce:
Any comments/feedback?

trishankatdatadog:
Hi @shibumi and @erickt, could we get some help reviewing here? 🙂

shibumi:
@cce  your code will not work with PKCS1, correct? Maybe we should mention this somewhere in a comment. Despite that it looks good to me :)

cce:
Oh hrmm. I see you are trying each of `Parse{PKCS8,PKCS1,PKIX}` in order in [in-toto-golang `parseKey`](https://github.com/in-toto/in-toto-golang/blob/master/in_toto/keylib.go#L192-L217). I could implement the same parse order in this PR if that works for you.

I was initially just trying to get compatibility for verifying Python TUF where you see lots of "BEGIN PUBLIC KEY" in the examples — but now I am catching up on threads like https://github.com/secure-systems-lab/securesystemslib/issues/251 and #127 (from my colleague @pawel-kedzior-sw) and reading up on the [in-toto roadmap doc](https://github.com/in-toto/in-toto/blob/develop/roadmap-reviews/2021/review_1_december_20.md) about new signing specs and an abstract signing interface for Python.

@shibumi do you think in-toto-golang and go-tuf may end up sharing a library like the Python implementations do? It would be nice to have a signing interface to plug in alternate implementations, e.g. HSMs or AWS KMS asymmetric key APIs, without maintaining a fork or adding the AWS SDK to these projects directly.

cce:
@shibumi OK! I added PKCS8/PKCS1/PKIX as per your feedback and think this is ready now.

shibumi (approved):

cce:
@erickt I believe this is approved and ready to merge.
@shibumi still curious your thoughts on a shared Go library for the signing spec and abstract Signer interface used by the Go TUF and Go in-toto implementations.

shibumi:
@cce I am not sure if in-toto-go and tuf-go share so much common code for signing. If they do, we should definitely use a shared signing library for this. I really like your abstract signer interface implementation, I am just afraid it take us a lot of work to come up with a solution that will work for both projects.

trishankatdatadog:
Yeah, I'm afraid a shared signing library for Go might be too much right now, although we have one in Python for both, but the future [signing-spec](https://github.com/secure-systems-lab/signing-spec) might help with this...

lukpueh:
Here's what I was trying to say yesterday at the [in-toto community meeting](https://hackmd.io/Czv1Si4jQkieo_qByoMkPQ?view#02112021-Meeting) in regards to the here discussed `go-securesystemslib` (based on my experience with python-{[`securesystemslib`](https://github.com/secure-systems-lab/securesystemslib), [`in-toto`](https://github.com/in-toto/in-toto) and [`tuf`](https://github.com/theupdateframework/tuf)}):

It's generally a good idea to unify the implementation of shared aspects of tuf and in-toto, but a separate library does add maintenance overhead and requires more coordination between tuf and in-toto, which the current {python, golang}-{in-toto, tuf} core dev/maintainer team probably does not have the resources for.

Moreover, python-securesystemslib, which seems like the obvious reference for a golang port, currently has a lot of [moving parts](https://github.com/secure-systems-lab/securesystemslib/milestone/1), i.e. cleaning up legacy code and preparing for the adoption of the new [signing-spec](https://github.com/secure-systems-lab/signing-spec), so a verbatim port is not advisable right now. That's why I suggested to defer work on go-securesystemslib until python-securesystemslib has become more stable, which in addition should also free resources on above mentioned core dev/maintainer team that then could go into the port.

Alternatively, and as @SantiagoTorres suggested, go-securesystemslib could become a completely independent implementation of the new signing-spec instead of a python-securesystemslib go port. While this would require less coordination with the core team, it would mean some duplication of efforts regarding design and implementation that already went or are currently going into python-securesystemslib.

cc @joshuagl, @mnm678, @jku

trishankatdatadog:
Update: sorry, we've all been busy, but rumors are that we might get some help soon, stay tuned...

mnm678:
cc @asraa is this superceded by #148?

asraa:
> cc @asraa is this superceded by #148?

Sort of. #148 has a followup that adds the RSA verifier (not pushed yet, but I can draft it). Most of the functionality is the same, this PR had some issues when I tested it against python generated RSA keys (It doesn't assume pem encoding of the public key string)

asraa:
This is now totally superseded by #148.

PR #129, opened 2020-12-23
Ignore unknown keyids

erickt:
In [TAP-12], we're considering removing the requirement that keyids are derived using `hexdigest(sha256(cjson(public_key_data)))`. This patch updates go-tuf to ignore, rather than fail out, if we see a keyid that wasn't derived with this algorithm. This makes us both forward compatible to TUF-1.0, and backwards-compatible with go-tuf's hybrid 0.9/1.0 transition.

Test: Added tests that make sure repos can generate new versions of metadata and preserve unknown keyids, and that clients ignore repos with random string keyids.

[TAP-12]: https://github.com/theupdateframework/taps/blob/master/tap12.md

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -104,6 +104,7 @@ func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
 		return err
 	}
 
+	// create a new key database, and add all the public `rootKeys` to it.
 	c.db = verify.NewDB()
 	rootKeyIDs := make([]string, 0, len(rootKeys))
 	for _, key := range rootKeys {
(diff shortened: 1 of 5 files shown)


trishankatdatadog (commented):
Am I right in saying that this PR covers arbitrary key IDs, but only for keyIDs distributed inside the root role? What about targets roles (i.e., delegations)? Oh wait, IIRC, there is no support for delegations yet, right?

erickt:
@trishankatdatadog - correct, go-tuf doesn't support delegations yet. It also doesn't yet support the TUF-1.0 workflow either,  we need to land https://github.com/theupdateframework/go-tuf/pull/124 for that first :)

trishankatdatadog (approved):
LGTM!

PR #130, opened 2021-01-04
Delegation

SimonMen65:
Happy New Year Everyone!

In this PR I added the delegation functionality to the project, details are basically described in commit messages.

The branch has passed all tests that & already rebased on the latest theupdateframework:master.

There are 198 files being changed but most of them are regenerated testdata by using client/testdata/go-tuf/regenerate-metadata.sh 

The basic idea of adding this functionality is by changing the Target data-struct similar to Root data-struct, which enables Target role to delegate other target roles. 

Readme is updated which contains detailed procedure of add/remove delegations, add/remove target files to delegated target roles.


lukpueh:
Happy New Year to you too, @SimonMen65. Please fix the git history in your PR and let us know when this is ready for review.

SimonMen65:
@lukpueh Can you be more specific about "fix git history"? I'm confused if you mean there are some lab guidelines I didn't follow or something else. Thanks!

lukpueh:
Did you take a look at the commits pane of your PR? Does anything in that commit history look odd to you and/or remind you of something mentioned very prominently in the [git history guideline document](https://github.com/secure-systems-lab/lab-guidelines/blob/master/git-history.md)?

SimonMen65:
@lukpueh I have read the git history guideline and I think it's ready to review. Leave comments if there's anything wrong.

lukpueh:
The TL;DR of the git history guideline document says, ["avoid merge commits in your feature branch"](https://github.com/secure-systems-lab/lab-guidelines/blame/53f1c1283a0d3748e0abc0f77f2505b924e68b73/git-history.md#L19-L22). If you look at your git history in the commits pane of this PR, you can see that there is a merge commit, which is also the reason why most of your commits appear twice, which is what I meant when I asked you if anything looked odd to you.

SimonMen65:
Hmm looks strange, because I didn't use "merge" command but used "rebase" command and now it appears to be a merge.

SimonMen65:
@lukpueh Now I reset to previous versions and it turns out to be wired. In the checks I saw that they collapsed during "format Unix" stage and I have no idea what happened, nor did the raw log had given me any clues. The tests have all passed at my local so it shouldn't be the problem of codes. I've also checked go.sum with theupdateframework:master:go.sum and it seems the same. Hop you can give me a clue of this clash.

SimonMen65:
Problem solved, simply because didn't format the code; already turned on "Format On Save" mode. I think you can start review the code

repo.go
```diff
@@ -877,3 +1048,315 @@ func (r *Repo) timestampFileMeta(name string) (data.TimestampFileMeta, error) {
...
+func (r *Repo) DelegateAddTargetWithExpires(nameJSON string, path string, custom json.RawMessage, expires time.Time) error {
+	if !strings.Contains(nameJSON, ".json") {
+		nameJSON = nameJSON + ".json"
+	}
+	return r.DelegateAddTargetsWithExpires(nameJSON, []string{path}, custom, expires)
+}
+
+//DelegateAddTargetsWithExpires add targets to non-top target role with expire date
+func (r *Repo) DelegateAddTargetsWithExpires(nameJSON string, paths []string, custom json.RawMessage, expires time.Time) error {
+	if !strings.Contains(nameJSON, ".json") {
```
mnm678, line 1257: I'd also make sure the '.json' is at the end of the string, here and elsewhere in this file

repo.go
```diff
@@ -664,6 +772,17 @@ func (r *Repo) SnapshotWithExpires(t CompressionType, expires time.Time) error {
...
 		return err
 	}
+
+	//Expand snapshotManifest to include delegated roles
+	targ, err := r.targets()
+	if err != nil {
+		return err
+	}
+
+	for name := range targ.Roles {
```
mnm678, line 782: What if delegated roles have delegations?

mnm678 (commented):
Great work! I left a couple of comments inline, but I also have some more general comments/questions:
* I'm not sure if this will support nested delegations (for example if targets delegates to role A who further delegates to role B). 
* It might be nice to generate some test metadata that includes delegations, just to make sure everything works as expected

data/types.go
```diff
@@ -208,11 +244,13 @@ func (f TargetFileMeta) HashAlgorithms() []string {
 }
 
 type Targets struct {
```
asraa, line 246: should this also include a Delegations JSON object? It seems like keys are added into a Keys JSON dictionary, rather than nested in a Delegations object like these examples:
https://github.com/theupdateframework/tuf/blob/f4ffb9dbaabfa6e0df22091b7a6ca63f0dec9377/tests/repository_data/repository/metadata/targets.json#L10-L11

trishankatdatadog:
Closing in favor of #175. Thanks for all your hard work, Simon, I think they were useful for that PR!

PR #132, opened 2021-05-21
Add public function to parse root keys

alexey-igrychev:


alexey-igrychev:
I apologize, PR was created by accident

PR #133, opened 2021-06-02
[types] add custom json message to keyval

asraa:
This adds a custom json message to `KeyValue`. By the specification, `KeyValue` must contain the public data associated with the key, and here I'd like to extend it to include information other than the Public bytes associated with the key:

I'd like to include a hardware and device cert for the offline keys generated, and attach their certificates in the KeyValue in the root metadata. It seems appropriate to attach it to the Key. 

WDYT? Happy to make a PR in the python repository as well, the change will actually need to be in securesystemslib https://github.com/secure-systems-lab/securesystemslib/blob/1983feda61da8a9ce8c6d2bc0955c1d2be08ad6f/securesystemslib/formats.py#L179

### DIFF
--- a/data/types.go
+++ b/data/types.go
@@ -61,7 +61,8 @@ func (k *Key) ContainsID(id string) bool {
 }
 
 type KeyValue struct {
-	Public HexBytes `json:"public"`
+	Public HexBytes         `json:"public"`
+	Custom *json.RawMessage `json:"custom,omitempty"`
 }
 
 func DefaultExpires(role string) time.Time {
--- a/data/types_test.go
+++ b/data/types_test.go
@@ -11,9 +11,10 @@ const (
 	//
 	// https://github.com/theupdateframework/specification
 	//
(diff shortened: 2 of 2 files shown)


trishankatdatadog:
I don't see a problem with this, but we do need to [clarify](https://github.com/theupdateframework/specification/issues/163) first in the TUF spec how to deal with unknown/unexpected superset fields. Could you help us there?

jku:
this should just work already for the newer python Metadata API (but a test case would definitely be valued).

The current/legacy python-tuf code does verify input data against the many related schemas... so something probably needs to change to make a current python-tuf client compatible with this.

trishankatdatadog:
> The current/legacy python-tuf code does verify input data against the many related schemas... so something probably needs to change to make a current python-tuf client compatible with this.

This is partly what I mean by we need to clarify the spec to tell implementations what to do...

trishankatdatadog:
Do we have a resolution on this?

asraa:
superseded by https://github.com/theupdateframework/go-tuf/pull/148

Issue #6, opened 2014-12-30
Support modifying role thresholds via the CLI


lmars:
I should be able to change role thresholds using the CLI, for example:

```
$ tuf set-threshold root 2
```

PR #135, opened 2021-06-24
Support modifying role thresholds

asraa:
Adds the ability to modify role thresholds in the CLI

Fixes https://github.com/theupdateframework/go-tuf/issues/6

Question: are there CLI tests? 

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -181,7 +181,7 @@ func (r *Repo) GetThreshold(keyRole string) (int, error) {
 }
 
 func (r *Repo) SetThreshold(keyRole string, t int) error {
-	if !validManifest(keyRole) {
+	if !validManifest(keyRole + ".json") {
 		// Delegations are not currently supported, so return an error if this is not a
 		// top-level manifest.
 		return ErrInvalidRole{keyRole}
--- a/repo.go
+++ b/repo.go
@@ -167,6 +167,40 @@ func (r *Repo) RootVersion() (int, error) {
 	return root.Version, nil
 }
 
+func (r *Repo) GetThreshold(keyRole string) (int, error) {
+	root, err := r.root()
+	if err != nil {
+		return -1, err
+	}
+	role, ok := root.Roles[keyRole]
+	if !ok {
+		return -1, ErrInvalidRole{keyRole}
+	}
+
+	return role.Threshold, nil
+}
+
+func (r *Repo) SetThreshold(keyRole string, t int) error {
+	root, err := r.root()
+	if err != nil {
+		return err
+	}
+	role, ok := root.Roles[keyRole]
+	if !ok {
+		return ErrInvalidRole{keyRole}
+	}
+	if role.Threshold == t {
+		// Change was a no-op.
+		return nil
+	}
+	role.Threshold = t
+	if _, ok := r.versionUpdated["root.json"]; !ok {
+		root.Version++
+		r.versionUpdated["root.json"] = struct{}{}
+	}
+	return r.setMeta("root.json", root)
+}
+
 func (r *Repo) Targets() (data.TargetFiles, error) {
 	targets, err := r.targets()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


repo.go
```diff
@@ -167,6 +167,40 @@ func (r *Repo) RootVersion() (int, error) {
...
+	role, ok := root.Roles[keyRole]
+	if !ok {
+		return ErrInvalidRole{keyRole}
+	}
+	if role.Threshold == t {
+		// Change was a no-op.
+		return nil
+	}
+	role.Threshold = t
+	if _, ok := r.versionUpdated["root.json"]; !ok {
```
mnm678, line 197: Hmm, this may have to change for delegations. If the threshold of a delegated target is updated, it's the parent target that will need a new version.

repo_test.go
```diff
@@ -705,7 +705,7 @@ func (rs *RepoSuite) TestCommitVersions(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(snapshotVersion, Equals, 3)
 
-	timestampVersion, err = r.SnapshotVersion()
+	timestampVersion, err = r.TimestampVersion()
```
mnm678, line 708: Nice catch!

repo.go
```diff
@@ -167,6 +167,40 @@ func (r *Repo) RootVersion() (int, error) {
 	return root.Version, nil
 }
 
+func (r *Repo) GetThreshold(keyRole string) (int, error) {
```
mnm678, line 170: Is this function only used for testing? I have no real objection to that, but in the future it might be nice to be able to check the current thresholds from the cli.

mnm678 (commented):
My go skills are a bit rusty (no pun intended), but this looks good to me. There's a couple of bits that would need to change for delegations, but I don't think that's a blocker.

repo.go
asraa: Hmmmmm that's a good point. I took a look at https://github.com/theupdateframework/go-tuf/pull/130 and I don't even think that properly handles incrementing the parent roles (only targets, maybe not an intermediate).

should i error out if it's not a top-level role right now, or since delegations aren't even supported hold off?

repo.go
asraa: Oh yeah, I'll expose a `get-threshold` in the CLI. I figured it'd be a useful getter in using the Repo implementation

repo.go
mnm678: Yeah, maybe error out for now. Eventually this'll need some kind of knowledge of who the parent of a role is, but that can be left to the delegations pr.

repo.go
```diff
@@ -167,6 +167,45 @@ func (r *Repo) RootVersion() (int, error) {
...
+		return ErrInvalidRole{keyRole}
+	}
+	if role.Threshold == t {
+		// Change was a no-op.
+		return nil
+	}
+	role.Threshold = t
+	if _, ok := r.versionUpdated["root.json"]; !ok {
+		root.Version++
+		r.versionUpdated["root.json"] = struct{}{}
```
trishankatdatadog, line 204: Interesting: what does this line do?

repo_test.go
trishankatdatadog, line 708: Good catch!

trishankatdatadog (approved):
Very nice functionality, thanks! Clearly the code will need to change when we add delegations, but this is very much needed.

repo.go
asraa: Ah so `versionUpdated` is just a map that stores whether the role has been version incremented or not before committing. So if it hasn't yet been incremented (because `r.versionUpdated["root.json"]` entry isn't there), only then will be update and then store the entry in the map. This prevents incrementing versions multiple changes before committing.

Tbh I don't know why this is a map into an empty struct, rather than a boolean..

trishankatdatadog:
I am happy with merging. Marina?

mnm678 (approved):

mnm678:
> I am happy with merging. Marina?

LGTM, but I'm not a maintainer

PR #137, opened 2021-07-12
Support delegations in client

raphaelgavache:
This PR adds the support of delegations in the client

#### Notable changes
- support delegations in client.Download, following spec 1.0.19 section 5.6
- root is no longer required in the snapshot meta

#### Notes
- ignore the commit `Add TUF3 php test`, this commit only imports test files
- copied TUF3 files from php fixtures to test delegations logic

#### Follow ups
- add chained root downloading to comply with spec and add other fixtures from php

##### 1. `data/types.go data/types_test.go` 
-  delegations fields are added to the structs with JSON support
-  `MatchesPath` matches a file to a delegation with the support of Paths and PathHashPrefixes
##### 2. `verify/db.go`
- reuse DB struct verifiers to verify delegated roles from a parent Targets keys
- `delegationsVerifier` is added to bypass specific logic dedicated to the 4 top roles
##### 3. `client/client.go`
- client.Download will now try to fetch FileMeta through the graph of delegations if it's not already cached. Previous pattern was  check cache, check local top level Targets. New pattern is check cache, check local top level Targets, check local delegations, download missing delegations.
- root is no longer required in snapshot
- MaxDelegations parameter is introduced to bound a search in the delegations
##### 4. `client/delegations.go client/delegations_test.go`
- `getTargetFileMeta` implements the delegation search logic


arbll (approved):

trishankatdatadog:
Thanks, @raphaelgavache!

Would you please summarize the changes over the source code, so that we know what to expect while reviewing?

ethan-lowman-dd (commented):
Thanks Raphael. Here's a first batch of comments. I hope to complete a first pass by tomorrow.

raphaelgavache:
Thanks Ethan, the delegation verifier code is now a lot cleaner! 
I'll apply your other comments tomorrow morning and add an error when both Paths and PathsPrefixes are set to stick to the spec

trishankatdatadog:
Thanks for your review, @ethan-lowman-dd!

@raphaelgavache: does this implement a single-pass (i.e., look for only one target at a time) or a multi-pass (i.e., look for multiple targets at once) approach to delegations?

raphaelgavache:
It's the single-pass one that searches for one target at a time

trishankatdatadog:
As a sanity-check, can we cross-check for correctness against [this recursive implementation](https://github.com/theupdateframework/tuf/pull/1463)?

data/types.go
```diff
@@ -208,11 +210,44 @@ func (f TargetFileMeta) HashAlgorithms() []string {
...
+type DelegatedRole struct {
+	Name             string   `json:"name"`
+	KeyIDs           []string `json:"keyids"`
+	Threshold        int      `json:"threshold"`
+	PathHashPrefixes []string `json:"path_hash_prefixes,omitempty"`
+	Paths            []string `json:"paths"`
+	Terminating      bool     `json:"terminating"`
+}
+
+func (d DelegatedRole) MatchesPath(file string) bool {
```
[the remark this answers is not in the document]
raphaelgavache: I updated to allow only one of `paths` or `path_hash_prefixes` at json decoding level https://github.com/theupdateframework/go-tuf/pull/137/commits/8eac49ac1dfc789722d85f5c53731a07a89e012d
This will err when decoding, and enforce in the DelegatedRole struct that both are never set together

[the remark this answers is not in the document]
ethan-lowman-dd: I opened a PR suggesting what I think is a cleaner way that doesn't require multiple structs, copying, or renaming: https://github.com/raphaelgavache/go-tuf/pull/3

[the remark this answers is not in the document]
raphaelgavache: Thank you, I accepted your suggestions

client/delegations.go
```diff
@@ -0,0 +1,201 @@
...
+	for id, _ := range r.KeyIDs {
+		keyIDs = append(keyIDs, id)
+	}
+
+	// root delegates the signing of all files to the top level targets
+	return data.DelegatedRole{
+		Name:             role,
+		KeyIDs:           keyIDs,
+		Threshold:        r.Threshold,
+		PathHashPrefixes: []string{""},
```
ethan-lowman-dd, line 139: Instead of using `PathHashPrefixes` treat the root differently, could we just make a special case for when `role = "targets"`? The behavior in the spec produces is unrelated to hash prefixes.

client/delegations_test.go
```diff
@@ -0,0 +1,408 @@
...
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/verify"
+)
+
+var (
+	defaultPathPatterns = []string{"tmp", "*"}
+	noMatchPathPatterns = []string{"vars", "null"}
+)
+
+func TestDelegationsIterator(t *testing.T) {
```
ethan-lowman-dd, line 24: Let's add some hash prefix tests. I didn't get to writing these tests today, but I tested the client on an example delegated hashed-bins repo, and it worked properly.

data/types.go
```diff
@@ -208,11 +212,79 @@ func (f TargetFileMeta) HashAlgorithms() []string {
...
+}
+
+// DelegatedRole enforces the spec 1.0.19 section 4.5 'role MUST specify only one of the "path_hash_prefixes" or "paths"'
+// by forcing one of path_hash_prefixes, paths  with MatchWithHashPrefixes and PathPatterns fields.
+// delegatedRoleJSON and delegatedRoleCopy help decoding and encoding jsons.
+// DelegatedRole UnmarshalJSON will fail and return ErrPathsAndPathHashesSet if both fields are set and not empty
+type DelegatedRole struct {
+	Name             string   `json:"name"`
+	KeyIDs           []string `json:"keyids"`
+	Threshold        int      `json:"threshold"`
```
ethan-lowman-dd, line 235: Can we make this a `uint`?

ethan-lowman-dd (commented):
See [this branch](https://github.com/ethan-lowman-dd/go-tuf/commits/ethan.lowman/delegations-in-client) for the proposed changes from today.

client/delegations_test.go
raphaelgavache: HashPrefix / path logic is handled by DelegatedRole in data/types_go and is already tested in data/types_test.go `TestDelegatedRolePathMatch`. We shouldn't test the internals of dependencies

data/types.go
raphaelgavache: All existing Threshold are ints, so for consistency I prefer keeping it as a int

client/delegations.go
raphaelgavache: A special case with role = "targets" could be dangerous if the role name is reused by mistake. This part works well and is clean, I don't think we should bother changing it

verify/db.go
```diff
@@ -26,6 +26,36 @@ func NewDB() *DB {
...
+}
+
+func (d *DelegationsVerifier) Unmarshal(b []byte, v interface{}, role string, minVersion int) error {
+	return d.DB.Unmarshal(b, v, role, minVersion)
+}
+
+// NewDelegationsVerifier returns a DelegationsVerifier that verifies delegations
+// of a given Targets. It reuses the DB struct to leverage verified keys, roles
+// unmarshals.
+func NewDelegationsVerifier(d *data.Delegations) (DelegationsVerifier, error) {
```
ethan-lowman-dd, line 40: Does this need to be exported by the `verify` package? Alternatively, could it be private to the client (the only place it's used)? I think all we'd need is a way to make a delegated role pass the validation of `AddRole()`.

raphaelgavache: Yes, we could add an exportable AddDelegatedRole and have everything in the client code. Current implem avoids adding delegated roles to DB and using a shared DB to check delegations. Having a separate struct with only Unmarshal enforces that the DelegatedVerifier is used for delegations: parent verifies childs. When we add Delegations in the repo  we can reuse this verifier to check that they are valid

data/types.go
```diff
@@ -208,11 +212,92 @@ func (f TargetFileMeta) HashAlgorithms() []string {
...
+	Terminating      bool     `json:"terminating"`
+	PathHashPrefixes []string `json:"path_hash_prefixes,omitempty"`
+	Paths            []string `json:"paths"`
+}
+
+// MatchesPath evaluates whether the path patterns or path hash prefixes match
+// a given file. This determines whether a delegated role is responsible for
+// signing and verifying the file.
```
trishankatdatadog, line 243: I think this function should be split into two: one for matching paths, and one for match path hash prefixes. Less likely to make an error that way.

data/types.go
```diff
@@ -208,11 +212,92 @@ func (f TargetFileMeta) HashAlgorithms() []string {
...
+	for _, hashPrefix := range d.PathHashPrefixes {
+		if strings.HasPrefix(pathHash, hashPrefix) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// validateFields enforces the spec 1.0.19 section 4.5:
```
trishankatdatadog, line 265: For this kind of references to the spec, please link directly, like so: https://theupdateframework.github.io/specification/v1.0.19/index.html#file-formats-targets

verify/db.go
```diff
@@ -26,6 +26,36 @@ func NewDB() *DB {
...
+	db := &DB{
+		roles: make(map[string]*Role, len(d.Roles)),
+		keys:  make(map[string]*data.Key),
```
trishankatdatadog, line 43: Why not allocate length of keys here also?

verify/db.go
```diff
@@ -43,22 +73,31 @@ func (db *DB) AddKey(id string, k *data.Key) error {
 	return nil
 }
 
-var validRoles = map[string]struct{}{
+var topLevelRoles = map[string]struct{}{
```
trishankatdatadog, line 76: Hmm. I'm not crazy about go-tuf inheriting the bad design decision from the original python-tuf of keeping keys in a module-wide dictionary, but this is not our doing.

verify/verify.go
```diff
@@ -24,12 +23,24 @@ func (db *DB) Verify(s *data.Signed, role string, minVersion int) error {
 	if err := json.Unmarshal(s.Signed, sm); err != nil {
 		return err
 	}
-	if strings.ToLower(sm.Type) != strings.ToLower(role) {
-		return ErrWrongMetaType
+
+	if isTopLevelRole(role) {
+		// Top-level roles can only sign metadata of the same type (e.g. snapshot
+		// metadata must be signed by the snapshot role).
+		if sm.Type != role {
```
trishankatdatadog, line 30: Why no lowercasing of both sides here?

verify/verify.go
```diff
@@ -24,12 +23,24 @@ func (db *DB) Verify(s *data.Signed, role string, minVersion int) error {
...
+			return ErrWrongMetaType
+		}
+	} else {
+		// Delegated (non-top-level) roles may only sign targets metadata.
+		if sm.Type != "targets" {
```
trishankatdatadog, line 35: Why no lowercasing of left side?

client/delegations.go
```diff
@@ -0,0 +1,213 @@
...
+// getTargetFileMeta searches for a verified TargetFileMeta matching a file name
+// Requires a local snapshot to be loaded and is locked to the snapshot versions.
+// Searches through delegated targets following TUF spec 1.0.19 section 5.6.
+func (c *Client) getTargetFileMeta(file string) (data.TargetFileMeta, error) {
+	snapshot, err := c.loadLocalSnapshot()
+	if err != nil {
+		return data.TargetFileMeta{}, err
+	}
+
+	verifiers := map[string]verify.DelegationsVerifier{
```
trishankatdatadog, line 17: What are the string keys here? Delegated rolenames?

client/delegations.go
```diff
@@ -0,0 +1,213 @@
...
+		"root": {DB: c.db},
```
trishankatdatadog, line 18: Why is "root" here?

client/delegations.go
```diff
@@ -0,0 +1,213 @@
...
+		if err := c.local.SetMeta(fileName, raw); err != nil {
+			return nil, err
+		}
+	}
+	return target, nil
+}
+
+func (c *Client) rootTargetDelegation() data.DelegatedRole {
+	role := "targets"
+	r := c.db.GetRole(role)
```
trishankatdatadog, line 128: What does "r" mean? Seems like a bad variable name.

client/delegations.go
```diff
@@ -0,0 +1,213 @@
...
+	// 5.6.6 persist
+	if !alreadyStored {
+		if err := c.local.SetMeta(fileName, raw); err != nil {
+			return nil, err
+		}
+	}
+	return target, nil
+}
+
+func (c *Client) rootTargetDelegation() data.DelegatedRole {
```
trishankatdatadog, line 126: Don't call it "rootTargetDelegation", please. Very confusing with "root" and "targets". Maybe "topLevelTargetsDelegations"? You get the idea.

client/delegations.go
trishankatdatadog: Yeah, I don't like this sentinel either, seems wrong. Someone who knows how the code works can try to exploit this. If need be, we can start with `paths=["*"]` (which is an implicit, automatic "delegation" to the top-level targets role anyway).

trishankatdatadog (changes requested):
first set of comments

verify/db.go
```diff
@@ -26,6 +26,32 @@ func NewDB() *DB {
 	}
 }
 
+type DelegationsVerifier interface {
+	Unmarshal([]byte, interface{}, string, int) error
```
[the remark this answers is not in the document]
raphaelgavache: This is an outdated commit and that code does not exist in latest version. I've updated PR description to stop pointing to the first commit

client/client.go
```diff
@@ -81,12 +82,17 @@ type Client struct {
 	// consistentSnapshot indicates whether the remote storage is using
 	// consistent snapshots (as specified in root.json)
 	consistentSnapshot bool
+
+	// MaxDelegations limits by default the number of delegations visited for any
+	// target
+	MaxDelegations int
```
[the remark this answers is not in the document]
raphaelgavache: This is a configuration and it bounds [each delegations search](https://github.com/raphaelgavache/go-tuf/blob/c226180061a67312c194abc184a8516aef94d374/client/delegations.go#L24). So it's not mutated by each search

```
c:= NewClient()
c.MaxDelegations = 3
// c.Download() iterates through delegation graph with a max of 3 nodes visited
```

verify/verify.go
```diff
@@ -24,7 +24,7 @@ func (db *DB) Verify(s *data.Signed, role string, minVersion int) error {
 	if err := json.Unmarshal(s.Signed, sm); err != nil {
 		return err
 	}
-	if strings.ToLower(sm.Type) != strings.ToLower(role) {
+	if typ := strings.ToLower(sm.Type); typ != strings.ToLower(role) && typ != "targets" {
```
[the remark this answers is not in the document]
raphaelgavache: Same as https://github.com/theupdateframework/go-tuf/pull/137#discussion_r671110601 this code was removed in further commits

client/delegations.go
raphaelgavache: [I reworked and cleaned up the iteration initialization ](https://github.com/raphaelgavache/go-tuf/pull/5). In that process I removed all references to root. Closing the comments on the removed code.

verify/verify.go
raphaelgavache: good catch, covered here https://github.com/raphaelgavache/go-tuf/pull/5/commits/3078b69258457de0967f5dc136a38a8b1109b62c

client/delegations.go
raphaelgavache: I added comments above to explain the keys https://github.com/theupdateframework/go-tuf/blob/9631aebe2de0c383a399a93f3dcbb0f162d721a9/client/delegations.go#L17-L19
Could you check if they are clear?

data/types.go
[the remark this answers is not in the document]
raphaelgavache: Closing as it is also covered here https://github.com/theupdateframework/go-tuf/pull/137#discussion_r670915358. This one is on outdated code.
`MatchesPath` now returns an error if both paths hash prefixes are set

data/types.go
trishankatdatadog: Ok, I see now. So if you rename `validateFields` to `validatePaths` we are good to go, bcos otherwise not obvious.

client/client.go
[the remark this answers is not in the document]
trishankatdatadog: Got it.

client/delegations.go
```diff
@@ -0,0 +1,202 @@
...
+// Searches through delegated targets following TUF spec 1.0.19 section 5.6.
+func (c *Client) getTargetFileMeta(file string) (data.TargetFileMeta, error) {
+	snapshot, err := c.loadLocalSnapshot()
+	if err != nil {
+		return data.TargetFileMeta{}, err
+	}
+
+	// verifiers is map of parent targets name to an associated DelegationsVerifier
+	// that can verify all child targets pointed by delegatedRoles in the parent targets
+	verifiers := make(map[string]verify.DelegationsVerifier)
```
trishankatdatadog, line 19: Note to self: think about how the same role can be given a different threshold of keys in different delegations.

Example: imagine A delegates to B and C, and both B and C delegate to D. However, B and C use different keys to delegate to D.

Like Ethan said, consider not having any variable that persists across iterations. Easier to understand and not make mistakes.

client/delegations.go
```diff
@@ -0,0 +1,202 @@
...
+	// https://github.com/theupdateframework/specification/issues/168
+	if delegation.child.Terminating {
+		d.stack = d.stack[0:0]
+	}
+	return delegation, true
+}
+
+func (d *delegationsIterator) add(roles []data.DelegatedRole, parent string) error {
+	for i := len(roles) - 1; i >= 0; i-- {
+		// Push the roles onto the stack in reverse so we get an in-order traversal
```
trishankatdatadog, line 189:
```suggestion
		// Push the roles onto the stack in reverse so we get an preorder traversal
```

client/delegations.go
```diff
@@ -0,0 +1,202 @@
...
+	if _, ok := d.visited[id]; ok {
+		return d.next()
+	}
+	d.visited[id] = struct{}{}
+
+	// 5.6.7.2 trim delegations to visit, only the current role and its delegations
+	// will be considered
+	// https://github.com/theupdateframework/specification/issues/168
+	if delegation.child.Terminating {
+		d.stack = d.stack[0:0]
```
ethan-lowman-dd, line 182:
```suggestion
		// Empty the stack.
		d.stack = d.stack[0:0]
```

client/delegations.go
```diff
@@ -0,0 +1,202 @@
...
+			{
+				child: data.DelegatedRole{Name: "targets"},
+			},
+		},
+		visited: make(map[delegationID]struct{}),
+	}
+	return i
+}
+
+func (d *delegationsIterator) next() (delegation, bool) {
```
trishankatdatadog, line 164: Ethan: pls name return arguments.

trishankatdatadog (changes requested):
second round of comments

client/delegations.go
```diff
@@ -0,0 +1,198 @@
...
+		err = c.db.Unmarshal(raw, targets, role, fileMeta.Version)
+	} else {
+		err = verifier.Unmarshal(raw, targets, role, fileMeta.Version)
+	}
+	if err != nil {
+		return nil, ErrDecodeFailed{fileName, err}
+	}
+
+	// 5.6.4 check against snapshot version
+	if targets.Version != fileMeta.Version {
```
trishankatdatadog, line 109: I think this is checked anyway with localMetaFromSnapshot/downloadMetaFromSnapshot, but not a bad idea to check again anyway.

client/delegations.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, ErrTargetsSnapshotVersionMismatch{
+			Role:                     fileName,
+			DownloadedTargetsVersion: fileMeta.Version,
+			TargetsSnapshotVersion:   targets.Version,
+			SnapshotVersion:          snapshot.Version,
+		}
+	}
+	// 5.6.6 persist
+	if !alreadyStored {
+		if err := c.local.SetMeta(fileName, raw); err != nil {
```
trishankatdatadog, line 119: It's possible that a delegation might use a top-level rolename to try overwrite this local storage, but I think the scenario is very unlikely given other checks.

client/delegations.go
```diff
@@ -0,0 +1,198 @@
...
+		return delegation{}, false
+	}
+	delegation := d.stack[len(d.stack)-1]
+	d.stack = d.stack[:len(d.stack)-1]
+
+	// 5.6.7.1: If this role has been visited before, then skip this role (so
+	// that cycles in the delegation graph are avoided).
+	roleName := delegation.delegatee.Name
+	if _, ok := d.visitedRoles[roleName]; ok {
+		return d.next()
```
trishankatdatadog, line 164: Right now, a simple cycle could cause a stack overflow. Maybe consider a more graceful approach?

trishankatdatadog (approved):
LGTM, the code is super nice!

@joshuagl and @asraa can you pls help review? 🙂

client/delegations.go
ethan-lowman-dd: I think this should be fine -- from the beginning of `next()` to `return d.next()`, we pop from the stack but do not push to it, so the recursion depth is bounded.

raphaelgavache: Actually a cycle can't cause a stack overflow as we don't add the delegated roles of a node already seen. I'm still adding a simple a->a cycle check in the test cases. I didn't think clearly during our sync

client/client.go
```diff
@@ -198,7 +204,7 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
 	if err != nil {
 		return nil, err
 	}
-	rootMeta, targetsMeta, err := c.decodeSnapshot(snapshotJSON)
+	snapshotMetas, err := c.decodeSnapshot(snapshotJSON)
```
asraa, line 207: maybe update the comment in line 198 that root.json file meta should not be there but use if there?

client/client.go
```diff
@@ -210,13 +216,17 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	if rootMeta, ok := snapshotMetas["root.json"]; ok {
+		if !c.hasMetaFromSnapshot("root.json", rootMeta) {
+			return c.updateWithLatestRoot(&rootMeta)
+		}
 	}
 
 	// If we don't have the targets.json, download it, determine updated
 	// targets and save targets.json in local storage
 	var updatedTargets data.TargetFiles
+	targetsMeta := snapshotMetas["targets.json"]
```
asraa, line 229: since this is a client downloading possibly bad metadata, does this always have to exist? should it fail early?

It seems like it was always assumed to exist in the previous code too

client/delegations.go
```diff
@@ -0,0 +1,190 @@
...
+			{
+				delegatee: data.DelegatedRole{Name: "targets"},
+			},
+		},
+		visitedRoles: make(map[string]struct{}),
+	}
+	return i
+}
+
+func (d *delegationsIterator) next() (value delegation, ok bool) {
```
asraa, line 145: wouldn't this code enable it so the top-level target is searched first? i think you don't need to check that before you call `getTargetFileMeta`

client/client.go
asraa: maybe i'm confused what's being downloaded unsafely or not...

client/delegations.go
raphaelgavache: You are right, we search a second time the FileMeta in the top-level target from the Download function
Passing through the top-level targets is useful to reuse:
- load top-level targets from local store
- load delegations from top-level targets
- add a verifier for them

It also makes the function getTargetFileMeta complete as it searches through all targets

client/client.go
raphaelgavache: on this part I kept the existing behavior, would you rather have me change it and skip the targets step if no target are found in the validated snapshot?
If no targets are set in the snapshot meta, the current logic will try to download a targets of version 0. It will fail if previous targets were seen on the version check.

client/delegations.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (d *delegationsIterator) next() (delegation, bool) {
+	if len(d.stack) == 0 {
+		return delegation{}, false
+	}
+	delegation := d.stack[len(d.stack)-1]
+	d.stack = d.stack[:len(d.stack)-1]
+
+	// 5.6.7.1 cycles protection
+	id := delegationID{delegation.parent, delegation.child.Name}
```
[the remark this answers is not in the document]
ethan-lowman-dd: I raised an issue on the specification repo to clarify this: https://github.com/theupdateframework/specification/issues/177

client/client.go
asraa: I *think* the existing behavior should be fine based on the tests, so I would say that's good!

client/delegations.go
asraa: Right! definitely agree, I think `getTargetFileMeta` should start with the top-level ones! looking back, i'm confused where the calling site was also doing it... but anyway, thank you!

asraa (approved):

trishankatdatadog:
@raphaelgavache, one more thing: can you check whether you match paths using [glob-style matching](https://github.com/theupdateframework/specification/pull/174)?

Meanwhile, I will read your tests to check for correctness...

data/types.go
```diff
@@ -208,11 +212,93 @@ func (f TargetFileMeta) HashAlgorithms() []string {
...
+// MatchesPath evaluates whether the path patterns or path hash prefixes match
+// a given file. This determines whether a delegated role is responsible for
+// signing and verifying the file.
+func (d *DelegatedRole) MatchesPath(file string) (bool, error) {
+	if err := d.validatePaths(); err != nil {
+		return false, err
+	}
+
+	for _, pattern := range d.Paths {
+		if matched, _ := filepath.Match(pattern, file); matched {
```
hosseinsia, line 250: wrt glob-style matching, the filepath.Match we are using looks conformant.
I checked the golang doc on filepath.Match (https://pkg.go.dev/path/filepath#Match). The separator characters are not matched. For example this filepath.Match("*.txt", "sss/1.txt"") doesn't pass.

trishankatdatadog:
Hmm, looks like @mnm678 and @asraa are not maintainers of go-tuf, which is not cool! We should fix this. I'm going to go ahead and merge this in the interesting of unblocking this PR, which has been through extensive review. Thanks again everyone, especially @raphaelgavache!

raphaelgavache:
Thank you for the reviews!

PR #138, opened 2021-07-20
Add the ability to sign with an external tool (Dump signable content + append signature)

asraa:
Allows users to sign content with an external tool. Required
* Making `SignedMeta` public to access the signable content as the payload to sign
* AppendSignature: adds a signature to the metadata. Checks that this is a valid key for the role and replaces existing sigs with the same ID
* Adding a method to add a verification key, similar to the python reference implementation. This adds the public key data for a role to the root metadata without needing the private key

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -551,7 +551,7 @@ func (r *Repo) AppendSignature(name string, signature data.Signature) error {
 		return ErrInvalidRole{role}
 	}
 	if !roleData.ValidKey(signature.KeyID) {
-		return verify.ErrInvalidKeyID
+		return verify.ErrInvalidKey
 	}
 
 	s, err := r.SignedMeta(name)
--- a/repo.go
+++ b/repo.go
@@ -322,11 +322,6 @@ func (r *Repo) AddPrivateKey(role string, key *sign.PrivateKey) error {
 }
 
 func (r *Repo) AddPrivateKeyWithExpires(keyRole string, key *sign.PrivateKey, expires time.Time) error {
-	root, err := r.root()
-	if err != nil {
-		return err
-	}
-
 	if !verify.ValidRole(keyRole) {
 		return ErrInvalidRole{keyRole}
 	}
@@ -339,6 +334,19 @@ func (r *Repo) AddPrivateKeyWithExpires(keyRole string, key *sign.PrivateKey, ex
 		return err
 	}
 	pk := key.PublicData()
+
+	return r.AddVerificationKeyWithExpiration(keyRole, pk, expires)
+}
+
+func (r *Repo) AddVerificationKey(keyRole string, pk *data.Key) error {
+	return r.AddVerificationKeyWithExpiration(keyRole, pk, data.DefaultExpires(keyRole))
+}
+
+func (r *Repo) AddVerificationKeyWithExpiration(keyRole string, pk *data.Key, expires time.Time) error {
+	root, err := r.root()
+	if err != nil {
+		return err
+	}
 
 	role, ok := root.Roles[keyRole]
 	if !ok {
@@ -502,7 +510,7 @@ func (r *Repo) Sign(name string) error {
 		return ErrInvalidRole{role}
 	}
 
-	s, err := r.signedMeta(name)
+	s, err := r.SignedMeta(name)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


repo.go
```diff
@@ -526,6 +534,50 @@ func (r *Repo) Sign(name string) error {
 	return r.local.SetMeta(name, b)
 }
 
+// AppendSignature allows users to append a signature generated with an external tool.
+func (r *Repo) AppendSignature(name string, signature data.Signature) error {
```
trishankatdatadog, line 538: What is `name` exactly? Only the base, role filename?

repo.go
```diff
@@ -526,6 +534,50 @@ func (r *Repo) Sign(name string) error {
...
+	if !roleData.ValidKey(signature.KeyID) {
+		return verify.ErrInvalidKey
+	}
+
+	s, err := r.SignedMeta(name)
+	if err != nil {
+		return err
+	}
+
+	// Iterate through signatures and remove a signature with a matching ID
```
trishankatdatadog, line 562: I think removing existing signatures for the keyID, and adding a new one, is a good idea, but I'm afraid the name of the function is then a bit misleading. Two choices: either we split adding/appending and updating into two functions, or we give the function a better name (like AddOrUpdateSignature).

repo.go
```diff
@@ -526,6 +534,50 @@ func (r *Repo) Sign(name string) error {
...
+	// Check Key ID is in valid for the role.
+	db, err := r.db()
+	if err != nil {
+		return err
+	}
+	roleData := db.GetRole(role)
+	if roleData == nil {
+		return ErrInvalidRole{role}
+	}
+	if !roleData.ValidKey(signature.KeyID) {
```
trishankatdatadog, line 553: Good idea, but we should also check that the signature is valid for this key for the signed message.

trishankatdatadog (changes requested):

repo.go
asraa: Manifest name! Clarified in the comment. name typically referred to that in this file, although I agree it's not clear...

repo.go
asraa: This is also how `Sign` works! Signing twice over the same manifest with the same key replaces the old one. On the other hand, it's the same to me to skip appending if already present. Would that make sense so it doesn't update?

repo.go
asraa: done, added test

repo.go
trishankatdatadog: What do you think about "roleFilename"?

asraa: Something like that sounds fine! I'll make the change consistent though through the file

repo.go
```diff
@@ -526,6 +534,60 @@ func (r *Repo) Sign(name string) error {
...
+	}
+
+	// Add signature if one doesn't already exist with the same key ID.
+	found := false
+	for _, sig := range s.Signatures {
+		if sig.KeyID == signature.KeyID {
+			found = true
+		}
+	}
+	if !found {
```
trishankatdatadog, line 570: Should we return an error if found? Otherwise, function can "fail" silently, surprising the user.

asraa: of, this is adding the signature if it's not already there. so this is expected actually...

trishankatdatadog: No, no, I understand. I guess, what should we make of the user who appends a new signature w/o checking that there is an old one (which is perhaps obsolete now)? We could punt it to them to check, which is not a wrong answer.

asraa: Hmmm whenever there is a repo change, setMeta will remove existing signatures and re-sign https://github.com/theupdateframework/go-tuf/blob/22a94818d17b266491bd7bedb8f311c7ebced9c5/repo.go#L482

In the case where you aren't storing keys and you are using external tools, you'll have no signatures when you setMeta. So I'm not sure this is possible? It should not. I will also add a test for this case.

(This makes me think we should be overriding though, just in case, for a latest signature)

trishankatdatadog: We just have to make a decision: either error on duplicate signatures, or do not allow duplicate signatures. If the latter, I recommend re-adding the logic, and renaming the function to follow principle of least surprise. I will let you decide 🙂

asraa: Agreed, I think it's better for AddOrUpdate now.

trishankatdatadog:
Please ping me when ready to review again. Thanks!

repo_test.go
```diff
@@ -1430,3 +1432,151 @@ func (rs *RepoSuite) TestThreshold(c *C) {
...
+			Signature: badSig}), Equals, verify.ErrInvalid)
+	}
+
+	// add the correct root signature
+	for _, id := range rootKey.Signer().IDs() {
+		c.Assert(r.AddOrUpdateSignature("root.json", data.Signature{
+			KeyID:     id,
+			Signature: rootSig}), IsNil)
+	}
+	checkSigIDs := func(role string, keyIDs ...string) {
```
trishankatdatadog, line 1565: Not immediately obvious how this function checks for uniqueness of keyIDs... can you pls briefly explain?

asraa: Changed to check against the role data Key IDs, which are a map necessarily providing unique key IDs https://github.com/theupdateframework/go-tuf/blob/aee6270feb5596036edde4b6d7564fa17db811cb/data/types.go#L139

trishankatdatadog (approved):
💯 

Thanks, Asra!

trishankatdatadog:
@shibumi or @mnm678, can you pls help review?

mnm678 (approved):
LGTM

shibumi (approved):

shibumi:
Mh looks like @mnm678's and my approval do not count. @joshuagl  can you have a look and approve this if fine?

trishankatdatadog:
Again, we need to fix these permission issues for @shibumi and @mnm678.

trishankatdatadog:
Will go ahead and merge in the meantime.

PR #143, opened 2021-08-10
Rotate roots per spec

hosseinsia:
Update root following the TUF specification v1.0.19
https://theupdateframework.github.io/specification/v1.0.19/index.html#load-trusted-root

Logic:
1. Call the update root first in any client updates.
2. Start from the current root version (n), loop through next versions by trying to fetch (n+1).root.json file.
3. For any new root, verify the signatures.


joshuagl (commented):
Some drive by comments. I'm not very familiar with go-tuf, so take them with a grain of salt.

This PR is very hard to review because of the number of added files. Do we need them all? Perhaps separate them into a separate commit so that the core logic change is easier to review?

Personally, I'm not a fan of the very large comments which seem likely to become out-of-date. But that appears to be the style for the go-tuf implementation.

client/client.go
```diff
@@ -179,52 +171,35 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
 
 	// Return ErrLatestSnapshot if we already have the latest snapshot.json
-	if c.hasMetaFromTimestamp("snapshot.json", snapshotMeta) {
-		return nil, ErrLatestSnapshot{c.snapshotVer}
-	}
+	//if c.hasMetaFromTimestamp("snapshot.json", snapshotMeta) {
+	// FIXME: Should we return with an error?
+	//	fmt.Println("Hey you!")
+	//	return nil, ErrLatestSnapshot{c.snapshotVer}
+	//}
```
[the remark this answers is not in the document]
hosseinsia: no. debugging remnant. removed.

client/client.go
```diff
@@ -179,52 +171,35 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
 
 	// Get snapshot.json, then extract file metas.
 	//
 	// The snapshot.json is only saved locally after checking
 	// targets.json so that it will be re-downloaded on subsequent updates
 	// if this update fails.
-	// root.json meta should not be stored in the snapshot, if it is,
+	// root.json meta should not be stored in the snapshot(???), if it is,
```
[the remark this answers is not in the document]
hosseinsia: Thanks for confirming. yup, removed.

client/client.go
```diff
@@ -239,6 +214,7 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
 			return nil, err
 		}
 		if err := c.local.SetMeta("targets.json", targetsJSON); err != nil {
+			fmt.Println("problem here 220")
```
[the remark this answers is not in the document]
hosseinsia: old debugging remnant. removed.

client/clienttest.go
```diff
@@ -0,0 +1,126 @@
...
+	return l, err
+}
+
+// Initializes the client object with local files without fetching
+// the latest version from the server.
+func (c *Client) initWithLocal(rootKeys []*data.Key, threshold int, localrootpath string) error {
+	if len(rootKeys) < threshold {
+		return ErrInsufficientKeys
+	}
+	rootJSON, err := ioutil.ReadFile(localrootpath) //c.downloadMetaUnsafe("root.json", defaultRootDownloadLimit)
```
trishankatdatadog, line 30: pls remove comment

client/client_test.go
```diff
@@ -360,6 +362,107 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+}
+
+func (s *ClientSuite) TestUpdateRoots(c *C) {
+	var tests = []struct {
+		fixturePath         string
+		isExpired           bool // Value retuned by verify.IsExpired.
+		extpectedError      error
+		expectedRootVersion int // -1 means no check is performed on this.
+	}{
+		// Good new root update succeeds (the timestamp check disabled).
```
trishankatdatadog, line 423: What timestamp check?

trishankatdatadog (changes requested):
First round of comments. Thanks for the great work, @hosseinsia! 👏🏽 💯

client/client.go
```diff
@@ -247,26 +214,170 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	// then delete the trusted timestamp and snapshot metadata files.
+	getKeyIDs := func(role string) []string {
+		keyIDs := make([]string, 0, len(c.db.GetRole(role).KeyIDs))
+		for k := range c.db.GetRole(role).KeyIDs {
+			keyIDs = append(keyIDs, k)
+		}
+		sort.Strings(keyIDs)
+		return keyIDs
+	}
+	startKeyIDs := map[string][]string{"timestamp": {}, "snapshot": {}, "targets": {}}
```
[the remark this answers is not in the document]
asraa: i would consider factoring these names into "nonRootManifests" maybe?

client/client.go
```diff
@@ -247,26 +214,170 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		}
+
+		for k := range startKeyIDs {
+			if !reflect.DeepEqual(startKeyIDs[k], getKeyIDs(k)) {
+				c.local.SetMeta(k, json.RawMessage{})
+			}
+		}
+
+		// 5.3.4.2 check that N+1 signed itself.
+		//This is different from the previous call because now the threashold and the keys are updated.
```
asraa, line 326: nit: threshold typo

is it really that they have been updated, or that `getLocalRootMeta()` verifies using the updated keys (by creating a NewDB https://github.com/theupdateframework/go-tuf/blob/31841328f398d1235ef6da054162985747f000eb/client/client.go#L450)

client/errors.go
```diff
@@ -133,3 +133,23 @@ type ErrRoleNotInSnapshot struct {
 func (e ErrRoleNotInSnapshot) Error() string {
 	return fmt.Sprintf("tuf: role %s not in snapshot version %d", e.Role, e.SnapshotVersion)
 }
+
+//func (e ErrTargetsSnapshotVersionMismatch) Error() string {
```
asraa, line 137: remove?

client/client.go
```diff
@@ -247,26 +214,170 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+}
+
 func (c *Client) updateWithLatestRoot(m *data.SnapshotFileMeta) (data.TargetFiles, error) {
 	var rootJSON json.RawMessage
 	var err error
-	if m == nil {
-		rootJSON, err = c.downloadMetaUnsafe("root.json", defaultRootDownloadLimit)
+	if m == nil && c.rootVer != 0 {
+		if err = c.updateRoots(); err != nil {
+			if _, ok := err.(ErrEmptyTimestampOrSnapshot); !ok {
```
asraa, line 362: am i missing where this error is ever used? if i am, can you please comment when/where `updateRoots()` may return this?

client/client.go
```diff
@@ -350,6 +427,52 @@ func (c *Client) getLocalMeta() error {
...
+	// keys first
+	s := &data.Signed{}
+	if err := json.Unmarshal(rootJSON, s); err != nil {
+		return err
+	}
+	root := &data.Root{}
+	if err := json.Unmarshal(s.Signed, root); err != nil {
+		return err
+	}
+	c.db = verify.NewDB()
```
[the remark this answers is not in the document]
asraa: Yeah, this honestly  confuses me (but it was pre-existing code):
if you fail with an unrelated error like verify.ErrWrongID, then the db is cleared as a side-effect.... hmmm

verify/verify.go
```diff
@@ -118,6 +118,26 @@ func (db *DB) Unmarshal(b []byte, v interface{}, role string, minVersion int) er
...
+	}
+	// Note: If verification fails, then we wont attemp to unmarshal
+	// unless when verification error is errExpired.
+	verifyErr := db.Verify(s, role, minVersion)
+	if verifyErr != nil {
+		if _, ok := verifyErr.(ErrExpired); !ok {
+			return verifyErr
+		}
+	}
+	if err := json.Unmarshal(s.Signed, v); err != nil {
```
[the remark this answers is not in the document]
hosseinsia: done!

verify/verify.go
```diff
@@ -118,6 +118,26 @@ func (db *DB) Unmarshal(b []byte, v interface{}, role string, minVersion int) er
 	return json.Unmarshal(s.Signed, v)
 }
 
+// UnmarshalExpired is exactly like Unmarshal except ignores expired timestamp error.
+func (db *DB) UnmarshalExpired(b []byte, v interface{}, role string, minVersion int) error {
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client.go
```diff
@@ -19,6 +21,7 @@ const (
 	defaultRootDownloadLimit      = 512000
 	defaultTimestampDownloadLimit = 16384
 	defaultMaxDelegations         = 32
+	defaultMaxRoots               = 10000
```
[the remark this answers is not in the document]
hosseinsia: Functions are not allowed for setting the consts so we can't use Pow(2, 10). 1e3 or just 1000 sounds a reasonable replacement.

client/client.go
```diff
@@ -86,13 +89,17 @@ type Client struct {
 	// MaxDelegations limits by default the number of delegations visited for any
 	// target
 	MaxDelegations int
+
+	// UpdaterMaxRoots limits the number of downloaded roots in 1.0.19 root updater
+	UpdaterMaxRoots int
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client_test.go
```diff
@@ -360,6 +362,107 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+
+		tufClient, closer := initTestClient(c, test.fixturePath /* initWithLocalMetadata = */, true /* ignoreExpired = */, true)
+		_, err := tufClient.Update()
+		if test.extpectedError == nil {
+			c.Assert(err, IsNil)
+			// Check if the root.json is being saved in non-volatile storage.
+			tufClient.getLocalMeta()
+			assert.Equal(c, test.expectedRootVersion, tufClient.rootVer)
+		} else {
+			if _, ok := test.extpectedError.(ErrDecodeFailed); ok {
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client_test.go
```diff
@@ -567,10 +670,9 @@ func (s *ClientSuite) TestUpdateLocalRootExpired(c *C) {
 		if _, ok := err.(verify.ErrExpired); !ok {
 			c.Fatalf("expected err to have type signed.ErrExpired, got %T", err)
 		}
-
-		client := NewClient(s.local, s.remote)
 		_, err = client.Update()
 		c.Assert(err, IsNil)
+		c.Assert(client.rootVer, Equals, 2)
```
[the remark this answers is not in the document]
hosseinsia: This is the root version number we expect. I defined a constant to make this more readable.

client/client_test.go
```diff
@@ -625,8 +727,10 @@ func (s *ClientSuite) TestUpdateLocalRootExpiredKeyChange(c *C) {
 	// replace all keys
 	newKeyIDs := make(map[string][]string)
 	for role, ids := range s.keyIDs {
-		c.Assert(s.repo.RevokeKey(role, ids[0]), IsNil)
-		newKeyIDs[role] = s.genKey(c, role)
+		if role != "snapshot" && role != "timestamp" && role != "targets" {
```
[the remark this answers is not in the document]
hosseinsia: removed.

client/client_test.go
```diff
@@ -704,7 +808,7 @@ func (s *ClientSuite) TestUpdateReplayAttack(c *C) {
 	c.Assert(s.repo.Timestamp(), IsNil)
 	s.syncRemote(c)
 	_, err := client.Update()
-	c.Assert(IsLatestSnapshot(err), Equals, true)
+	//c.Assert(IsLatestSnapshot(err), Equals, true)
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client.go
hosseinsia: The intent is to verify them. However, `getLocalRootMeta` comes with the side effect of removing all the keys by creating a NewDB and loading new keys.

client/errors.go
hosseinsia: done!

client/client.go
hosseinsia: The whole `updateWithLatestRoot` is now obsolete. Removing the method altogether.

client/client.go
hosseinsia: Unfortunately, the current implementation does not comply with what you described. In order for the getLocalRootMeta to work, we store the 'root.json' permanently by calling `c.local.SetMeta("root.json", ...)` and then do one extra check (N+1 signed itself). Therefore, in case of failure, it can not be recovered next time (@trishankatdatadog can you please double check).

The issue at the moment is first line `meta, err := c.local.GetMeta()` in the `getLocalRootMeta` (as well as the previous implementation of the getLocalMeta) that loads metadata from levelDB. 

I am up for a change if we agree on more changes here. It does not make sense to leave the client in a bad state. WDYT?

client/client.go
```diff
@@ -247,26 +223,170 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-	}
-	if err := c.decodeRoot(rootJSON); err != nil {
-		return nil, err
-	}
-	if err := c.local.SetMeta("root.json", rootJSON); err != nil {
-		return nil, err
+		if m == nil {
+			rootJSON, err = c.downloadMetaUnsafe("root.json", defaultRootDownloadLimit)
+		} else {
+			rootJSON, err = c.downloadMetaFromSnapshot("root.json", *m)
```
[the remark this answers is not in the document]
hosseinsia: The method containing this is obsolete now. No test fails, so seems like we are not using it anymore.

hosseinsia (commented):
updated according to comments. PTAL.

Major changes:
* The code is being changes such won't persist a bad root metadata (except for when it is expired).
* Added a few more tests to check for the cases of non-root key change.
* Refactored/simplified the client_test

client/errors.go
```diff
@@ -133,3 +133,12 @@ type ErrRoleNotInSnapshot struct {
 func (e ErrRoleNotInSnapshot) Error() string {
 	return fmt.Sprintf("tuf: role %s not in snapshot version %d", e.Role, e.SnapshotVersion)
 }
+
+type ErrWrongRootVersion struct {
```
raphaelgavache, line 137: This error isn't used, `ErrWrongVersion` is used instead

verify/errors.go
```diff
@@ -51,6 +51,15 @@ func (e ErrLowVersion) Error() string {
 	return fmt.Sprintf("version %d is lower than current version %d", e.Actual, e.Current)
 }
 
+type ErrWrongVersion struct {
+	Given    int
+	Expected int
+}
+
+func (e ErrWrongVersion) Error() string {
+	return fmt.Sprintf("version %d does not match the extepcted version %d", e.Given, e.Expected)
```
raphaelgavache, line 60: nit: s/extepcted/expected

verify/verify.go
```diff
@@ -118,6 +118,23 @@ func (db *DB) Unmarshal(b []byte, v interface{}, role string, minVersion int) er
 	return json.Unmarshal(s.Signed, v)
 }
 
+// UnmarshalExpired is exactly like Unmarshal except ignores expired timestamp error.
+func (db *DB) UnmarshalIgnoreExpired(b []byte, v interface{}, role string, minVersion int) error {
+	s := &data.Signed{}
+	if err := json.Unmarshal(b, s); err != nil {
+		return err
+	}
+	// Note: If verification fails, then we wont attemp to unmarshal
```
raphaelgavache, line 127: nit: s/attemp/attempt

client/client_test.go
```diff
@@ -360,6 +364,113 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+// startTUFRepoServer starts a HTTP server to serve a TUF Repo.
+func startTUFRepoServer(baseDir string, relPath string) (net.Listener, error) {
+	serverDir := filepath.Join(baseDir, relPath)
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	go http.Serve(l, http.FileServer(http.Dir(serverDir)))
+	return l, err
+}
+
+// newClientWithMeta creates new client and sets the root metadata for it.
+func newClientWithMeta(baseDir string, relPath string, serverAddr string, initWithLocalMetadata bool) (*Client, error) {
```
raphaelgavache, line 376: `initWithLocalMetadata` seems to be a left over: it's not used in this function, it's always set to true when called, local meta are always loaded at initialisation

raphaelgavache (commented):
Nice PR! This is my first pass, I've left a few minor comments. I'll go through the test cases next

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
 	if err := c.local.SetMeta("snapshot.json", snapshotJSON); err != nil {
```
trishankatdatadog, line 216: Acording to [spec](https://theupdateframework.github.io/specification/v1.0.19/index.html#update-snapshot), this should be done right before going to targets.

client/client.go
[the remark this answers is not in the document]
trishankatdatadog: Agreed. Let's try not nuking the client DB until success...

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		}
 	}
+	m, err := c.local.GetMeta()
 	if err != nil {
-		return nil, err
+		return err
 	}
-	if err := c.decodeRoot(rootJSON); err != nil {
-		return nil, err
+	nRootMetadata := m["root.json"]
```
trishankatdatadog, line 239: Maybe add a comment and optionally move it to right before the FOR loop to indicate what it's for.

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	// 5.3.2 Let N denote the version number of the trusted root metadata file.
+	for i := 0; i < c.MaxRootRotations; i++ {
+		// 5.3.3 Try downloading version nPlusOne of the root metadata file.
+		// NOTE: as a side effect, we do update c.rootVer to nPlusOne between iterations.
+		nPlusOne := c.rootVer + 1
+		nPlusOneRootPath := util.VersionedPath("root.json", nPlusOne)
+		nPlusOneRootMetadata, err := c.downloadMetaUnsafe(nPlusOneRootPath, defaultRootDownloadLimit)
+
+		if err != nil {
+			if _, ok := err.(ErrMissingRemoteMetadata); ok {
```
trishankatdatadog, line 289: Do you get this error for both 403 and 404 HTTP responses? S3 uses 403 as a response code.

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-	if err := c.local.SetMeta("root.json", rootJSON); err != nil {
-		return nil, err
+
+	// 5.3.11 If the timestamp and / or snapshot keys have been rotated,
+	// then delete the trusted timestamp and snapshot metadata files.
+	for topLevelRolename := range nonRootManifests {
+		if !reflect.DeepEqual(
+			nonRootManifests[topLevelRolename],
+			getKeyIDs(topLevelRolename)) {
+			c.local.SetMeta(topLevelRolename, json.RawMessage{})
```
trishankatdatadog, line 347: There are two sets of tests we need:
1. When timestamp/snapshot/targets have _not_ been rotated (which is the common case), we should check that rollback attack preventions work.
2. Otherwise, we should check that rollback attack preventions _don't_ work.

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-		}
-		c.consistentSnapshot = root.ConsistentSnapshot
-	} else {
-		return ErrNoRootKeys
+	meta, err := c.local.GetMeta()
+	if err != nil {
+		return nil
 	}
 
 	if snapshotJSON, ok := meta["snapshot.json"]; ok {
```
trishankatdatadog, line 369: The rest of this logic is irrelevant to this PR, but the logic is not even correct. First you should verify timestamp against root, then snapshot against timestamp, then targets against snapshot.

trishankatdatadog (changes requested):
So close! Thank you very much.

client/client.go
raphaelgavache: This error is set by the [remote store interface](https://github.com/theupdateframework/go-tuf/blob/master/client/client.go#L35-L56). The responsibility to implement it well for s3 or any remote store is left to a remote store implementation and can be ignored here

client/client.go
```diff
@@ -148,27 +155,13 @@ func (c *Client) Update() (data.TargetFiles, error) {
 }
 
 func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
-	// Always start the update using local metadata
-	if err := c.getLocalMeta(); err != nil {
+	if err := c.updateRoots(); err != nil {
```
raphaelgavache, line 158: nit: as latestRoot bool is not used anymore `update` func can be removed in favor of Update()

Update func comment is also outdated

client/client.go
```diff
@@ -208,23 +191,9 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-		// ErrRoleThreshold could indicate snapshot keys have been
-		// revoked, so retry with the latest root.json
-		if isDecodeFailedWithErrRoleThreshold(err) && !latestRoot {
-			return c.updateWithLatestRoot(nil)
-		}
 		return nil, err
 	}
 
-	// If we don't have the root.json, download it, save it in local
-	// storage and restart the update
```
raphaelgavache, line 220: Removing this and enforcing the new root rotation on each update is a breaking change with the [implemented workflow spec 0.9.9 ](https://github.com/theupdateframework/tuf/blob/v0.9.9/docs/tuf-spec.txt#L714-L737).

I think this should be explicitly mentioned in the PR description in case some still rely on it.

client/client.go
```diff
@@ -247,76 +216,154 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-		return nil, err
+
+	// 5.3.11 If the timestamp and / or snapshot keys have been rotated,
+	// then delete the trusted timestamp and snapshot metadata files.
+	for topLevelRolename := range nonRootManifests {
+		if !reflect.DeepEqual(
+			nonRootManifests[topLevelRolename],
+			getKeyIDs(topLevelRolename)) {
+			c.local.SetMeta(topLevelRolename, json.RawMessage{})
+		}
```
raphaelgavache, line 348: Until now if a meta is found in the local store it would have the matching json encoding. This adds a new state where the key "timestamp" can be found but no valid matching file is stored.

What do you think of adding a DeleteMeta func in the local store interface? Since we are enforcing the latest spec it might be the right moment to update the local store interface.

I didn't find anything currently broken with the invalid metas stored so it might not be relevant.

client/client.go
trishankatdatadog: Well, we should check how the remote store interface handles returns a 403, because it might break any client that depends on S3.

client/client.go
trishankatdatadog: In fact, for (2) you should go further and check that we can also recover from fast-forward attacks.

client/client.go
hosseinsia: I added the DeleteMeta and updated the `updateRoots` accordingly. PTAL.

client/client_test.go
```diff
@@ -360,6 +362,107 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+func (s *ClientSuite) TestUpdateRoots(c *C) {
+	var tests = []struct {
+		fixturePath         string
+		isExpired           bool // Value retuned by verify.IsExpired.
+		extpectedError      error
+		expectedRootVersion int // -1 means no check is performed on this.
+	}{
+		// Good new root update succeeds (the timestamp check disabled).
+		{"testdata/PublishedTwiceWithRotatedKeys_root", false, nil, 2},
+		// Good new root update with a new key for timestamp succeeds.
```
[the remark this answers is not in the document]
trishankatdatadog: We decided not to test the whole thing (A->P->Z), but just what this PR does (A->P), which means just testing that timestamp/snapshot/targets get deleted when their keys are rotated.

client/client.go
trishankatdatadog: See my earlier comment about going from A->P instead of A->Z.

raphaelgavache (dismissed):
lgtm!

client/client_test.go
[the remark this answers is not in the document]
hosseinsia: implementation/test added for the fast forward attack recovery. PTAL.

client/client.go
```diff
@@ -243,80 +219,197 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	}
+
+	// 5.3.11 To recover from fast-forward attack, certain metadata files need
+	// to be deleted if a threshold of keys are revoked.
+	// List of metadata that should be deleted per role if a threshold of keys
+	// are revoked:
+	// timestamp -> delete timestamp.json
+	// snapshot ->  delete timestamp.json and snapshot.json
+	// targets ->   delete snapshot.json and targets.json
+	for topLevelRolename := range nonRootManifests {
```
trishankatdatadog, line 374: Let's link to the [PR](https://github.com/theupdateframework/specification/pull/150) detailing this logic, pls, and make an issue documenting that this has not been definitely resolved yet.

client/client_test.go
```diff
@@ -360,6 +364,168 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		verify.IsExpired = e
+	}
+}
+
+func (s *ClientSuite) TestFastForwardAttackRecovery(c *C) {
+	var tests = []struct {
+		fixturePath       string
+		expectMetaDeleted map[string]bool
+	}{
+		// No non-root-metadata recovery if root keys are revoked only.
```
trishankatdatadog, line 479: I don't understand the tests, especially by reading the comments. Please clarify?

trishankatdatadog (changes requested):
Almost there!

client/client.go
```diff
@@ -138,37 +149,21 @@ func (c *Client) Init(rootKeys []*data.Key, threshold int) error {
...
+// It always performs root update (5.2 and 5.3) section of the v1.0.19 spec.
 //
-// It performs the update part of "The client application" workflow from
-// section 5.1 of the TUF spec:
-//
-// https://github.com/theupdateframework/tuf/blob/v0.9.9/docs/tuf-spec.txt#L714
+// https://theupdateframework.github.io/specification/v1.0.19/index.html#load-trusted-root
 func (c *Client) Update() (data.TargetFiles, error) {
-	return c.update(false)
+	return c.update()
```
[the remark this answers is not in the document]
hosseinsia: probably not, at least for now. Removing it.

client/client.go
```diff
@@ -243,80 +219,197 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+
+	// 5.3.10 Check for a freeze attack.
+	// NOTE: this will check for any, including freeze, attack.
+	if err := c.loadAndVerifyLocalRootMeta(); err != nil {
+		return err
 	}
-	if err := c.local.SetMeta("root.json", rootJSON); err != nil {
-		return nil, err
+
+	countDeleted := func(s1 []string, s2 []string) int {
```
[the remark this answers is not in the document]
hosseinsia: The *** golang does not have set! Using idiomatic map trick!

client/client.go
```diff
@@ -243,80 +219,197 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	// The manifest looks like this:
+	// {
+	//	"timestamp": ["KEYID1", "KEYID2"],
+	//	"snapshot": ["KEYID3"],
+	//	"targets": ["KEYID4", "KEYID5", "KEYID6"]
+	// }
+	nonRootManifests := map[string][]string{"timestamp": {}, "snapshot": {}, "targets": {}}
+	nonRootThreshold := map[string]int{"timestamp": 1, "snapshot": 1, "targets": 1}
+	for k := range nonRootManifests {
+		nonRootManifests[k], nonRootThreshold[k] = getKeyIDs(k)
```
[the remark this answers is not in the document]
hosseinsia: done

client/client.go
hosseinsia: Done! Done! New issue: https://github.com/theupdateframework/go-tuf/issues/149

client/client_test.go
hosseinsia: Tried to reorganize and improve the comments. PTAL.

client/client.go
```diff
@@ -243,80 +219,197 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		return keyIDs, c.db.GetRole(role).Threshold
+	}
+
+	// The manifest looks like this:
+	// {
+	//	"timestamp": ["KEYID1", "KEYID2"],
+	//	"snapshot": ["KEYID3"],
+	//	"targets": ["KEYID4", "KEYID5", "KEYID6"]
+	// }
+	nonRootManifests := map[string][]string{"timestamp": {}, "snapshot": {}, "targets": {}}
```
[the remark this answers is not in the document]
hosseinsia: had to change the naming to merge the keyid and snapshot .

hosseinsia:
@trishankatdatadog addressed all comments. PTAL.

client/client.go
```diff
@@ -338,18 +429,107 @@ func (c *Client) getLocalMeta() error {
...
+	}
+	for name, role := range root.Roles {
+		if err := ndb.AddRole(name, role); err != nil {
 			return err
 		}
-		c.timestampVer = timestamp.Version
 	}
-
-	c.localMeta = meta
+	// Any trusted local root metadata version must be greater than 0.
```
raphaelgavache, line 477: The client.rootVer variable needs to be updated with the trusted meta root version. This needs to be done before the Verify err in case that meta is expired.

Else if we start with a root of version 5, the download algorithm will try to load root 1,2,3..+. This will likely fail as root5 keys do not sign root1 after a rotation

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		if err := c.local.SetMeta("root.json", nPlusOneRootMetadata); err != nil {
+			return err
+		}
+		nRootMetadata = nPlusOneRootMetadata
+		// 5.3.9 Repeat steps 5.3.2 to 5.3.9
+
+	} // End of the for loop.
+
+	// Activate the check for expiration.
+	verify.IsExpired = e
```
asraa, line 346: Do you need the `defer func() { verify.IsExpired = e }()` since there's this manual re-activation, or is it just extra redundancy?

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		KeyIDs    map[string]bool
+		Threshold int
 	}
-	if err := c.local.SetMeta("root.json", rootJSON); err != nil {
-		return nil, err
+
+	// Prepare for 5.3.11: If the timestamp and / or snapshot keys have been rotated,
+	// then delete the trusted timestamp and snapshot metadata files.
+	getKeyInfo := func(role string) KeyInfo {
+		//keyIDs := make([]string, 0, len(c.db.GetRole(role).KeyIDs))
```
asraa, line 248: nit: remove these commented out lines of code from old versions

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+			//keyIDs = append(keyIDs, k)
+			keyIDs[k] = true
+		}
+		//sort.Strings(keyIDs)
+		return KeyInfo{keyIDs, c.db.GetRole(role).Threshold}
+	}
+
+	// The manifest looks like this:
+	// {
+	//	"timestamp": ["KEYID1", "KEYID2"],
```
asraa, line 260: update: it includes the threshold

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+	//	"snapshot": ["KEYID3"],
+	//	"targets": ["KEYID4", "KEYID5", "KEYID6"]
+	// }
+
+	nonRootKeyInfo := map[string]KeyInfo{"timestamp": {}, "snapshot": {}, "targets": {}}
+	//nonRootThreshold := map[string]int{"timestamp": 1, "snapshot": 1, "targets": 1}
```
asraa, line 266: remove here again

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+		// NOTE: as a side effect, we do update c.rootVer to nPlusOne between iterations.
+		nPlusOne := c.rootVer + 1
+		nPlusOneRootPath := util.VersionedPath("root.json", nPlusOne)
+		nPlusOneRootMetadata, err := c.downloadMetaUnsafe(nPlusOneRootPath, defaultRootDownloadLimit)
+
+		if err != nil {
+			if _, ok := err.(ErrMissingRemoteMetadata); ok {
+				// stop when the next root can't be downloaded
+				break
+			} else {
```
asraa, line 302: do you need explicit else?

client/client.go
```diff
@@ -243,80 +214,200 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
+
+	// 5.3.11 To recover from fast-forward attack, certain metadata files need
+	// to be deleted if a threshold of keys are revoked.
+	// List of metadata that should be deleted per role if a threshold of keys
+	// are revoked:
+	// (based on the ongoing PR: https://github.com/mnm678/specification/tree/e50151d9df632299ddea364c4f44fe8ca9c10184)
+	// timestamp -> delete timestamp.json
+	// snapshot ->  delete timestamp.json and snapshot.json
+	// targets ->   delete snapshot.json and targets.json
+	for topLevelRolename := range nonRootKeyInfo {
```
asraa, line 372: a little confused about this: we are checking if a threshold of keys was deleted between the original trusted out-of-band root (nonRootKeyInfo) and the latest valid root?

Just double checking. I would appreciate a comment so it's clear that nonRootKeyInfo is supposed to be referring to the out of band one.

asraa (commented):
Thanks so much for continuing to review!

client/client.go
hosseinsia: If the code exists at any other point before this, the defer is going to take care of setting IsExpired to its initial state. Do you agree?

asraa: YES Thank you!!!

client/client.go
hosseinsia: Done!

client/client.go
hosseinsia: done!

client/client.go
hosseinsia: done!

client/client.go
hosseinsia: removed.

client/client.go
hosseinsia: > we are checking if a threshold of keys was deleted between the original trusted out-of-band root (nonRootKeyInfo) and the latest valid root?

Almost true; The nonRootKeyInfo is not necessarily out-of-band. Only for the first time the client calls update the nonRootKeyInfo is the very first root.json provided to the client out of band. For later updates, it is the the last root.json from the previous update. 

I'll try to make it clear by commenting.

client/client_test.go
```diff
@@ -360,6 +364,176 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		{"testdata/PublishedTwiceMultiKeysadd_9_revoke_2_threshold_4_targets",
+			map[string]bool{"root.json": false, "timestamp.json": false, "snapshot.json": false, "targets.json": false}},
+		{"testdata/PublishedTwiceMultiKeysadd_9_revoke_2_threshold_4_timestamp",
+			map[string]bool{"root.json": false, "timestamp.json": false, "snapshot.json": false, "targets.json": false}},
+
+		// Fast-forward recovery not needed if root keys are revoked, even when the threshold number of root keys are revoked.
+		{"testdata/PublishedTwiceMultiKeysadd_9_revoke_4_threshold_4_root",
+			map[string]bool{"root.json": false, "timestamp.json": false, "snapshot.json": false, "targets.json": false}},
+
+		// Delete snapshot and timestamp metadata if threashold number of snapshot keys are revoked.
```
trishankatdatadog, line 502:
```suggestion
		// Delete snapshot and timestamp metadata if a threshold number of snapshot keys are revoked.
```

client/client_test.go
```diff
@@ -360,6 +364,176 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		{"testdata/PublishedTwiceMultiKeysadd_9_revoke_4_threshold_4_snapshot",
+			map[string]bool{"root.json": false, "timestamp.json": true, "snapshot.json": true, "targets.json": false}},
+		// Delete targets and snapshot metadata if threashold number of targets keys are revoked.
```
trishankatdatadog, line 505:
```suggestion
		// Delete targets and snapshot metadata if a threshold number of targets keys are revoked.
```

client/client_test.go
```diff
@@ -360,6 +364,176 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		{"testdata/PublishedTwiceMultiKeysadd_9_revoke_4_threshold_4_targets",
+			map[string]bool{"root.json": false, "timestamp.json": false, "snapshot.json": true, "targets.json": true}},
+		// Delete timestamp metadata if threashold number of timestamp keys are revoked.
```
trishankatdatadog, line 508:
```suggestion
		// Delete timestamp metadata if a threshold number of timestamp keys are revoked.
```

trishankatdatadog (dismissed):
Fantastic work, thank you very much, Hossein! 🎉

client/client.go
hosseinsia: Nice catch. Updated the method to set the rootVersion -- although we are updating in the updateRoots as well.
But this is the right thing to do.

client/client.go
```diff
@@ -243,80 +214,199 @@ func (c *Client) update(latestRoot bool) (data.TargetFiles, error) {
...
-	} else {
-		rootJSON, err = c.downloadMetaFromSnapshot("root.json", *m)
+func (c *Client) updateRoots() error {
+	// https://theupdateframework.github.io/specification/v1.0.19/index.html#load-trusted-root
+	// Temporarily deactivate expiration check. This is consistent with 5.3.6:
+	// Note that the expiration of the trusted root metadata file does not
+	// matter, because we will attempt to update it in the next step.
+	e := verify.IsExpired
+	defer func() { verify.IsExpired = e }()
+	verify.IsExpired = func(t time.Time) bool { return false }
```
raphaelgavache, line 227: You added `UnmarshalIgnoreExpired` but it doesn't seem to be used anymore. I think you should remove the verify.IsExpired override from non testing code and use the IgnoreExpired func.
Exemple of side effect of updating a global variable, if two different clients are run concurrently on the same process, there's a race condition on verify.IsExpired

Adding a test and running `go test -race` in the CI would catch it
```
func TestRace(t *testing.T) {
        for i := 0; i < 2; i++ {
                go func() {
                        c := NewClient(MemoryLocalStore(), newFakeRemoteStore())
                        c.Update()
                }()
        }
}
==================
WARNING: DATA RACE
Read at 0x000001aa7330 by goroutine 10:
  github.com/theupdateframework/go-tuf/client.(*Client).updateRoots()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client.go:225 +0x64
  github.com/theupdateframework/go-tuf/client.(*Client).Update()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client.go:155 +0x55
  github.com/theupdateframework/go-tuf/client.TestRace.func1()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client_test.go:1158 +0x164

Previous write at 0x000001aa7330 by goroutine 8:
  github.com/theupdateframework/go-tuf/client.(*Client).updateRoots()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client.go:227 +0xbe
  github.com/theupdateframework/go-tuf/client.(*Client).Update()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client.go:155 +0x55
  github.com/theupdateframework/go-tuf/client.TestRace.func1()
      /Users/raphael.gavache/go/src/github.com/hosseinsia/go-tuf/client/client_test.go:1158 +0x164
```

raphaelgavache (dismissed):
lgtm!

trishankatdatadog:
@hosseinsia do we have tests for whether expirations are ignored and not ignored correctly?

hosseinsia:
> @hosseinsia do we have tests for whether expirations are ignored and not ignored correctly?

Just added a test. ptal.

client/client.go
hosseinsia: fixed.

client/client_test.go
```diff
@@ -360,6 +364,185 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+	return c, nil
+}
+
+func initRootTest(c *C, baseDir string) (*Client, func() error) {
+	l, err := startTUFRepoServer(baseDir, "server")
+	c.Assert(err, IsNil)
+	//e := verify.IsExpired
+	//if ignoreExpired {
+	//	verify.IsExpired = func(t time.Time) bool { return false }
+	//}
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client_test.go
```diff
@@ -360,6 +364,185 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+	tufClient, err := newClientWithMeta(baseDir, "client/metadata/current", l.Addr().String())
+	//verify.IsExpired = e
```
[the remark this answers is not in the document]
hosseinsia: done!

raphaelgavache (dismissed):

joshuagl (commented):
I reviewed the changes in client.go and this looks like a clean implementation of [5.3](https://theupdateframework.github.io/specification/latest/#update-root) – nice work @hosseinsia !

One thing I did notice while reading through client.go is that [5.2](https://theupdateframework.github.io/specification/latest/#load-trusted-root) does not appear to be implemented, instead a client is initialised with a list of trusted keys (rather than a bootstrap trusted root metadata). I assume fixing that is out of scope of this PR? Perhaps we could file an issue to track the work?

client/client.go
```diff
@@ -19,6 +20,7 @@ const (
 	defaultRootDownloadLimit      = 512000
 	defaultTimestampDownloadLimit = 16384
 	defaultMaxDelegations         = 32
+	defaultMaxRootRotations       = 1e3
```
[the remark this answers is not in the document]
hosseinsia: The spec meant to suggest a reasonable number (not very big and not so small like 32) and 1024 is only a suggestion. I do not have any objection if you think 1024 makes it more readable/consistent with the spec. WDYT?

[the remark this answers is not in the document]
joshuagl: Absolutely no objection to 1000, just curious how you chose it.

client/client_test.go
```diff
@@ -360,6 +364,198 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		expectedError    error
+		expectedVersions map[string]int
+	}{
+		// Test cases for Client's root update:
+		// (1) Succeeds when there is no root update.
+		// (2) Succeeds updating root to version 2 when the client's initial root version is 1.
+		// (3) Succeeds updating root to version 2 when the client's initial root version is 1 and version 1 is expired.
+		// (4) Succeeds updating root to version 3 when the client's initial root version is 1 and root versions 1 and 2 are expired.
+		// (5) Succeeds updating root to version 3 when the client's initial root version is 2.
+		// (6) Fails updating root to version 3 when the client's initial root version is 1 and root versions 3 is expired.
```
trishankatdatadog, line 418:
```suggestion
		// (6) Fails updating root to version 3 when the client's initial root version is 1 and root version 3 is expired.
```

client/client_test.go
```diff
@@ -360,6 +364,198 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		// (7) Fails updating root to version 2 when the root the old root's signature is invalid (nth root didn't sign off n+1).
```
trishankatdatadog, line 419:
```suggestion
		// (7) Fails updating root to version 2 when old root 1 did not sign off on it (nth root didn't sign off n+1).
```

client/client_test.go
```diff
@@ -360,6 +364,198 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		// (8) Fails updating root to version 2 when the root the new root's signature is invalid (n+1th root didn't sign off n+1).
```
trishankatdatadog, line 420:
```suggestion
		// (8) Fails updating root to version 2 when the new root 2 did not sign itself (n+1th root didn't sign off n+1).
```

client/client_test.go
```diff
@@ -360,6 +364,198 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+	return tufClient, l.Close
+}
+
+func (s *ClientSuite) TestUpdateRoots(c *C) {
+	var tests = []struct {
+		fixturePath      string
+		expectedError    error
+		expectedVersions map[string]int
+	}{
+		// Test cases for Client's root update:
```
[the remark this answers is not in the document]
hosseinsia: done!

client/client_test.go
```diff
@@ -360,6 +364,198 @@ func (s *ClientSuite) TestNewRoot(c *C) {
...
+		// (1) Succeeds when there is no root update.
+		// (2) Succeeds updating root to version 2 when the client's initial root version is 1.
```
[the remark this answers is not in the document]
hosseinsia: done!

trishankatdatadog (commented):
💯

trishankatdatadog (approved):

trishankatdatadog:
Hmm, not sure why we're seeing some files in master that are not supposed to be there... let's discuss offline

raphaelgavache (dismissed):

trishankatdatadog (approved):

hosseinsia:
@asraa and @joshuagl ptal!

raphaelgavache (dismissed):

client/client_test.go
```diff
@@ -411,6 +413,8 @@ func (s *ClientSuite) TestUpdateRoots(c *C) {
 	}{
 		// Succeeds when there is no root update.
 		{"testdata/Published1Time", nil, map[string]int{"root": 1, "timestamp": 1, "snapshot": 1, "targets": 1}},
+		// Succeeds when client only has root.json
+		{"testdata/Published1Time_client_root_only", nil, map[string]int{"root": 1, "timestamp": 1, "snapshot": 1, "targets": 1}},
```
trishankatdatadog, line 417: Doesn't make sense. The client does have only root in the current metadata cache, but it has all files in the previous metadata cache. Should the previous cache be there?

hosseinsia: I removed the files in the 'previous' folder for this test . -- regardless the root update does not use the previous metadata cache, based on the way that we set-up the local repo using the "current" path.

trishankatdatadog (approved):

raphaelgavache (approved):

asraa (approved):
Took another pass, LGTM!

PR #148, opened 2021-09-01
WIP Refactor keys to make key additions friendly 

asraa:
Local tests passes, still working on interop tests and double-checking.

This change makes it more friendly to add new keytypes:
* There are now two data types, `data.Key` and `data.PrivateKey` representing the json data for public and private keys
* There are two interfaces `Signer` and `Verifier` in the keys package that define conversions between the data formats and methods that are needed by the other modules
* There is a map `KeyMap` that registers a key type with the Signer/Verifier implementation. You don't need both to register in the key map (e.g. go-tuf only verifies ecdsa, does not support signing)


client/client_test.go
```diff
@@ -352,7 +353,7 @@ func (s *ClientSuite) TestNewRoot(c *C) {
 		for _, id := range ids {
 			key := client.db.GetKey(id)
 			c.Assert(key, NotNil)
-			c.Assert(key.IDs(), DeepEquals, ids)
+			c.Assert((*key).IDs(), DeepEquals, ids)
```
asraa, line 356: go help: key is an implementation of an interface so i think i had to dereference this
maybe instead i should return the object from GetKey along with an error rather than checking IsNil?

client/python_interop/python_interop_test.go
```diff
@@ -42,6 +43,21 @@ func (t *testDestination) Delete() error {
 	return nil
 }
 
+func getVerifier(c *C, key *data.Key) keys.Verifier {
```
asraa, line 46: maybe i should expose this in the keys package...

local_store.go
```diff
@@ -23,15 +43,15 @@ func MemoryStore(meta map[string]json.RawMessage, files map[string][]byte) Local
...
+		signers:    make(map[string][]*data.PrivateKey),
 	}
 }
 
 type memoryStore struct {
 	meta       map[string]json.RawMessage
 	stagedMeta map[string]json.RawMessage
 	files      map[string][]byte
-	signers    map[string][]sign.Signer
+	signers    map[string][]*data.PrivateKey
```
asraa, line 54: i should make this consistent: and store the signers rather than the data like i do for filesystemstore

verify/verifiers.go
```diff
@@ -1,75 +1 @@
 package verify
```
d-niu, line 1: Should this file be deleted?

client/client_test.go
asraa: changed to key, err

asraa:
addressed some comments

d-niu (commented):
Overall, I really like the generalized approach to key types and separating the interfaces for signing/verification accordingly.

My biggest question is around the structure of the Signer interface + implementing a signer struct for each key. Right now, it's not very flexible for generating signatures (or verifications) in a kms.

Ideally, I would like to use as much of the same code to create keys with go's crypto package and any type of external key gen system. Right now, I'm thinking that creating keys, signing, verifying, and getting public keys should all be interface methods of kms. I'm not too sure what the best approach is here, but I think making the Signer and Verifier more abstract or even changing how we store things in the Key Map would work.

Here's the [general structure](https://gist.github.com/d-niu/e501285d352f8599435f569882d7ea05) of what I was thinking for key management systems. Though I have to admit, I wasn't really writing this with the current way keys are generating in memory in mind :/

keys/keys.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return s.Verifier, nil
+}
+
+func GetSigner(key *data.PrivateKey) (Signer, error) {
+	st, ok := KeyMap.Load(key.Type)
+	if !ok {
+		return nil, ErrInvalidKey
+	}
+	s := st.(func() SignerVerifier)()
```
trishankatdatadog, line 94: What does this function do exactly? Why instantiate a SignerVerifier when it looks like you only use Signer?

trishankatdatadog (changes requested):
I think I am ok with this. I understand @d-niu's concerns about extending this API to abstract away things like where the signing operation is actually being done, but I think this can be fixed later.

Can you finish moving the files for keys under keys/ to pkg?

asraa:
>  Right now, I'm thinking that creating keys, signing, verifying, and getting public keys should all be interface methods of kms. I'm not too sure what the best approach is here, but I think making the Signer and Verifier more abstract or even changing how we store things in the Key Map would work.

Hm I think Signer and Verify should be separate (although they can be the same object in the case of a KMS client). Looking at your code, the part that probably won't fit is the crypto.Signer operation, since I don't think KMS easily supports `crypto.Signer` interface without a lot of finagling. Do you know what the Sign parameters should look like?

keys/keys.go
asraa: We store constructor methods in the map, so this invokes the constructor.

>  Why instantiate a SignerVerifier when it looks like you only use Signer?

Given that there's helper methods for loading signer/verifies I might as well store two separate maps (instead of just one that goes type -> SignerVerifier, split to type -> Signer, type -> Verifier).

asraa:
OK simplified the interface even more.

asraa:
> I might be missing something, but why can't GetVerifier() be a method of the key?

GetVerifier as a method `data.Key`? because it would introduce a circular dependency. (`data.Key` shouldn't know about the key implementations)

trishankatdatadog (dismissed):
Happy to unblock this for now, we can refine API if need be later

trishankatdatadog:
Oh, one more thing: we need RSA support, no?

asraa:
Sounds good, we'll need to work how KMS signers will deal with `crypto.Signer`. (Probably we'll just define the `Sign(...)` method in the interface). Sigstore has a massive workaround for it, but we should keep it simple.

asraa:
> Oh, one more thing: we need RSA support, no?

I have a local follow-up PR!

trishankatdatadog:
Who else should we have review this? @hosseinsia @joshuagl @mnm678? Should not both be DD

d-niu:
> Edit: Wanna chat about the way things are stored in the Key Map? The Key Map is just meant to translate "type name" to an "empty type implementation". `UnmarshalKey` is meant to take the public key data from the JSON (probably including key resource ID) and try to instantiate the kms client. A method (NOT in the interface), like `GenerateHashiVaultKey` would look like the `func HVClient(keyResourceID string) (KMSClient, error)` you have. Does that make sense?

So, generating different types of keys, signing, and verifying in kms should not necessarily be under a separate kms interface. Instead, the connection to a kms client like hashicorp vault should be generated when you use a key from the key map and realize that it has some other attributes during UnmarshalKey, for instance.

asraa:
> Instead, the connection to a kms client like hashicorp vault should be generated when you use a key from the key map and realize that it has some other attributes during UnmarshalKey, for instance.

Maybe the constructors held in the map should have an `Context` or `Params` object?

d-niu:
> Maybe the constructors held in the map should have an `Context` or `Params` object?

For sure. Would context/params just hold something to indicate whether its locally generated or in a kms though? We probably won't store kms access tokens or paths in these params.

asraa:
> Would context/params just hold something to indicate whether its locally generated or in a kms though? We probably won't store kms access tokens or paths in these params.

I think I would need to see the interface to understand. We should chat on Slack! 
* Does the client ever even need to know that this was a KMS key? Or does the JSON just include the public key (and it will thus use a normal ed25519/ecdsa verifier). 
* If it's just the signers that are need to know the KMS path or tokens, I think the only questionable thing is how do you know the paths/tokens needed when you unmarshal a key from the `keys` storage? Is this info contained in the data.PrivateKey JSON? If yes, then doesn't `UnmarshalSigner(key data.PrivateKey)` already handle this?

d-niu:
> > Would context/params just hold something to indicate whether its locally generated or in a kms though? We probably won't store kms access tokens or paths in these params.
> 
> I think I would need to see the interface to understand. We should chat on Slack!

dm'd you on the cncf slack!

repo.go
```diff
@@ -392,6 +401,10 @@ func validExpires(expires time.Time) bool {
 	return expires.Sub(time.Now()) > 0
 }
 
+/*
+
```
joshuagl, line 405: Are we missing a docstring here?

joshuagl (commented):
Just to be clear – is this PR ready for review/approval/merge now? Or is it blocked on the API discussion?

d-niu:
> Just to be clear – is this PR ready for review/approval/merge now? Or is it blocked on the API discussion?

Not blocked on API discussion!

repo.go
asraa: Removed, think it was a remnant

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// A Verifier verifies public key signatures.
+type Verifier interface {
+	// UnmarshalKey takes key data to a working verifier implementation for the key type.
+	// This performs any validation over the data.Key to ensure that the verifier is usable
+	// to verify signatures.
+	UnmarshalKey(key *data.Key) error
+
+	// Key returns the data.Key object associated with the verifier.
```
joshuagl, line 29:
```suggestion
	// MarshalKey returns the data.Key object associated with the verifier.
```
?

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+	// and determines whether the signature is valid for the given
+	// key and message.
+	Verify(msg, sig []byte) error
+}
+
+type Signer interface {
+	// Marshal into a private key.
+	MarshalSigner() (*data.PrivateKey, error)
+
+	// UnmarshalKey takes private key data to a working Signer implementation for the key type.
```
joshuagl, line 45:
```suggestion
	// UnmarshalSigner takes private key data to a working Signer implementation for the key type.
```

joshuagl (approved):
This is a nice cleanup, thanks @asraa

asraa:
@joshuagl @trishankatdatadog re-pinging for a new review because new commits made them stale

cce:
re: supporting KMS keys, if you're curious to see a similar full implementation of adding a `Signer` interface, as well as a `KeyManager` interface supporting key generation / management with AWS KMS, local files, and in-memory keys, there is a related work by @pawel-kedzior-sw in https://github.com/theupdateframework/go-tuf/compare/master...solarwinds:feature/kms#diff-465f5c46821c1f954b47a6e16744c3e80b3e013998a214134f8929931518efaa

d-niu:
> re: supporting KMS keys, if you're curious to see a similar full implementation of adding a `Signer` interface, as well as a `KeyManager` interface supporting key generation / management with AWS KMS, local files, and in-memory keys, there is a related work by @pawel-kedzior-sw [in this branch](https://github.com/theupdateframework/go-tuf/compare/master...solarwinds:feature/kms#diff-465f5c46821c1f954b47a6e16744c3e80b3e013998a214134f8929931518efaa)

😮  Interesting, I will take a look

asraa:
ping @trishankatdatadog @joshuagl could you please take a look?

joshuagl (dismissed):

trishankatdatadog:
@hosseinsia I am reviewing while I can. In the meantime, could you also please help review? Thank you!

pkg/keys/ecdsa.go
```diff
@@ -0,0 +1,71 @@
...
+	return nil
+}
+
+func (p *p256Verifier) MarshalKey() *data.Key {
+	return p.key
+}
+
+func (p *p256Verifier) UnmarshalKey(key *data.Key) error {
+	if err := json.Unmarshal(key.Value, p); err != nil {
+		return errors.New("unmarshalling key")
```
trishankatdatadog, line 63: A bit vague. Could we specify the key type+scheme?

pkg/keys/ecdsa.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	x, _ := elliptic.Unmarshal(elliptic.P256(), p.PublicKey)
+	if x == nil {
+		return errors.New("unmarshalling key")
```
trishankatdatadog, line 67: A bit vague. Could we specify the key type+scheme?

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (e *ed25519Verifier) MarshalKey() *data.Key {
+	return e.key
+}
+
+func (e *ed25519Verifier) UnmarshalKey(key *data.Key) error {
+	e.key = key
+	if err := json.Unmarshal(key.Value, e); err != nil {
+		return errors.New("unmarshalling key")
```
trishankatdatadog, line 49: A bit vague. Could we specify the key type+scheme?

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	if len(e.PublicKey) != ed25519.PublicKeySize {
+		return errors.New("unmarshalling key")
```
trishankatdatadog, line 52: A bit vague. Could we specify the key type+scheme?

verify/verify_test.go
```diff
@@ -51,12 +59,12 @@ func (s ecdsaSigner) ContainsID(id string) bool {
 	return s.PublicData().ContainsID(id)
 }
 
-func (ecdsaSigner) Type() string {
-	return data.KeyTypeECDSA_SHA2_P256
+func (ecdsaSigner) MarshalSigner() (*data.PrivateKey, error) {
+	return nil, errors.New("not implemented")
```
trishankatdatadog, line 63: What is going on here? Totally not obvious.

verify/verify_test.go
```diff
@@ -51,12 +59,12 @@ func (s ecdsaSigner) ContainsID(id string) bool {
...
 }
 
-func (ecdsaSigner) Scheme() string {
-	return data.KeySchemeECDSA_SHA2_P256
+func (ecdsaSigner) UnmarshalSigner(key *data.PrivateKey) error {
+	return errors.New("not implemented")
```
trishankatdatadog, line 67: What is going on here? Totally not obvious.

trishankatdatadog (changes requested):
Largely looks good to me, thanks, Asra!

I have open questions about how this will change @d-niu's external KMS providers, but we can figure that out later.

If not too much to ask, would you please add the RSA key support here in a single shot? Thanks in advance!

verify/verify_test.go
asraa: Ah, the verify package doesn't need to marshal/unmarshal signers for keys, so I left as unimplemented -- I'll clarify that this is for the test

asraa:
> would you please add the RSA key support here in a single shot?

Sure! Was just worried about the review burden but it's just a simple addition for a file.

asraa:
Added RSA key implementation.

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,124 @@
...
+	"encoding/json"
+	"errors"
+
+	"github.com/theupdateframework/go-tuf/data"
+)
+
+func init() {
+	SignerMap.Store(data.KeySchemeEd25519, NewP256Signer)
+	VerifierMap.Store(data.KeySchemeEd25519, NewP256Verifier)
+
```
hosseinsia, line 15: nit: remove extra line

local_store.go
```diff
@@ -304,7 +317,7 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 	return f.Clean()
 }
 
-func (f *fileSystemStore) GetSigningKeys(role string) ([]sign.Signer, error) {
+func (f *fileSystemStore) GetSigningKeys(role string) ([]keys.Signer, error) {
```
hosseinsia, line 320: Can we rename this to GetSigners/Signers?   --- because these are Signers we are returning, basically?

local_store.go
```diff
@@ -362,21 +375,13 @@ func (f *fileSystemStore) SavePrivateKey(role string, key *sign.PrivateKey) erro
...
-	for i, k := range keys {
-		res[i] = k.Signer()
-	}
-	return res
-}
-
 // loadKeys loads keys for the given role and returns them along with the
 // passphrase (if read) so that callers don't need to re-read it.
-func (f *fileSystemStore) loadKeys(role string) ([]*sign.PrivateKey, []byte, error) {
+func (f *fileSystemStore) loadKeys(role string) ([]*data.PrivateKey, []byte, error) {
```
hosseinsia, line 384: Can we rename to loadPrivateKeys?

local_store.go
```diff
@@ -11,10 +11,22 @@ import (
 
 	"github.com/theupdateframework/go-tuf/data"
 	"github.com/theupdateframework/go-tuf/encrypted"
-	"github.com/theupdateframework/go-tuf/sign"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
 	"github.com/theupdateframework/go-tuf/util"
 )
 
+func privateKeySigners(privateKeys []*data.PrivateKey) []keys.Signer {
```
hosseinsia, line 18: Consider renaming to signers?

hosseinsia (commented):
Great job! Started reviewing. Some minor comments so far.

local_store.go
asraa: Would be a big API change -- WDYT? Yeah, they're the Signers, which are signing keys

hosseinsia: Are we breaking anyone? -- Do we know if any external system using this directly?

hosseinsia: btw, you are already changing the method's signature.

asraa: Good point :P I think the only person that breaks is me in some sigstore code :dagger: 
So i'm happy to do so -- we also don't have a 1.0 release so I'm not too worried about conflicts

asraa: done

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (e *ed25519Verifier) MarshalKey() *data.Key {
+	return e.key
+}
+
+func (e *ed25519Verifier) UnmarshalKey(key *data.Key) error {
+	e.key = key
+	if err := json.Unmarshal(key.Value, e); err != nil {
+		return errors.New("unmarshalling ed25519 key")
```
hosseinsia, line 48: In the error message, please indicate that something went wrong. Example: "tuf: signature verification **failed**". I think prefixing that with "tuf:" is also a good idea.

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return &ed25519Signer{
+		PrivateKey:    ed25519.PrivateKey(data.HexBytes(private)),
+		keyType:       data.KeyTypeEd25519,
+		keyScheme:     data.KeySchemeEd25519,
+		keyAlgorithms: data.KeyAlgorithms,
+	}, nil
+}
+
+func (e *ed25519Signer) MarshalSigner() (*data.PrivateKey, error) {
```
hosseinsia, line 85: rename to MarshalPrivateKey?

pkg/keys/ed25519.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return &data.PrivateKey{
+		Type:       e.keyType,
+		Scheme:     e.keyScheme,
+		Algorithms: e.keyAlgorithms,
+		Value:      valueBytes,
+	}, nil
+}
+
+func (e *ed25519Signer) UnmarshalSigner(key *data.PrivateKey) error {
```
hosseinsia, line 101: rename to newSigner?

data/types.go
```diff
@@ -37,15 +39,22 @@ type Signature struct {
...
-	Type       string   `json:"keytype"`
-	Scheme     string   `json:"scheme"`
-	Algorithms []string `json:"keyid_hash_algorithms,omitempty"`
-	Value      KeyValue `json:"keyval"`
+	Type       string          `json:"keytype"`
+	Scheme     string          `json:"scheme"`
+	Algorithms []string        `json:"keyid_hash_algorithms,omitempty"`
+	Value      json.RawMessage `json:"keyval"`
 
 	ids    []string
```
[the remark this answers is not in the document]
hosseinsia: actually, you have the map from ids to verifiers in the db.keys so again, I am not sure what would be the usage of the ids here.

repo_test.go
```diff
@@ -39,9 +39,36 @@ func (RepoSuite) TestNewRepoIndent(c *C) {
 	})
 }
 
+// UniqueKeys returns the unique keys for each associated role.
+// We might have multiple key IDs that correspond to the same key.
+func UniqueKeys(r *data.Root) map[string][]*data.Key {
```
hosseinsia, line 44: Don't we need this in non-test as well?

data/types.go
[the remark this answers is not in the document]
asraa: > I am developing a bit of doubt around having ids here. 

The reason why is that so `IDs()` is only calculated once https://github.com/theupdateframework/go-tuf/blob/9545f820e4ee7995841c80d28cbac8c859b233f0/data/types.go#L58-L65 by holding on to the member variable.

> And now without it, Key would be same as PrivateKey.

Actually this is true as a result of the making the `Value` arbitrary dictionaries. But I think for clarity we should keep the Public and Private Keys different -- for a single key type, they respectively serialize the public and private values into those data types. Would it be more clear if I call `Value` `PublicValue` and `PrivateValue` for each of the two types? 

>  I am not sure what would be the usage of the ids here.

Calculating key ID is always done from the key data type since it's a hash of the JSON data of the key -- holding it as a variable is only to stop re-calculation. I don't see how the map from ID to verifiers would remove the need for it though? Since it is the reverse map.

pkg/keys/ed25519.go
asraa: Should I just return the JSON error then? 
`json: cannot unmarshal bool into Go value of type keys.ed25519Verifier"`

Ultimately this gets called through `GetVerifier` -- which will wrap the error string like `tuf: error unmarshalling key: json: cannot unmarshal..."

asraa: Fixed up error messages to contain `tuf: ` if they are generated from logic here and updated the style to indicate failure

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+	// UnmarshalSigner takes private key data to a working Signer implementation for the key type.
+	UnmarshalSigner(key *data.PrivateKey) error
+
+	// Returns the public data.Key from the private key
+	PublicData() *data.Key
+
+	// Signer is used to sign messages and provides access to the public key.
+	// The signer is expected to do its own hashing, so the full message will be
+	// provided as the message to Sign with a zero opts.HashFunc().
+	crypto.Signer
```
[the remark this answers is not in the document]
asraa: Are you suggesting adding Sign methods in the interface rather than using crypto.Signer? 

I think using crypto.Signer simplifies the interface a bit, because most keys already have those methods (like we can use rsa.GenerateKeyPair) so we don't have to implement a `Sign` method, and just have to embed the crypto.Signer into the type. But I would consider it.

pkg/keys/ed25519.go
asraa: I do need this to be exported so that someone can implement a signer in a separate library, and wanted parity between UnmarshalSigner and MarshalSigner for naming. WDYT?

local_store.go
```diff
@@ -82,12 +94,13 @@ func (m *memoryStore) Commit(consistentSnapshot bool, versions map[string]int, h
...
 }
 
-func (m *memoryStore) GetSigningKeys(role string) ([]sign.Signer, error) {
+func (m *memoryStore) GetSigners(role string) ([]keys.Signer, error) {
 	return m.signers[role], nil
 }
 
-func (m *memoryStore) SavePrivateKey(role string, key *sign.PrivateKey) error {
-	m.signers[role] = append(m.signers[role], key.Signer())
+func (m *memoryStore) SavePrivateKey(role string, key *data.PrivateKey) error {
```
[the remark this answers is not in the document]
asraa: might just be better to make it take in the Signer directly and call it `SaveSigner`.

I don't like `Create` -- because the signer was already created, just represented as the data.PrivateKey.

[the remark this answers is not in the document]
asraa: Renamed to SaveSigner and take in a signer

repo.go
```diff
@@ -323,11 +324,11 @@ func (r *Repo) GenKeyWithExpires(keyRole string, expires time.Time) ([]string, e
 	return key.PublicData().IDs(), nil
 }
 
-func (r *Repo) AddPrivateKey(role string, key *sign.PrivateKey) error {
+func (r *Repo) AddPrivateKey(role string, key keys.Signer) error {
```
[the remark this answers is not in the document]
asraa: Yep! That addressed the other comment as well

data/types.go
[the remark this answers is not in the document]
asraa: if PrivateKey stays, this should be renamed to PublicKey

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+	// This is the public string used as a unique identifier for the verifier instance.
+	Public() string
+
+	// Verify takes a message and signature, all as byte slices,
+	// and determines whether the signature is valid for the given
+	// key and message.
+	Verify(msg, sig []byte) error
+}
+
+type Signer interface {
```
asraa, line 41: add a convenience function to get `IDs()`?

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+	ErrInvalid    = errors.New("tuf: signature verification failed")
+	ErrInvalidKey = errors.New("invalid key")
+)
+
+// A Verifier verifies public key signatures.
+type Verifier interface {
+	// UnmarshalKey takes key data to a working verifier implementation for the key type.
+	// This performs any validation over the data.Key to ensure that the verifier is usable
+	// to verify signatures.
+	UnmarshalKey(key *data.Key) error
```
asraa, line 27: -> UnmarshalPublicKey/ MarshalPublicKey

pkg/keys/keys.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Verify takes a message and signature, all as byte slices,
+	// and determines whether the signature is valid for the given
+	// key and message.
+	Verify(msg, sig []byte) error
+}
+
+type Signer interface {
+	// MarshalSigner returns the private key data.
+	MarshalSigner() (*data.PrivateKey, error)
```
asraa, line 43: MarshalPrivateKey/UnmarshalPrivateKey

pkg/keys/keys.go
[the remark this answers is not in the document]
asraa: AI: Replace with `Sign(msg []byte)`

pkg/keys/rsa_test.go
```diff
@@ -0,0 +1,96 @@
...
+	keyValBytes, _ := json.Marshal(rsaPublic{PublicKey: string(pubBytes)})
+	return &data.Key{
+		Type:       data.KeyTypeRSASSA_PSS_SHA256,
+		Scheme:     data.KeySchemeRSASSA_PSS_SHA256,
+		Algorithms: data.KeyAlgorithms,
+		Value:      keyValBytes,
+	}
+}
+
+func (s rsaSigner) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) ([]byte, error) {
```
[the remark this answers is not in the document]
asraa: AI: move to rsa.go

data/types.go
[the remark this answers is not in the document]
asraa: Discussed offline. This will stay in data.PublicKey

pkg/keys/ed25519.go
asraa: Changed to (Un)MarshalPrivateKey

pkg/keys/keys.go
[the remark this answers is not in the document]
asraa: changed to `SignMessage` -- go doesn't like the conflict with crypto.Signer Sign if you embed one inside your Signer.

pkg/keys/keys.go
asraa: @trishankatdatadog @hosseinsia I haven't yet done this -- it's a tradeoff. You either call `signer.PublicData().IDs()`, or have every signer implementation implement the boiler plate code (since go won't have "base impl methods")

```
func (s someSigner) IDs() []byte {
  return s.PublicData().IDs()
}
```

data/types.go
```diff
@@ -36,17 +38,24 @@ type Signature struct {
...
-	Scheme     string   `json:"scheme"`
-	Algorithms []string `json:"keyid_hash_algorithms,omitempty"`
-	Value      KeyValue `json:"keyval"`
+type PublicKey struct {
+	Type       string          `json:"keytype"`
+	Scheme     string          `json:"scheme"`
+	Algorithms []string        `json:"keyid_hash_algorithms,omitempty"`
+	Value      json.RawMessage `json:"keyval"`
 
 	ids    []string
```
hosseinsia, line 47: I am wondering why id is an array of string, not a string?

data/types_test.go
```diff
@@ -23,23 +23,29 @@ type TypesSuite struct{}
...
+	c.Assert(err, IsNil)
 
-	key := Key{
+	key := PublicKey{
 		Type:   KeyTypeEd25519,
 		Scheme: KeySchemeEd25519,
-		Value:  KeyValue{Public: hexbytes},
+		Value:  keyValBytes,
 	}
 	c.Assert(key.IDs(), DeepEquals, []string{keyid10})
```
hosseinsia, line 42: Can you also please test a case that ids is an array longer than 1.

data/types.go
```diff
@@ -14,11 +14,13 @@ import (
...
+	KeyIDLength                = sha256.Size * 2
+	KeyTypeEd25519             = "ed25519"
+	KeyTypeECDSA_SHA2_P256     = "ecdsa-sha2-nistp256"
+	KeySchemeEd25519           = "ed25519"
+	KeySchemeECDSA_SHA2_P256   = "ecdsa-sha2-nistp256"
+	KeyTypeRSASSA_PSS_SHA256   = "rsa"
+	KeySchemeRSASSA_PSS_SHA256 = "rsassa-pss-sha256"
 )
 
 var (
```
hosseinsia, line 26: KeyAlgorithms -> HashAlgorithms?

data/types.go
```diff
@@ -36,17 +38,24 @@ type Signature struct {
...
+type PrivateKey struct {
+	Type       string          `json:"keytype"`
+	Scheme     string          `json:"scheme,omitempty"`
+	Algorithms []string        `json:"keyid_hash_algorithms,omitempty"`
+	Value      json.RawMessage `json:"keyval"`
+}
+
+func (k *PublicKey) IDs() []string {
 	k.idOnce.Do(func() {
 		data, _ := cjson.Marshal(k)
```
hosseinsia, line 60: Is it safe to ignore this error?

data/types.go
asraa: I have no idea why it's a list -- I thought there was some context behind keys having multiple IDs but that doesn't seem to be the case in the Python implementation either. Should I open an issue to clean this up?

data/types_test.go
asraa: I think we should clean this up per https://github.com/theupdateframework/go-tuf/pull/148#discussion_r717629525
Our `IDs` function always returns a single value, it's hardcoded in `data.PublicKey` to do that.

local_store.go
```diff
@@ -11,10 +11,22 @@ import (
...
+	"github.com/theupdateframework/go-tuf/pkg/keys"
 	"github.com/theupdateframework/go-tuf/util"
 )
 
+func signers(privateKeys []*data.PrivateKey) []keys.Signer {
+	res := make([]keys.Signer, len(privateKeys))
+	for i, k := range privateKeys {
+		signer, err := keys.GetSigner(k)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
asraa: This only appears in a file system store, but I resolved this so that the capacity is len(privateKeys) but the array is initialized with size 0 and we append. So the returned array only contains valid signers

repo.go
```diff
@@ -311,23 +312,23 @@ func (r *Repo) GenKey(role string) ([]string, error) {
 }
 
 func (r *Repo) GenKeyWithExpires(keyRole string, expires time.Time) ([]string, error) {
```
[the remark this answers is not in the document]
asraa: done, but return values in go are a little weird, so leaving this open for you to resolve if that looks good

data/types.go
asraa: You'll never hit this error from deserializing a bad key in metadata since that'll fail already on ingestion when unmarshalling -- so you can assume that `PublicKey` is valid unless somehow a key implementation purposely puts in some invalid data when they implement `PublicData` https://github.com/theupdateframework/go-tuf/blob/736fcc61962c9cece79df91d0e8393e351d0fd8f/pkg/keys/ed25519.go#L120

Testing this is weird, because if it does fail, it'll still produce a hash, just a bad hash -- what do you think about a `panic`?

repo.go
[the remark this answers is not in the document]
hosseinsia: looks good! -- but yah this is also weird.

data/types.go
hosseinsia: panic is good. Something should really be wrong if this fails!

data/types.go
hosseinsia: Yup. An issue does that. -- If I were to guess, there might have been a reason to use multiple hashing algorithms there to diversify crypto, but that is a guess only.

trishankatdatadog: Yeah, this could be related to the old `key_hashing_algorithms` kerfuffle (TLDR: we allowed using SHA2-256 and/or SHA2-512 to compute different, collision-resistant keyids for the same key).

asraa: https://github.com/theupdateframework/go-tuf/issues/158

hosseinsia:
Fantastic work! 
LGTM!

hosseinsia (approved):
LGTM

trishankatdatadog:
Thank you both very much, @asraa and @hosseinsia!

I would strongly encourage @joshuagl to review and approve to ensure there is no collusion 🙂

trishankatdatadog (approved):
Approving in order to unblock. Great work, Asra! 🎉

tamird:
This PR added a dependency on github.com/pkg/errors. I've sent #166 to replace the dependency with standard library functions.

PR #150, opened 2021-09-08
build: update go versions used in CI

joshuagl:
Our CI should be testing recent Go. Arguably we should not worry about
unsupported versions of Go.
According to the Go docs (https://golang.org/doc/devel/release):
"Each major Go release is supported until there are two newer major
releases."

Therefore, with current major Go being 1.17, update our CI to the three
supported verions -- 1.15, 1.16 and 1.17

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


asraa (approved):
Thanks!

zwass:
This looks like a pretty easy to review PR. Can we get a second reviewer to merge it?

mnm678 (approved):

PR #155, opened 2021-09-16
Sort signatures by keys IDs for deterministic signature order

ethan-lowman-dd:
In the course of implementing delegations in the repo writer ([WIP branch](https://github.com/ethan-lowman-dd/go-tuf/pull/1/files)), I had to change how keys are stored internally, which introduced non-determinism in the order of signatures. This caused tests which relied on fixtures to become flaky, as the order of signatures was not consistent. In order to keep support for static fixtures (using deterministic signature schemes like Ed25519), this PR sorts signatures by key ID.


repo.go
```diff
@@ -605,7 +607,10 @@ func (r *Repo) getSigningKeys(name string) ([]sign.Signer, error) {
 		return nil, err
 	}
 	if name == "root" {
-		return signingKeys, nil
+		sorted := make([]sign.Signer, len(signingKeys))
+		copy(sorted, signingKeys)
+		sort.Sort(signer.ByIDs(sorted))
```
trishankatdatadog, line 612: This line is a bit hard to parse. I am sort sorting signer which are already sorted?

repo_test.go
```diff
@@ -543,9 +552,17 @@ func (rs *RepoSuite) TestSign(c *C) {
...
+		// Signatures may be in any order, so must sort key IDs before comparison.
+		wantKeyIDs := append([]string{}, keyIDs...)
+		sort.Strings(wantKeyIDs)
+
+		gotKeyIDs := []string{}
+		for _, sig := range s.Signatures {
+			gotKeyIDs = append(gotKeyIDs, sig.KeyID)
 		}
+
+		c.Assert(sortStrings(keyIDs), DeepEquals, sortStrings(gotKeyIDs))
```
trishankatdatadog, line 565: What are we really testing here? That we sorted strings correctly? So why not just reuse the built-in function for sorting strings in the first place?

rdimitrov:
Apart from #143, it's also worth syncing with @asraa because of #148 as both of the PRs alter/depend on the same `sign` package.

In that sense, maybe it would be better to prioritize which of these should be merged first to avoid the burden of fixing conflicts afterwards?

internal/signer/sort.go
```diff
@@ -0,0 +1,48 @@
...
+	iIDs := make([]string, len(ids))
+	copy(iIDs, ids)
+	sort.Strings(iIDs)
+
+	ids = b[j].IDs()
+	jIDs := make([]string, len(ids))
+	copy(jIDs, ids)
+	sort.Strings(jIDs)
+
+	minLen := len(iIDs)
```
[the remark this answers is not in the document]
ethan-lowman-dd: Go doesn't have a min function for two ints, so these four lines (33-36) are the best we can do.

repo_test.go
ethan-lowman-dd: This tests that keyIDs and gotKeyIDs have the same elements, disregarding order. Previously the test only passed if the key IDs were in the same order.

I had a duplicate sorting of `keyIDs`/`wantKeyIDs`, which was perhaps a cause of confusion -- I've fixed this.

The `sortStrings` helper made a copy of the string slice before sorting (since `sort.Strings` sorts in place), but I removed the helper since it was confusing and only used in this one test.

repo.go
ethan-lowman-dd: Lines 610-611 makes a copy of `signingKeys` called `sorted` (which is not yet sorted until the next line). Line 612 sorts `sorted` in place using the `signer.ByIDs` comparator.

asraa (commented):
Thanks! I'm not opposed to this change, but could you point me to where the testdata needed to be ordered?

I'm seeing a few checks in the interop_test/client_test but they don't seem to require ordering.

Just wondering if it makes more sense to change tests to test equality/expectation without ordering

repo_test.go
```diff
@@ -543,9 +546,18 @@ func (rs *RepoSuite) TestSign(c *C) {
...
+		wantKeyIDs := append([]string{}, keyIDs...)
+		sort.Strings(wantKeyIDs)
+
+		gotKeyIDs := []string{}
+		for _, sig := range s.Signatures {
+			gotKeyIDs = append(gotKeyIDs, sig.KeyID)
 		}
+		sort.Strings(gotKeyIDs)
+
+		c.Assert(wantKeyIDs, DeepEquals, gotKeyIDs)
```
asraa, line 560: client_test approaches this with https://github.com/theupdateframework/go-tuf/blob/f38e930df33905abc5bb7be5fcd0668231807a88/util/util.go#L261

ethan-lowman-dd:
@asraa The tests that are flaky if signature order is nondeterministic are:

1. File hash checks against generated fixtures:
    * https://github.com/theupdateframework/go-tuf/blob/master/client/interop_test.go#L41-L48
2. Checking key IDs in order:
    * https://github.com/theupdateframework/go-tuf/blob/master/repo_test.go#L545-L548
    * https://github.com/theupdateframework/go-tuf/blob/master/repo_test.go#L1181-L1184

I addressed (1) by sorting signatures by key IDs and regenerating these fixtures. I addressed (2) by sorting keys IDs in the tests before making comparisons.

repo_test.go
ethan-lowman-dd: This isn't quite the same as comparing sorted lists, right? For example:
```
a := []string{"AAA", "BBB", "CCC"}
b := []string{"AAA, "BBB", "CCC", "CCC"}

aSet := StringSliceToSet(a) 
bSet := StringSliceToSet(b)

// This evaluates to true, but the slices we're comparing are not equal.
reflect.DeepEqual(aSet, bSet)
```

I think in this case, we would want tests to fail if signatures are erroneously duplicated. Comparing sorted lists accomplishes this, but comparing sets does not.

asraa: good point with duplicates. i'll have to look into that if that test makes sense to me now too

internal/signer/sort.go
```diff
@@ -0,0 +1,48 @@
+package signer
```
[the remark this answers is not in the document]
ethan-lowman-dd: Added tests

repo.go
```diff
@@ -605,7 +607,10 @@ func (r *Repo) getSigningKeys(name string) ([]sign.Signer, error) {
 		return nil, err
```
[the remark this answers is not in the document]
ethan-lowman-dd: With your other comment, this method no longer sorts the signers.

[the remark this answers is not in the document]
ethan-lowman-dd: I updated the PR to sort on usage rather than on retrieval. I'm not too opinionated about which is better.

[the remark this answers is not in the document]
ethan-lowman-dd: As discussed, I reverted this change and renamed the method to `getSortedSigningKeys`.

trishankatdatadog (dismissed):

asraa (dismissed):

hosseinsia (approved):
LGTM!

internal/signer/sort_test.go
```diff
@@ -0,0 +1,71 @@
...
+		value: json.RawMessage(`{"mock": 5}`),
+	}
+
+	s := []keys.Signer{
+		s1, s2, s3, s4, s5,
+	}
+
+	sort.Sort(signer.ByIDs(s))
+
+	sorted := []keys.Signer{
```
hosseinsia, line 64: Sorry I did not realize this part last time. I think the reason for this random order is because the IDs is the hash of the value? This makes this test very unreadable. 
Can you please mock the IDs in the PublicKey to return the exact value (instead of the hash of the value)? 
and in line 59 input a random order of the signers ... in line 65 change the variable to expected and orders to be s1, s2, s3, s4, s5?

hosseinsia: Hmmm I think that can not quite address the issue.

trishankatdatadog (approved):

PR #160, opened 2021-09-29
Improve the CLI experience

### LINKED ISSUE #11: Improve CLI experience
The CLI should output informative messages, for example:

```
$ tuf gen-key root
Created root key with ID 8f782b389ce30f9296d5b850cfd21a2eb55f134ba6af8397dbed7b0fba259228

$ tuf sign root.json
Added 1 signature to root.json
```

rdimitrov:
The following PR adds informative messages about each CLI command upon successful completion. Closes #11 

One thing I'd like to note is the output for `tuf root-keys` previously consisted only of the JSON encoded root keys and now there's a helpful message before that. This might break stuff in case there's some automation built around it, so let me know if it might make sense to revert that change specifically. 

Examples of the resulting output:

```bash
./tuf init
Repository initialized
------------
# Same for root, targets, snapshot and timestamp
./tuf gen-key root
Enter root keys passphrase: 
Repeat root keys passphrase: 
Generated root key with ID 157a50daaa57d68518ea1a47190e9ffd4be1985647f320409fa28d8c07b1ab6a
------------
./tuf revoke-key root a94ce6fd989d04d5d86202a80b320b3ab1ae60d44c693aad41be76ab5e6f568c
Revoked root key with ID a94ce6fd989d04d5d86202a80b320b3ab1ae60d44c693aad41be76ab5e6f568c
------------
./tuf sign root.json
Signed root.json with 1 key(s)
------------
./tuf add bb/cc/aa
Added/staged targets:
* bb/cc/aa
------------
./tuf add
Added/staged targets:
* bb/vv
* bbbb
* bb/cc/aa
* aaaa
------------
./tuf remove bb/vv
Removed targets:
* bb/vv
Added/staged targets:
* aaaa
* bb/cc/aa
* bbbb
------------
./tuf remove --all
Removed targets:
* aaaa
* bb/cc/aa
* bbbb
There are no added/staged targets
------------
./tuf snapshot
Staged snapshot.json metadata with expiration date: 2021-10-06 10:56:49 +0000 UTC
------------
./tuf timestamp
Staged timestamp.json metadata with expiration date: 2021-09-30 10:57:27 +0000 UTC
------------
./tuf set-threshold root 2
The threshold for root role is 2
------------
./tuf get-threshold targets
The threshold for targets role is 1
------------
./tuf commit
Committed successfully
------------
./tuf root-keys
The resulting JSON should be distributed to clients for performing initial updates:

[{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"3e15b2f364cd35f2a2ee3019089e9dace8fa6daaaa2a895ff0ad13f6d365d9b2"}}]
------------
./tuf clean
Removed all staged metadata and target files
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,10 +80,11 @@ Expects a staged, fully signed `targets` metadata file and stages an appropriate
 `snapshot` metadata file. Optionally one can set number of days after which
 the `snapshot` metadata will expire.
 
-#### `tuf timestamp`
+#### `tuf timestamp [--expires=<days>]`
 
 Stages an appropriate `timestamp` metadata file. If a `snapshot` metadata file is staged,
-it must be fully signed.
+it must be fully signed. Optionally one can set number of days after which
+the timestamp metadata will expire.
 
 #### `tuf sign <metadata>`
 
--- a/README.md
+++ b/README.md
@@ -115,8 +115,12 @@ should be distributed to clients for performing initial updates.
 
 #### `tuf set-threshold <role> <threshold>`
 
-Sets the `role` threshold, the required number of keys for signing, to
+Sets the `role` threshold (required number of keys for signing) to
 `threshold`.
+
+#### `tuf get-threshold <role>`
+
+Outputs the `role` threshold (required number of keys for signing).
 
 #### Usage of environment variables
 
--- a/README.md
+++ b/README.md
@@ -115,12 +115,12 @@ should be distributed to clients for performing initial updates.
 
 #### `tuf set-threshold <role> <threshold>`
 
-Sets the `role` threshold (required number of keys for signing) to
+Sets `role`'s threshold (required number of keys for signing) to
 `threshold`.
 
 #### `tuf get-threshold <role>`
 
-Outputs the `role` threshold (required number of keys for signing).
+Outputs `role`'s threshold (required number of keys for signing).
 
 #### Usage of environment variables
 
--- a/cmd/tuf/get_threshold.go
+++ b/cmd/tuf/get_threshold.go
@@ -23,6 +23,6 @@ func cmdGetThreshold(args *docopt.Args, repo *tuf.Repo) error {
 		return err
 	}
 
-	fmt.Println("Got", role, "threshold", threshold)
+	fmt.Println("The threshold for", role, "role is", threshold)
 	return nil
 }
--- a/cmd/tuf/main.go
+++ b/cmd/tuf/main.go
@@ -42,6 +42,7 @@ Commands:
   clean         Remove all staged metadata files
   root-keys     Output a JSON serialized array of root keys to STDOUT
   set-threshold Sets the threshold for a role
+  get-threshold Outputs the threshold for a role
 
 See "tuf help <command>" for more information on a specific command
 `
--- a/cmd/tuf/root_keys.go
+++ b/cmd/tuf/root_keys.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"encoding/json"
-	"os"
+	"fmt"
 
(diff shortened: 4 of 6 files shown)


mnm678:
> I wonder whether we add a `--verbose/-v` option to enable the richer output you've implemented here? Or, maybe default to this new (breaking) behaviour and instead add a `--silent/-s` option to disable the detailed output?

My vote would be for the `-v`, especially for the `tuf root-keys` command, as it's designed to output json that can be used directly by `tuf-client init`.

mnm678 (commented):
I left a few nits, but mostly this looks good. My only concern relates to my comment above about `root-keys`.

rdimitrov:
> > I wonder whether we add a `--verbose/-v` option to enable the richer output you've implemented here? Or, maybe default to this new (breaking) behaviour and instead add a `--silent/-s` option to disable the detailed output?
> 
> My vote would be for the `-v`, especially for the `tuf root-keys` command, as it's designed to output json that can be used directly by `tuf-client init`.

Thanks for your feedback, @mnm678 👍 

I'm leaning towards being verbose by default as it's more common for new users to use it interactively in the beginning (making it more user-friendly but also have the `-s` flag if one wants to automate/pipe the output). Nevertheless, I agree that this could be a breaking change for some, so I don't mind keeping `tuf root-keys` silent by default 👍

As for the other commands, I don't think they can break something, so I guess we can keep their messages verbose by default, right?

trishankatdatadog:
> I'm leaning towards being verbose by default as it's more common for new users to use it interactively in the beginning (making it more user-friendly but also have the `-s` flag if one wants to automate/pipe the output). Nevertheless, I agree that this could be a breaking change for some, so I don't mind keeping `tuf root-keys` silent by default 👍

I don't think it's such a breaking change. We have done #143 which was definitely breaking, and we haven't seen complaints yet. But anyone who depends on this definitely should speak up. I'm in favour of more, not less, explicit error messages by default. New users should not just see something like `tuf: signature verification failed`.

rdimitrov:
> > I'm leaning towards being verbose by default as it's more common for new users to use it interactively in the beginning (making it more user-friendly but also have the `-s` flag if one wants to automate/pipe the output). Nevertheless, I agree that this could be a breaking change for some, so I don't mind keeping `tuf root-keys` silent by default 👍
> 
> I don't think it's such a breaking change. We have done #143 which was definitely breaking, and we haven't seen complaints yet. But anyone who depends on this definitely should speak up. I'm in favour of more, not less, explicit error messages by default. New users should not just see something like `tuf: signature verification failed`.

I think so too 👍 In that sense, can we agree on leaving the more verbose output enabled by default and adding a `-s/--silent` option for commands such as `tuf root-keys`?

rdimitrov:
> Can we call it `-q/--quiet`, that appears to be a more common in UNIX commands.
> 
> To address the concern about command output going directly into another command, we should put informative/diagnostic output on stderr and conventional output (JSON for other commands) on stdout. See https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html

Neat 👍 Will do that.

joshuagl (approved):
LGTM thanks Radoslav! Could we get some review from others, especially our more Go savvy maintainers?

cc @hosseinsia @trishankatdatadog @asraa @mnm678

mnm678 (approved):

tamird:
This PR and various subsequent changes have made it considerably more noisy to use go-tuf as a library. Would it be possible to thread an `io.Writer` through rather than calling `fmt.Print{f,ln}` directly?

PR #166, opened 2021-10-08
Replace pkg/errors with fmt.Errorf

tamird:
Removes the dependency on github.com/pkg/errors.

Error wrapping is part of the standard library since go1.13.

https://go.dev/blog/go1.13-errors


hosseinsia (approved):
Nice catch! --slipped through my eyes, sorry!
Thanks!

Issue #167, opened 2021-10-12
Bug: revoking a key that is used in multiple roles


asraa:
https://github.com/theupdateframework/go-tuf/blob/0d31d49aa6415e59069b9e0ca202d70f9fa4fe1f/repo.go#L466

I believe this code is deleting a key from the root.Keys, which should only be happening if the key is no longer being used by *any* roles, not just the role that the key is being revoked by.

asraa:
Fixed by https://github.com/theupdateframework/go-tuf/pull/168

PR #168, opened 2021-10-20
bugfix: Fix key revocation when key is shared between roles

asraa:

As noted in issue https://github.com/theupdateframework/go-tuf/issues/167, when a key that is used between multiple roles (e.g. roleA and roleB) is revoked from roleA, the key is still deleted from `root.Keys`, so roleB is left in a bad state where its key is not present in root.Keys.

This fixes that by only deleting the key from `root.Keys` if NO role is using the key anymore. Added a test that would previously fail.

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -467,11 +467,22 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
 		return ErrKeyNotFound{keyRole, id}
 	}
 	role.KeyIDs = keyIDs
-
-	for _, keyID := range key.IDs() {
-		delete(root.Keys, keyID)
-	}
 	root.Roles[keyRole] = role
+
+	// Only delete the key from root.Keys if no other role is using that key.
+	delete_key := true
+	for _, role := range root.Roles {
+		for _, keyID := range role.KeyIDs {
+			if key.ContainsID(keyID) {
+				delete_key = false
+			}
+		}
+	}
+	if delete_key {
+		for _, keyID := range key.IDs() {
+			delete(root.Keys, keyID)
+		}
+	}
 	root.Expires = expires.Round(time.Second)
 	if _, ok := r.versionUpdated["root.json"]; !ok {
 		root.Version++
--- a/repo.go
+++ b/repo.go
@@ -469,16 +469,17 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
 	role.KeyIDs = keyIDs
 	root.Roles[keyRole] = role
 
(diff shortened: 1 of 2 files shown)


mnm678 (dismissed):

repo.go
```diff
@@ -467,11 +467,22 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
 		return ErrKeyNotFound{keyRole, id}
```
hosseinsia, line 467: I understand this might be a little bit unrelated to this PR. 

But, the logic of the previous loop as well as line 466 is unreadable -- basically we are creating a new slice which excludes all the KeyID s related to the key we are removing. Why not deleting them from the list? -- and I do not think performance really matters here. 

I am open to other ideas but the current state unreadable.

repo.go
```diff
@@ -467,11 +467,22 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
 	}
 	role.KeyIDs = keyIDs
+	root.Roles[keyRole] = role
 
-	for _, keyID := range key.IDs() {
-		delete(root.Keys, keyID)
+	// Only delete the key from root.Keys if no other role is using that key.
+	delete_key := true
```
hosseinsia, line 473: nit: please consider renaming delete_key-> key_in_use to make the logic more readable.

repo_test.go
```diff
@@ -563,6 +563,64 @@ func (rs *RepoSuite) TestRevokeKey(c *C) {
...
+
+	// check root was updated
+	root, err = r.root()
+	c.Assert(err, IsNil)
+	c.Assert(root.Roles, NotNil)
+	c.Assert(root.Roles, HasLen, 4)
+	c.Assert(root.Keys, NotNil)
+	// the shared root/targets signer should still be present in root keys
+	rs.assertNumUniqueKeys(c, root, "root", 2)
+	rs.assertNumUniqueKeys(c, root, "targets", 1)
```
hosseinsia, line 613: I am thinking whether checking the expected keys is more accurate than checking the count of them?

hosseinsia (commented):
Overall looks good. 
Added some comments.

repo_test.go
asraa: done!

repo.go
asraa: oh that's a good idea. it definitely took me a very careful read to understand the flow.

trishankatdatadog:
> ping on this for merge?

Needs 1 more review... @mnm678 can you 💬 and/or ✅ ?

mnm678 (approved):

Issue #165, opened 2021-10-07
Add support for signing non-local targets


mnm678:
Add the ability for Targets metadata to sign a target without needing to load the local target file. To do so, the repository could take the digest, length, and any custom metadata and create the Targets metadata for this information without needing to read a file from disk.

This would allow Targets metadata to sign OCI artifacts using the digest and size provided by registry APIs.

PR #170, opened 2021-10-22
Add digest to targets metadata directly

mnm678:
Allows users of go-tuf to sign oci images or other non-local targets by directly providing the hash and length of these artifacts. Solves #165

cc @ethan-lowman-dd @trishankatdatadog @sudo-bmitch

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -851,7 +851,7 @@ func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, pat
 		return err
 	}
 
-	if err := util.TargetFileMetaEqual(data.TargetFileMeta{actual}, localMeta); err != nil {
+	if err := util.TargetFileMetaEqual(data.TargetFileMeta{FileMeta: actual}, localMeta); err != nil {
 		if e, ok := err.(util.ErrWrongLength); ok {
 			return ErrWrongSize{path, e.Actual, e.Expected}
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -842,7 +842,7 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 	return nil
 }
 
-func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, path string) (err error) {
+func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, path string) error {
 	localMeta, ok := c.targets[path]
 	if !ok {
 		return ErrUnknownTarget{Name: path, SnapshotVersion: c.snapshotVer}
--- a/client/client.go
+++ b/client/client.go
@@ -849,6 +849,7 @@ func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, pat
 	}
 
 	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
+	var err error
 	actual.Hashes[digestAlg], err = hex.DecodeString(digest)
 	if err != nil {
 		return err
--- a/client/client.go
+++ b/client/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"strings"
 
 	"github.com/theupdateframework/go-tuf/data"
 	"github.com/theupdateframework/go-tuf/util"
@@ -842,8 +843,12 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 	return nil
 }
 
-func (c *Client) VerifyDigest(digest string, length int64) (err error) {
-	localMeta, _ := c.targets[digest]
+func (c *Client) VerifyDigest(digest string, length int64, path string) (err error) {
+	targetName := path
+	if path == "" {
+		targetName = digest
+	}
+	localMeta, _ := c.targets[targetName]
 
 	hashes := make([]string, 1)
 	hashes[0] = strings.Split(digest, ":")[1]
--- a/client/client.go
+++ b/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"bytes"
+	"encoding/hex"
 	"encoding/json"
 	"io"
 	"io/ioutil"
@@ -836,6 +837,25 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 			return ErrWrongSize{name, e.Actual, e.Expected}
 		}
 		return ErrDownloadFailed{name, err}
+	}
+
+	return nil
+}
+
+func (c *Client) VerifyDigest(digest string, length int64) (err error) {
+	localMeta, _ := c.targets[digest]
+
+	hashes := make([]string, 1)
+	hashes[0] = strings.Split(digest, ":")[1]
+
+	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, len(hashes))}
+	actual.Hashes["sha256"], err = hex.DecodeString(hashes[0])
+
+	if err := util.TargetFileMetaEqual(data.TargetFileMeta{actual}, localMeta); err != nil {
+		if e, ok := err.(util.ErrWrongLength); ok {
+			return ErrWrongSize{digest, e.Actual, e.Expected}
+		}
+		return ErrDownloadFailed{digest, err}
 	}
 
 	return nil
--- a/client/client.go
+++ b/client/client.go
@@ -843,7 +843,10 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 }
 
 func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, path string) (err error) {
-	localMeta, _ := c.targets[path]
+	localMeta, ok := c.targets[path]
+	if !ok {
+		return ErrUnknownTarget{Name: path, SnapshotVersion: c.snapshotVer}
+	}
 
 	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
 	actual.Hashes[digestAlg], err = hex.DecodeString(digest)
--- a/client/client.go
+++ b/client/client.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
-	"strings"
 
 	"github.com/theupdateframework/go-tuf/data"
 	"github.com/theupdateframework/go-tuf/util"
@@ -843,24 +842,17 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 	return nil
 }
 
-func (c *Client) VerifyDigest(digest string, length int64, path string) (err error) {
-	targetName := path
-	if path == "" {
-		targetName = digest
-	}
-	localMeta, _ := c.targets[targetName]
+func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, path string) (err error) {
+	localMeta, _ := c.targets[path]
 
-	hashes := make([]string, 1)
-	hashes[0] = strings.Split(digest, ":")[1]
-
-	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, len(hashes))}
-	actual.Hashes["sha256"], err = hex.DecodeString(hashes[0])
+	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
+	actual.Hashes[digestAlg], err = hex.DecodeString(digest)
 
 	if err := util.TargetFileMetaEqual(data.TargetFileMeta{actual}, localMeta); err != nil {
 		if e, ok := err.(util.ErrWrongLength); ok {
-			return ErrWrongSize{digest, e.Actual, e.Expected}
+			return ErrWrongSize{path, e.Actual, e.Expected}
 		}
-		return ErrDownloadFailed{digest, err}
+		return ErrDownloadFailed{path, err}
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


joshuagl:
The build is failing in `go fmt`, which only runs on Linux (i.e. Ubuntu), see https://github.com/theupdateframework/go-tuf/blob/fc0190d925a5f43942c51c5e519b8fce3b025610/.github/workflows/build.yml#L24

If I run `go fmt` locally on a clone of this branch it reformats repo_test.go thus:
```diff
diff --git a/repo_test.go b/repo_test.go
index bc55210..c764162 100644
--- a/repo_test.go
+++ b/repo_test.go
@@ -1684,7 +1684,7 @@ func (rs *RepoSuite) TestBadAddOrUpdateSignatures(c *C) {
        checkSigIDs("root.json")
 }

-func (rs *RepoSuite) TestSignDigest(c *C){
+func (rs *RepoSuite) TestSignDigest(c *C) {
        files := map[string][]byte{"foo.txt": []byte("foo")}
        local := MemoryStore(make(map[string]json.RawMessage), files)
        r, err := NewRepo(local)
@@ -1708,7 +1708,7 @@ func (rs *RepoSuite) TestSignDigest(c *C){
        c.Assert(err, IsNil)

        targets, err := r.targets()
-               c.Assert(err, IsNil)
+       c.Assert(err, IsNil)
        c.Assert(targets.Targets["sha256:bc11b176a293bb341a0f2d0d226f52e7fcebd186a7c4dfca5fc64f305f06b94c"].FileMeta.Length, Equals, size)
        c.Assert(targets.Targets["sha256:bc11b176a293bb341a0f2d0d226f52e7fcebd186a7c4dfca5fc64f305f06b94c"].FileMeta.Hashes["sha256"], DeepEquals, hex_digest_bytes)
```

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+	}
+
+	t.Targets[digest] = data.TargetFileMeta{meta}
+
+	t.Expires = expires.Round(time.Second)
+	if _, ok := r.versionUpdated["targets.json"]; !ok {
+		t.Version++
+		r.versionUpdated["targets.json"] = struct{}{}
+	}
+	return r.setMeta("targets.json", t)
```
ethan-lowman-dd, line 704: Since this is identical to how `AddTargetsWithExpires` writes targets metadata, can we factor this part out into its own function?

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+		return err
+	}
+	hashes := make([]string, 1)
+	split_digest := strings.Split(digest, ":")
+	hashes[0] = split_digest[1]
+
+	meta := data.FileMeta{Length: length, Hashes: make(data.Hashes, len(hashes))}
+	meta.Hashes[split_digest[0]], err = hex.DecodeString(hashes[0])
+
+	path := digest
```
ethan-lowman-dd, line 689: Should the name of the target path be customizable? I can imagine use cases where a remote artifact name is associated with a hash, but not addressed directly hash.

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+func (r *Repo) AddDigestTargets(digest string, length int64, custom json.RawMessage) error {
+	expires := data.DefaultExpires("targets")
+
+	t, err := r.targets()
+	if err != nil {
+		return err
+	}
+	hashes := make([]string, 1)
+	split_digest := strings.Split(digest, ":")
+	hashes[0] = split_digest[1]
```
ethan-lowman-dd, line 684: We probably should return an error here if `digest` is not the proper format, otherwise this could panic.

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
 
+func (r *Repo) AddDigestTargets(digest string, length int64, custom json.RawMessage) error {
+	expires := data.DefaultExpires("targets")
+
+	t, err := r.targets()
+	if err != nil {
+		return err
+	}
+	hashes := make([]string, 1)
+	split_digest := strings.Split(digest, ":")
```
ethan-lowman-dd, line 683: nit: this repo use camelCase pretty consistently.

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
 }
 
+func (r *Repo) AddDigestTargets(digest string, length int64, custom json.RawMessage) error {
+	expires := data.DefaultExpires("targets")
+
+	t, err := r.targets()
+	if err != nil {
+		return err
+	}
+	hashes := make([]string, 1)
```
ethan-lowman-dd, line 682: Since we never actually use `hashes` as slice, only access its zeroth element and its length, could we just use a `hash` `string` variable?

repo.go
```diff
@@ -671,6 +672,38 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+	split_digest := strings.Split(digest, ":")
+	hashes[0] = split_digest[1]
+
+	meta := data.FileMeta{Length: length, Hashes: make(data.Hashes, len(hashes))}
+	meta.Hashes[split_digest[0]], err = hex.DecodeString(hashes[0])
```
ethan-lowman-dd, line 687: Can we give names to `split_digest[0]` and `split_digest[1]`

ethan-lowman-dd: Additionally, `err` is unchecked.

repo.go
mnm678: Yes, I was originally thinking that this could support multiple hashes, but the rest of the code doesn't yet, so I'll make this a string.

mnm678:
Thank for the review @ethan-lowman-dd!

mnm678:
I added the client verification and fixed the test failures, so I think this is ready for another review @ethan-lowman-dd or @trishankatdatadog

trishankatdatadog:
> I added the client verification and fixed the test failures, so I think this is ready for another review @ethan-lowman-dd or @trishankatdatadog

Awesome! Ethan, would you please help? 🙂

repo.go
```diff
@@ -695,6 +696,41 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+	t, err := r.targets()
+	if err != nil {
+		return err
+	}
+	splitDigest := strings.Split(digest, ":")
+	if len(splitDigest) != 2 {
+		return fmt.Errorf("incorrect format for digest: %s", digest)
+	}
+	hash := splitDigest[1]
+	hashAlg := splitDigest[0]
```
ethan-lowman-dd, line 711: The `sha256:ABC123` format is pretty specific to OCI refs, right? What do you think about passing the hash algorithm and hash as separate arguments, instead of requiring a special format for `digest`?

repo.go
```diff
@@ -695,6 +696,41 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
 	return r.AddTargetsWithExpires(paths, custom, data.DefaultExpires("targets"))
 }
 
+func (r *Repo) AddDigestTargets(digest string, length int64, custom json.RawMessage, path string) error {
+	expires := data.DefaultExpires("targets")
+
+	t, err := r.targets()
```
ethan-lowman-dd, line 702: For a future PR:
```suggestion
	// TODO: support delegated targets
	t, err := r.targets()
```

repo.go
```diff
@@ -695,6 +696,41 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+
+	meta := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
+	meta.Hashes[hashAlg], err = hex.DecodeString(hash)
+	if err != nil {
+		return err
+	}
+
+	if path == "" {
+		path = digest
+	}
```
ethan-lowman-dd, line 721: Should this default repo design be baked into the the go-tuf library? For callers of the library, it's just as easy to pass `digest` instead of an empty string `""`, while being more explicit/readable, and then the library doesn't need to maintain this API promise.

repo.go
```diff
@@ -734,13 +770,17 @@ func (r *Repo) AddTargetsWithExpires(paths []string, custom json.RawMessage, exp
 	}); err != nil {
 		return err
 	}
+	return r.WriteTargetWithExpires(t, expires)
+}
+
+func (r *Repo) WriteTargetWithExpires(t *data.Targets, expires time.Time) error {
```
ethan-lowman-dd, line 776: Could we make this private, since it lacks the user-friendly validation of the higher-level methods?

repo.go
mnm678: Yes, this depends on the order of this and #175, but would be good to support

repo.go
```diff
@@ -695,6 +696,32 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+	meta.Hashes[digestAlg], err = hex.DecodeString(digest)
+	if err != nil {
+		return err
+	}
+
+	if len(custom) > 0 {
+		meta.Custom = &custom
+	} else if t, ok := t.Targets[path]; ok {
+		meta.Custom = t.Custom
+	}
```
ethan-lowman-dd, line 718: This merging logic makes it impossible to clear the `FileMeta.Custom` field by setting it to `nil`. However it does allow setting `Custom` to `{}`. What do you think about making the `custom` argument a `*json.RawMessage` and just setting `meta.Custom = custom` rather than merging, to make all operations possible?

repo.go
```diff
@@ -695,6 +696,32 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+
+	t.Targets[path] = data.TargetFileMeta{meta}
```
ethan-lowman-dd, line 720: This fails go vet with `github.com/theupdateframework/go-tuf/data.TargetFileMeta composite literal uses unkeyed fields`

repo.go
ethan-lowman-dd: I just noticed that this is a shortcoming of all the existing AddTargets functions, so no need to address here.

client/client.go
```diff
@@ -888,6 +889,22 @@ func (c *Client) Download(name string, dest Destination) (err error) {
 	return nil
 }
 
+func (c *Client) VerifyDigest(digest string, digestAlg string, length int64, path string) (err error) {
+	localMeta, _ := c.targets[path]
+
+	actual := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
+	actual.Hashes[digestAlg], err = hex.DecodeString(digest)
```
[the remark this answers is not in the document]
mnm678: fixed below

ethan-lowman-dd (approved):

repo.go
```diff
@@ -695,6 +696,32 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
 	return r.AddTargetsWithExpires(paths, custom, data.DefaultExpires("targets"))
 }
 
+func (r *Repo) AddDigestTargets(digest string, digestAlg string, length int64, custom json.RawMessage, path string) error {
+	expires := data.DefaultExpires("targets")
+
+	// TODO: support delegated targets
```
trishankatdatadog, line 702: pls open an issue for this

repo.go
```diff
@@ -695,6 +696,32 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
...
+
+	meta := data.FileMeta{Length: length, Hashes: make(data.Hashes, 1)}
+	meta.Hashes[digestAlg], err = hex.DecodeString(digest)
+	if err != nil {
+		return err
+	}
+
+	if len(custom) > 0 {
+		meta.Custom = &custom
+	} else if t, ok := t.Targets[path]; ok {
```
trishankatdatadog, line 716: what does this do? not obvious from first read

repo.go
```diff
@@ -695,6 +696,32 @@ func (r *Repo) AddTargets(paths []string, custom json.RawMessage) error {
 	return r.AddTargetsWithExpires(paths, custom, data.DefaultExpires("targets"))
 }
 
+func (r *Repo) AddDigestTargets(digest string, digestAlg string, length int64, custom json.RawMessage, path string) error {
```
trishankatdatadog, line 699: The name is a bit poor. Do you mean something like `AddTargetToTargets` (not great either, but you get the idea)?

trishankatdatadog (changes requested):

repo.go
mnm678: comment added

repo.go
mnm678: Changed to `AddTargetsWithDigest`, better?

repo.go
mnm678: Opened #184

ethan-lowman-dd (dismissed):

mnm678:
ping @trishankatdatadog you're marked as requesting changes. Do the latest commits address those?

trishankatdatadog:
> ping @trishankatdatadog you're marked as requesting changes. Do the latest commits address those?

Will review today during our meeting. Thanks!

mnm678:
Thanks @asraa. I addressed your comments and rebased for some of the delegations pr changes.

asraa (approved):

trishankatdatadog (approved):

PR #172, opened 2021-11-07
Fix JSON struct tag for FileMeta

zwass:
This looks like it was typoed.


trishankatdatadog (approved):
How did go not catch this before? Surprising.

asraa (approved):
Thanks for the catch!

Huh, yeah. I wonder if we never caught it because we always had length/hashes added.

PR #171, opened 2021-11-04
WIP: Support delegated targets roles in repo writer

ethan-lowman-dd:

PR #175, opened 2021-11-16
feat: Support delegated targets roles in repo writer

mnm678:
This pr is a takeover of #171, which I don't seem to be able to edit directly.

cc @ethan-lowman-dd @asraa @trishankatdatadog @hosseinsia

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -58,6 +58,7 @@ func isDecodeFailedWithErrRoleThreshold(err error) bool {
 	return isErrRoleThreshold(e.Err)
 }
 
+//lint:ignore U1000 unused
 func isErrRoleThreshold(err error) bool {
 	_, ok := err.(verify.ErrRoleThreshold)
 	return ok
--- a/client/errors.go
+++ b/client/errors.go
@@ -3,8 +3,6 @@ package client
 import (
 	"errors"
 	"fmt"
-
-	"github.com/theupdateframework/go-tuf/verify"
 )
 
 var (
@@ -47,21 +45,6 @@ type ErrMaxDelegations struct {
 
 func (e ErrMaxDelegations) Error() string {
 	return fmt.Sprintf("tuf: max delegation of %d reached searching for %s with snapshot version %d", e.MaxDelegations, e.Target, e.SnapshotVersion)
-}
-
-//lint:ignore U1000 unused
-func isDecodeFailedWithErrRoleThreshold(err error) bool {
-	e, ok := err.(ErrDecodeFailed)
-	if !ok {
-		return false
-	}
-	return isErrRoleThreshold(e.Err)
-}
-
-//lint:ignore U1000 unused
-func isErrRoleThreshold(err error) bool {
-	_, ok := err.(verify.ErrRoleThreshold)
-	return ok
 }
 
 type ErrNotFound struct {
--- a/local_store.go
+++ b/local_store.go
@@ -107,7 +107,6 @@ func (m *memoryStore) FileIsStaged(name string) bool {
 func (m *memoryStore) WalkStagedTargets(paths []string, targetsFn TargetsWalkFunc) error {
 	if len(paths) == 0 {
 		for path, data := range m.files {
-			fmt.Println("pd", path, data)
 			if err := targetsFn(path, bytes.NewReader(data)); err != nil {
 				return err
 			}
--- a/local_store.go
+++ b/local_store.go
@@ -42,6 +42,8 @@ type LocalStore interface {
 	Commit(bool, map[string]int64, map[string]data.Hashes) error
 
 	// GetSigners return a list of signers for a role.
+	// This may include revoked keys, so the signers should not
+	// be used without filtering.
 	GetSigners(role string) ([]keys.Signer, error)
 
 	// SaveSigner adds a signer to a role.
--- a/local_store.go
+++ b/local_store.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/theupdateframework/go-tuf/data"
 	"github.com/theupdateframework/go-tuf/encrypted"
-	"github.com/theupdateframework/go-tuf/internal/roles"
 	"github.com/theupdateframework/go-tuf/internal/sets"
 	"github.com/theupdateframework/go-tuf/pkg/keys"
 	"github.com/theupdateframework/go-tuf/util"
@@ -222,8 +221,7 @@ func (f *fileSystemStore) stagedDir() string {
 }
 
 func isMetaFile(e os.DirEntry) (bool, error) {
-	name := e.Name()
-	if e.IsDir() || !(filepath.Ext(name) == ".json" && roles.IsTopLevelManifest(name)) {
+	if e.IsDir() || filepath.Ext(e.Name()) != ".json" {
 		return false, nil
 	}
 
--- a/local_store.go
+++ b/local_store.go
@@ -107,6 +107,7 @@ func (m *memoryStore) FileIsStaged(name string) bool {
 func (m *memoryStore) WalkStagedTargets(paths []string, targetsFn TargetsWalkFunc) error {
 	if len(paths) == 0 {
 		for path, data := range m.files {
+			fmt.Println("pd", path, data)
 			if err := targetsFn(path, bytes.NewReader(data)); err != nil {
 				return err
 			}
--- a/repo.go
+++ b/repo.go
@@ -674,33 +674,52 @@ func (r *Repo) jsonMarshal(v interface{}) ([]byte, error) {
 	return json.MarshalIndent(v, r.prefix, r.indent)
 }
 
-func (r *Repo) setMeta(roleFilename string, meta interface{}) error {
-	role := strings.TrimSuffix(roleFilename, ".json")
-
+func (r *Repo) dbsForRole(role string) ([]*verify.DB, error) {
 	dbs := []*verify.DB{}
+
 	if roles.IsTopLevelRole(role) {
 		db, err := r.topLevelKeysDB()
 		if err != nil {
-			return err
+			return nil, err
 		}
 		dbs = append(dbs, db)
 	} else {
 		ddbs, err := r.delegatorDBs(role)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		dbs = append(dbs, ddbs...)
+	}
+
+	return dbs, nil
+}
+
+func (r *Repo) signersForRole(role string) ([]keys.Signer, error) {
+	dbs, err := r.dbsForRole(role)
+	if err != nil {
+		return nil, err
 	}
 
 	signers := []keys.Signer{}
 	for _, db := range dbs {
 		ss, err := r.getSignersInDB(role, db)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		signers = append(signers, ss...)
+	}
+
+	return signers, nil
+}
+
+func (r *Repo) setMeta(roleFilename string, meta interface{}) error {
+	role := strings.TrimSuffix(roleFilename, ".json")
+
+	signers, err := r.signersForRole(role)
+	if err != nil {
+		return err
 	}
 
 	s, err := sign.Marshal(meta, signers...)
@@ -717,20 +736,13 @@ func (r *Repo) setMeta(roleFilename string, meta interface{}) error {
 
 func (r *Repo) Sign(roleFilename string) error {
 	role := strings.TrimSuffix(roleFilename, ".json")
-	if !roles.IsTopLevelRole(role) {
-		return ErrInvalidRole{role, "only signing top-level metadata supported"}
-	}
 
 	s, err := r.SignedMeta(roleFilename)
 	if err != nil {
 		return err
 	}
 
-	db, err := r.topLevelKeysDB()
-	if err != nil {
-		return err
-	}
-	keys, err := r.getSignersInDB(role, db)
+	keys, err := r.signersForRole(role)
 	if err != nil {
 		return err
 	}
@@ -757,20 +769,28 @@ func (r *Repo) Sign(roleFilename string) error {
 // The name must be a valid metadata file name, like root.json.
 func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signature) error {
 	role := strings.TrimSuffix(roleFilename, ".json")
-	if !roles.IsTopLevelRole(role) {
-		return ErrInvalidRole{role, "only signing top-level metadata supported"}
-	}
 
 	// Check key ID is in valid for the role.
-	db, err := r.topLevelKeysDB()
+	dbs, err := r.dbsForRole(role)
 	if err != nil {
 		return err
 	}
-	roleData := db.GetRole(role)
-	if roleData == nil {
-		return ErrInvalidRole{role, "role missing from top-level keys"}
+
+	if len(dbs) == 0 {
+		return ErrInvalidRole{role, "no trusted keys for role"}
 	}
-	if !roleData.ValidKey(signature.KeyID) {
+
+	keyInDB := false
+	for _, db := range dbs {
+		roleData := db.GetRole(role)
+		if roleData == nil {
+			return ErrInvalidRole{role, "role is not in verifier DB"}
+		}
+		if roleData.ValidKey(signature.KeyID) {
+			keyInDB = true
+		}
+	}
+	if !keyInDB {
 		return verify.ErrInvalidKey
 	}
 
@@ -791,9 +811,11 @@ func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signatur
 
 	// Check signature on signed meta. Ignore threshold errors as this may not be fully
 	// signed.
-	if err := db.VerifySignatures(s, role); err != nil {
-		if _, ok := err.(verify.ErrRoleThreshold); !ok {
-			return err
+	for _, db := range dbs {
+		if err := db.VerifySignatures(s, role); err != nil {
+			if _, ok := err.(verify.ErrRoleThreshold); !ok {
+				return err
+			}
 		}
 	}
 
@@ -1476,18 +1498,9 @@ func (r *Repo) verifySignatures(metaFilename string) error {
 
 	role := strings.TrimSuffix(metaFilename, ".json")
 
-	dbs := []*verify.DB{}
-	if roles.IsTopLevelRole(role) {
-		db, err := r.topLevelKeysDB()
-		if err != nil {
-			return err
-		}
-		dbs = append(dbs, db)
-	} else {
-		dbs, err = r.delegatorDBs(role)
-		if err != nil {
-			return err
-		}
+	dbs, err := r.dbsForRole(role)
+	if err != nil {
+		return err
 	}
 
 	for _, db := range dbs {
--- a/repo.go
+++ b/repo.go
@@ -80,12 +80,12 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	root.ConsistentSnapshot = consistentSnapshot
 	// Set root version to 1 for a new root.
 	root.Version = 1
-	if err = r.setTopLevelMeta("root.json", root); err != nil {
+	if err = r.setMeta("root.json", root); err != nil {
 		return err
 	}
 
 	t.Version = 1
-	if err = r.setTopLevelMeta("targets.json", t); err != nil {
+	if err = r.setMeta("targets.json", t); err != nil {
 		return err
 	}
 
@@ -200,7 +200,7 @@ func (r *Repo) SetThreshold(keyRole string, t int) error {
 	if !r.local.FileIsStaged("root.json") {
 		root.Version++
 	}
-	return r.setTopLevelMeta("root.json", root)
+	return r.setMeta("root.json", root)
 }
 
 func (r *Repo) Targets() (data.TargetFiles, error) {
@@ -217,7 +217,7 @@ func (r *Repo) SetTargetsVersion(v int64) error {
 		return err
 	}
 	t.Version = v
-	return r.setTopLevelMeta("targets.json", t)
+	return r.setMeta("targets.json", t)
 }
 
 func (r *Repo) TargetsVersion() (int64, error) {
@@ -234,7 +234,7 @@ func (r *Repo) SetTimestampVersion(v int64) error {
 		return err
 	}
 	ts.Version = v
-	return r.setTopLevelMeta("timestamp.json", ts)
+	return r.setMeta("timestamp.json", ts)
 }
 
 func (r *Repo) TimestampVersion() (int64, error) {
@@ -252,7 +252,7 @@ func (r *Repo) SetSnapshotVersion(v int64) error {
 	}
 
 	s.Version = v
-	return r.setTopLevelMeta("snapshot.json", s)
+	return r.setMeta("snapshot.json", s)
 }
 
 func (r *Repo) SnapshotVersion() (int64, error) {
@@ -412,7 +412,7 @@ func (r *Repo) AddVerificationKeyWithExpiration(keyRole string, pk *data.PublicK
 		root.Version++
 	}
 
-	return r.setTopLevelMeta("root.json", root)
+	return r.setMeta("root.json", root)
 }
 
 func validExpires(expires time.Time) bool {
@@ -523,7 +523,7 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
 		root.Version++
 	}
 
-	err = r.setTopLevelMeta("root.json", root)
+	err = r.setMeta("root.json", root)
 	if err == nil {
 		fmt.Println("Revoked", keyRole, "key with ID", id, "in root metadata")
 	}
@@ -541,6 +541,8 @@ func (r *Repo) AddTargetsDelegation(delegator string, role data.DelegatedRole, k
 // should have corresponding Key entries in the keys argument. New metadata is
 // written with the given expiration time.
 func (r *Repo) AddTargetsDelegationWithExpires(delegator string, role data.DelegatedRole, keys []*data.PublicKey, expires time.Time) error {
+	expires = expires.Round(time.Second)
+
 	t, err := r.targets(delegator)
 	if err != nil {
 		return fmt.Errorf("error getting delegator (%q) metadata: %w", delegator, err)
@@ -567,18 +569,14 @@ func (r *Repo) AddTargetsDelegationWithExpires(delegator string, role data.Deleg
 		}
 	}
 	t.Delegations.Roles = append(t.Delegations.Roles, role)
-	t.Expires = expires.Round(time.Second)
+	t.Expires = expires
 
 	delegatorFile := delegator + ".json"
 	if !r.local.FileIsStaged(delegatorFile) {
 		t.Version++
 	}
 
-	delegatorSigners, err := r.local.GetSigners(delegator)
-	if err != nil {
-		return fmt.Errorf("error getting signers for delegator %v: %w", delegator, err)
-	}
-	err = r.setMetaWithSigners(delegatorFile, t, delegatorSigners)
+	err = r.setMeta(delegatorFile, t)
 	if err != nil {
 		return fmt.Errorf("error setting metadata for %q: %w", delegatorFile, err)
 	}
@@ -588,17 +586,14 @@ func (r *Repo) AddTargetsDelegationWithExpires(delegator string, role data.Deleg
 	if err != nil {
 		return fmt.Errorf("error getting delegatee (%q) metadata: %w", delegatee, err)
 	}
+	dt.Expires = expires
 
 	delegateeFile := delegatee + ".json"
 	if !r.local.FileIsStaged(delegateeFile) {
 		dt.Version++
 	}
 
-	delegateeSigners, err := r.local.GetSigners(delegatee)
-	if err != nil {
-		return fmt.Errorf("error getting signers for delegatee %v: %w", delegatee, err)
-	}
-	err = r.setMetaWithSigners(delegateeFile, dt, delegateeSigners)
+	err = r.setMeta(delegateeFile, dt)
 	if err != nil {
 		return fmt.Errorf("error setting metadata for %q: %w", delegateeFile, err)
 	}
@@ -665,11 +660,7 @@ func (r *Repo) ResetTargetsDelegationsWithExpires(delegator string, expires time
 		t.Version++
 	}
 
-	delegatorSigners, err := r.local.GetSigners(delegator)
-	if err != nil {
-		return fmt.Errorf("error getting signers for delegator %v: %w", delegator, err)
-	}
-	err = r.setMetaWithSigners(delegatorFile, t, delegatorSigners)
+	err = r.setMeta(delegatorFile, t)
 	if err != nil {
 		return fmt.Errorf("error setting metadata for %q: %w", delegatorFile, err)
 	}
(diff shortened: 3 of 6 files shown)


data/types.go
```diff
@@ -82,12 +82,13 @@ func DefaultExpires(role string) time.Time {
...
 	case "root":
 		t = time.Now().AddDate(1, 0, 0)
-	case "targets":
-		t = time.Now().AddDate(0, 3, 0)
 	case "snapshot":
 		t = time.Now().AddDate(0, 0, 7)
 	case "timestamp":
 		t = time.Now().AddDate(0, 0, 1)
+	default:
+		// Target
```
asraa, line 90: nit: targets and delegated roles

local_store.go
```diff
@@ -317,18 +401,64 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Cache the signers.
+		for _, keyID := range signer.PublicData().IDs() {
+			f.keyIDsForRole[role] = append(f.keyIDsForRole[role], keyID)
+			f.signerForKeyID[keyID] = signer
+		}
+		//TODO: don't add duplicate keyids?
```
asraa, line 429: would loadPrivateKeys possibly return duped keys?

local_store.go
```diff
@@ -378,7 +508,23 @@ func (f *fileSystemStore) SaveSigner(role string, signer keys.Signer) error {
...
+			return err
+		}
+
+		keyIDs := signer.PublicData().IDs()
+
+		for _, keyID := range keyIDs {
+			f.signerForKeyID[keyID] = signer
+		}
+
+		mergedKeyIDs := sets.DeduplicateStrings(append(f.keyIDsForRole[role], keyIDs...))
```
asraa, line 524: i think these lines can move outside of the for loop so you only dedupe once as long as you hold a local keyIDsForRole inside the for loop

asraa (commented):
Just started doing the review, just got caught up a little in the key generation not being compatible with delegations -- right now i think it is, except for the hashed bin delegations? (i think it should end up saving the keys in a role file, possibly duplicating them)

ethan-lowman-dd:
I haven't looked at the parts of this I wrote in a while, so I'm unfortunately a bit rusty on what's unfinished, but this is the testing script I was using (haven't converted it to unit tests yet): https://gist.github.com/ethan-lowman-dd/cbc403cd0e68215703f903b7b8777c9c

Running this script on the current branch, regular delegated role signing seems to work. Key management for hash bin delegated signing is not working -- the `leaf` signer is not being fetched properly, so no signature is written.

There's also a potential issue with delegation traversal, but I need to read the spec closer to verify -- in the test, A.txt can not be signed since there is no explicit delegation for its path, but I think the top level targets role should sign in that case.

Additionally, I think it's still unaddressed how hash bin delegated keys would be loaded after the delegation is first written. In the test script, we generate the `leaf` key before we use it, so it's cached in memory by key ID. However, since the hash bins role name `leaf` doesn't appear anywhere in TUF metadata (its metadata is named `bins_XX.json`, not `leaf.json`), when we load the repo a second time, we have no handle to the `leaf` key on disk. One way to fix this would be to require clients to manually load delegated keys before signing with them, but that's perhaps not so friendly. I would lean towards something  like storing keys on disk by their key ID rather than their role name, but that's a pretty big breaking change. Whatever the solution, it should ideally work for remote HSM signers too.

ethan-lowman-dd:
Ah @asraa good point, another way to make the disk-backed key store work with delegated roles would be to just write it to disk under each of `bins_00...bins_ff` for example -- `n` copies of the key for `n` bins. That actually might be the easiest.

mnm678:
> Just started doing the review, just got caught up a little in the key generation not being compatible with delegations -- right now i think it is, except for the hashed bin delegations? (i think it should end up saving the keys in a role file, possibly duplicating them)

Yes, I was also running into this [in tuf-notary](https://github.com/notaryproject/tuf/pull/25). You can use the `keys.GenerateEd25519Key()`, etc, but not the helper functions. That would be a good addition here.

trishankatdatadog:
Going to defer to the rest here. Great job, everyone!

local_store.go
mnm678: On another read, I think this shouldn't be an issue with the current design. It's possible that loadPrivateKeys would have duplicates if someone purposefully added them multiple times in different files, but I don't think we need to worry about that here (although we would for verification)

mnm678:
Thanks for the test script @ethan-lowman-dd. I think I fixed the `AddTarget` bug.

repo.go
```diff
@@ -501,6 +497,111 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
+// with a default expiration time.
+func (r *Repo) AddTargetsDelegation(delegator string, role data.DelegatedRole, keys []*data.PublicKey) error {
+	return r.AddTargetsDelegationWithExpires(delegator, role, keys, data.DefaultExpires("targets"))
+}
+
+// AddTargetsDelegationWithExpires adds a delegation from the delegator to the
+// role specified in the role argument. Key IDs referenced in role.KeyIDs
+// should have corresponding Key entries in the keys argument. New metadata is
+// written with the given expiration time.
+func (r *Repo) AddTargetsDelegationWithExpires(delegator string, role data.DelegatedRole, keys []*data.PublicKey, expires time.Time) error {
```
mnm678, line 510: Would there ever be a reason to use a version of this function that specifies which delegation to add the target to? Maybe if a higher level delegation decides to sign something instead of having the delegatee do so?

If that's something that might be useful, I can add a function to do so.

mnm678: I went ahead and added this

mnm678:
Could I get another review @ethan-lowman-dd or @asraa? The hashed-bin key storage seems to be working as long as the keys are each stored (`bins_0-1`, etc). There might be a more elegant way to automate this, but I think we can leave that to a future pr.

ethan-lowman-dd:
I pushed some failing tests -- currently, targets metadata that does not have any target files is not signed. I'm currently debugging.

Also, a note for myself -- all of my printf debugging in `repo.go` needs to be removed still.

ethan-lowman-dd:
I fixed the issues I mentioned above. I think it may be okay if the `SetThreshold` modifier does not support delegation for now -- it can probably be a separate PR. I'll try to find some time today or tomorrow to work on more cleanup and tests.

trishankatdatadog:
@lukpueh do you think you can help review?

ethan-lowman-dd:
Any thoughts on whether delegated targets key revocation needs to go in this PR, or whether we can we leave those as TODOs? I wouldn't consider the code useable in production without APIs to revoke keys, but I also don't want to bloat this PR with mostly-independent changes.

mnm678:
> Any thoughts on whether delegated targets key revocation needs to go in this PR, or whether we can we leave those changes as TODOs? I wouldn't consider the code useable in production without APIs to revoke keys, but I also don't want to bloat this PR with mostly-independent changes.

I think we can update all of the add/revoke key functions in a separate pr to keep this one a bit smaller/easier to review. But I agree that we should get those done before a release that includes this pr.

local_store.go
```diff
@@ -140,19 +216,32 @@ func (f *fileSystemStore) stagedDir() string {
...
+		if filepath.Ext(path) == ".json" {
+			metadataFiles = append(metadataFiles, path)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	//also walk repoDir
+	err = filepath.Walk(f.repoDir(), func(path string, info os.FileInfo, err error) error {
```
ethan-lowman-dd, line 233: @mnm678 This implementation gives precedence to unstaged metadata in the case of staged & unstaged metadata having the same name. Shouldn't staged metadata get priority, as in the implementation of `GetMeta()` on `memoryStore`?

repo.go
```diff
@@ -319,10 +295,6 @@ func (r *Repo) timestamp() (*data.Timestamp, error) {
 }
 
 func (r *Repo) ChangePassphrase(keyRole string) error {
-	if !verify.ValidRole(keyRole) {
```
rdimitrov, line 322: Maybe `!roles.IsTopLevelRole(keyRole)`? If I recall correctly, so far it's possible to generate a key (ergo have a keys file that can be passphrase encrypted) only for top-level roles.

repo.go
```diff
@@ -348,17 +324,13 @@ func (r *Repo) GenKeyWithExpires(keyRole string, expires time.Time) (keyids []st
...
 }
 
 func (r *Repo) AddPrivateKeyWithExpires(keyRole string, signer keys.Signer, expires time.Time) error {
-	if !verify.ValidRole(keyRole) {
-		return ErrInvalidRole{keyRole}
-	}
-
-	if !validExpires(expires) {
-		return ErrInvalidExpires{expires}
-	}
```
rdimitrov, line 361: I see this is moved to `AddVerificationKeyWithExpiration` and so is it okay to verify the role/expiration **after** we do `SaveSigner()`?

rdimitrov:
Just a general comment, but it would be really great for go-tuf if we can try to keep the PRs simpler and independent as much as possible so they are easy to review. For example, If there's some refactoring related to introducing a new feature, try to separate it in a different PR. Trying to make sense of a lot of changes put in a single PR is usually only feasible for people with very in-depth knowledge of the whole codebase. TUF, itself, is not an easy topic, so avoiding this would help in the direction of welcoming more people to review (and even contribute to) the project which would also limit the chance of introducing code that was not thought through well enough 👍

ethan-lowman-dd:
@rdimitrov Since supporting delegated roles touches almost everywhere roles are used, it's been tricky to keep this change compact. However, I think at this point we're confident enough in the necessary refactors that it makes sense to separate them out as you suggested. I'll spend some time on this today.

ethan-lowman-dd:
@mnm678 Mind if I force push a rebase of this branch that includes only the changes not split into the prereq PRs? The new size is `6 files changed, 745 insertions(+), 137 deletions(-)` which is hopefully a lot easier to review.

data/types.go
ethan-lowman-dd: This is fixed here: https://github.com/theupdateframework/go-tuf/pull/195/files

repo.go
ethan-lowman-dd: I don't think there is a security impact here, but in https://github.com/theupdateframework/go-tuf/pull/175/commits/00996cacd27ad5591d35e54ce50d93a36708adaa I reordered these calls so the verification key is added (with the given expiration date) before the signer is added.

local_store.go
ethan-lowman-dd: Is this comment still relevant? I don't think it applies to the current state of the code, since this line is already outside the for loop.

repo.go
ethan-lowman-dd: Fixed.

repo.go
ethan-lowman-dd: Actually this commit made lots of tests fail. Working on figuring out why.

ethan-lowman-dd: Addressed in https://github.com/theupdateframework/go-tuf/pull/175/commits/a983813a0fb5f631dee3352704b19898e2d00d7a: "Revert 00996cacd, add back arg validation to AddPrivateKeyWithExpires, and add comment about func call order"

repo.go
```diff
@@ -653,18 +818,132 @@ func (r *Repo) SignedMeta(roleFilename string) (*data.Signed, error) {
...
+// sign the given path. The final delegation that led to the returned target
+// metadata is also returned.
+//
+// Since there may be multiple targets roles that are able to sign a specific
+// path, we must choose which roles's metadata to return. If preferredRole is
+// specified (non-empty string) and eligible to sign the given path by way of
+// some delegation chain, targets metadata for that role is returned. If
+// preferredRole is not specified (""), we return targets metadata for the
+// final role visited in the depth-first delegation traversal.
+func (r *Repo) targetDelegationForPath(path string, preferredRole string) (*data.Targets, *targets.Delegation, error) {
```
asraa, line 870: Thanks for supported prioritized delegations!

ethan-lowman-dd:
@mnm678 When I merged the diffbase, it looks like this PR was closed. Since you own the PR, I'm not able to set the diffbase. Could you set it to `master` please and reopen the PR?

ethan-lowman-dd:
@mnm678 Ah, I figured it out. I just had to temporarily restore the diffbase branch to reopen, then edit the diffbase.

ethan-lowman-dd:
Something weird happened with the rebase and some changes from prereq PRs were reverted. Going to fix those and then squash the commits on this branch.

asraa:
Hey! Now that the pre-reqs are merged, can we update this one and I'll start a thorough review?

ethan-lowman-dd:
@asraa Sorry for the delay. Are there updates you need other than https://github.com/theupdateframework/go-tuf/issues/239? My plan was to finish the TODOs marked in comments here and add tests where missing.

asraa:
> @asraa Sorry for the delay. Are there updates you need other than https://github.com/theupdateframework/go-tuf/issues/239? My plan was to finish the TODOs marked in comments here and add tests where missing.

I think we can tackle that separately -- imo if people agree that we need easier ways to update fields (versions/expiration) for any data.Signed than this could go in a more general PR.

repo.go
```diff
@@ -476,6 +514,122 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
 	return err
 }
 
+// AddTargetsDelegation is equivalent to AddTargetsDelegationWithExpires, but
+// with a default expiration time.
+func (r *Repo) AddTargetsDelegation(delegator string, role data.DelegatedRole, keys []*data.PublicKey) error {
+	return r.AddTargetsDelegationWithExpires(delegator, role, keys, data.DefaultExpires("targets"))
```
ethan-lowman-dd, line 519: I think it might be nice to have a more compact set of APIs for manipulating targets, using the options pattern to customize expiration, preferred delegated target role, etc. This is out of scope for this PR, though.

repo.go
```diff
@@ -694,28 +975,88 @@ func (r *Repo) AddTargetsWithDigest(digest string, digestAlg string, length int6
...
+			return err
+		}
+
+		// The delegations determine the keys that need to be used to sign each
+		// target file. If a particular target role is reachable in the delegation
+		// graph through multiple paths, it may be the case that the targets
+		// manifest needs signatures from multiple sets of keys.  Therefore, we
+		// merge the keys specified by the current delegation with keys specified
+		// bypreviously seen delegations.
+		if prevMetaWithSigners, ok := targetsMetaToWrite[targetsRoleName]; ok {
```
ethan-lowman-dd, line 1035: I'm not sure this implementation is correct. I think in order to produce metadata that's verifiable by any valid incoming delegation key set, the metadata needs to be signed by the union of the set of all keys in all the delegations that lead to this role. Currently, I think this is only signing using keys for the first traversed delegation that leads to the `preferredRole` (or if `preferredRole` is not specified, the first delegation that can sign for the given path).

ethan-lowman-dd: This is not applicable if the delegation graph is a tree, but the more general case where a role can have multiple valid delegation key sets adds significant complexity like this.

repo.go
```diff
@@ -688,71 +983,177 @@ func (r *Repo) AddTargetsWithDigest(digest string, digestAlg string, length int6
...
+		}
+
 		return nil
 	}); err != nil {
 		return err
 	}
-	return r.writeTargetWithExpires(t, expires)
-}
 
-func (r *Repo) writeTargetWithExpires(t *data.Targets, expires time.Time) error {
```
ethan-lowman-dd, line 742: While making `AddTargetsWithDigest` support delegated targets, I removed `writeTargetWithExpires` since it was only used in two places (`AddTargetsWithDigest` and `Init`). Versions numbers are explicitly bumped/set everywhere else in the code, and having the side effect of mutating the passed `data.Targets` is slightly confusing with that convention.

ethan-lowman-dd:
@asraa Sorry for all the delays, it's been hard to find time for this. This should be ready for review now.

repo.go
```diff
@@ -477,13 +524,150 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
+// ResetTargetsDelegationsWithExpires removes all targets delegations from the
+// given delegator role. New metadata is written with the given expiration
+// time.
+func (r *Repo) ResetTargetsDelegationsWithExpires(delegator string, expires time.Time) error {
+	t, err := r.targets(delegator)
+	if err != nil {
+		return fmt.Errorf("error getting delegator (%q) metadata: %w", delegator, err)
+	}
+
+	t.Delegations = &data.Delegations{}
```
asraa, line 653: It seems like this removes all delegations of a target. If top level targets delegates to Foo and Bar, we can't remove just the Foo delegation, right?

Another comment about this: will this delete the Foo.json and Bar.json from the store?

repo.go
```diff
@@ -477,13 +524,150 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
+	err = r.setMeta("root.json", root)
 	if err == nil {
 		fmt.Println("Revoked", keyRole, "key with ID", id, "in root metadata")
 	}
 	return err
 }
 
+// AddTargetsDelegation is equivalent to AddTargetsDelegationWithExpires, but
+// with a default expiration time.
+func (r *Repo) AddTargetsDelegation(delegator string, delegatedRole data.DelegatedRole, keys []*data.PublicKey) error {
```
hosseinsia, line 536: nite: Make the method more readable. 
```suggestion
func (r *Repo) AddDelegatedRole(delegator string, delegatee data.DelegatedRole, keys []*data.PublicKey) error {
```

repo.go
```diff
@@ -477,13 +524,150 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
+	return r.AddTargetsDelegationWithExpires(delegator, delegatedRole, keys, data.DefaultExpires("targets"))
+}
+
+// AddTargetsDelegationWithExpires adds a delegation from the delegator to the
+// role specified in the role argument. Key IDs referenced in role.KeyIDs
+// should have corresponding Key entries in the keys argument. New metadata is
+// written with the given expiration time.
+func (r *Repo) AddTargetsDelegationWithExpires(delegator string, delegatedRole data.DelegatedRole, keys []*data.PublicKey, expires time.Time) error {
```
hosseinsia, line 544: nit: 
```suggestion
func (r *Repo) AddDelegatedRoleWithExpires(delegator string, delegatee data.DelegatedRole, keys []*data.PublicKey, expires time.Time) error {
```

repo.go
```diff
@@ -299,10 +309,16 @@ func (r *Repo) ChangePassphrase(keyRole string) error {
 }
 
 func (r *Repo) GenKey(role string) ([]string, error) {
```
hosseinsia, line 311: allowing non-tree makes this very complex. See this discussion: https://github.com/theupdateframework/specification/issues/214#issuecomment-1086228742

repo.go
```diff
@@ -530,7 +738,11 @@ func (r *Repo) Sign(roleFilename string) error {
 		return err
 	}
 
-	keys, err := r.getSortedSigningKeys(role)
+	db, err := r.topLevelKeysDB()
```
asraa, line 741: does this mean that `Sign` will only work for top-level roles? I would expect any delegation is allowed to be signed.

local_store.go
```diff
@@ -222,8 +223,7 @@ func (f *fileSystemStore) stagedDir() string {
 }
 
 func isMetaFile(e os.DirEntry) (bool, error) {
```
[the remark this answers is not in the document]
ethan-lowman-dd: I just verified that this is not an issue. Target files live within the `targets` directory, and `isMetaFile` is only called on files in the top-level listing of `repository` and `staged`, not a recursive listing. The directory `repository/targets` or `staged/targets` may be passed to `isMetaFile`, but malicious JSON files within such as `targets/root.json` will not be passed.

repo.go
ethan-lowman-dd: This API will remove all delegations of a target. Is that problematic? My thinking was that this was a minimal starting point to support revocation, and we can add APIs for updates and singular deletes in a separate PR.

This will not delete any of `foo.json, bar.json, x.foo.json, x.bar.json` from the store.

I added a missing test for this method.

repo.go
ethan-lowman-dd: This is easy to support and was an oversight. I added this in https://github.com/theupdateframework/go-tuf/pull/175/commits/189af3ec7bb2d58c4c0e5f3a8049a5a94ae72a66.

repo.go
```diff
@@ -688,71 +1018,127 @@ func (r *Repo) AddTargetsWithDigest(digest string, digestAlg string, length int6
...
-	return r.writeTargetWithExpires(t, expires)
-}
 
-func (r *Repo) writeTargetWithExpires(t *data.Targets, expires time.Time) error {
-	t.Expires = expires.Round(time.Second)
-	if !r.local.FileIsStaged("targets.json") {
-		t.Version++
+	if len(updatedTargetsMeta) == 0 {
+		// This is potentially unexpected behavior kept for backwards compatibility.
+		// See https://github.com/theupdateframework/go-tuf/issues/243
```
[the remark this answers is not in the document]
ethan-lowman-dd: Having a different interpretation of the empty targets slice depending on the role name feels more confusing to me than the original unexpected behavior. It's possible for clients to work around the unexpected behavior by not calling `AddTargets*` if the list of targets is empty, but having two different behaviors may lead to more subtle bugs and more complex functionality for us to support long term.

local_store.go
```diff
@@ -43,6 +42,8 @@ type LocalStore interface {
 	Commit(bool, map[string]int, map[string]data.Hashes) error
 
 	// GetSigners return a list of signers for a role.
+	// This may include revoked keys, so the signers should not
+	// be used without filtering.
```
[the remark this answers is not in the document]
ethan-lowman-dd: > The consequence of failing to filter would be having signatures that don't verify, correct

Correct. 

> It's not a security issue, just an availability one.

I don't think it's a security issue -- more of a usability issue. Signing with untrusted or revoked keys does not present any security risk in TUF, since keys must be explicitly trusted in the client workflow before the signatures count toward a threshold.

> Is the information in LocalStore sufficient to do this filtering

You need a `verify.DB` or something equivalent to do the filtering, which is outside the scope of what `GetSigners` does.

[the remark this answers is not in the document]
ethan-lowman-dd: Also note that the root role needs to sign with expired keys sometimes, so there's no universal way to verify whether a key is valid without a lot of context.

local_store.go
```diff
@@ -43,6 +42,8 @@ type LocalStore interface {
 	Commit(bool, map[string]int, map[string]data.Hashes) error
 
 	// GetSigners return a list of signers for a role.
+	// This may include revoked keys, so the signers should not
```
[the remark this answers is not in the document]
ethan-lowman-dd: I don't think it's possible, or necessarily desired. See this comment thread on the same topic: https://github.com/theupdateframework/go-tuf/pull/175/files#r839629591

repo.go
```diff
@@ -152,9 +162,9 @@ func (r *Repo) RootVersion() (int, error) {
 }
 
 func (r *Repo) GetThreshold(keyRole string) (int, error) {
-	if !roles.IsTopLevelRole(keyRole) {
-		// Delegations are not currently supported, so return an error if this is not a
-		// top-level metadata file.
+	if roles.IsDelegatedTargetsRole(keyRole) {
+		// The signature threshold for a delegated targets role
+		// depends on the incoming delegation edge.
```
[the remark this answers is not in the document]
ethan-lowman-dd: Tracking this issue here: https://github.com/theupdateframework/go-tuf/issues/266

repo.go
```diff
@@ -299,10 +309,16 @@ func (r *Repo) ChangePassphrase(keyRole string) error {
 }
 
 func (r *Repo) GenKey(role string) ([]string, error) {
+	// Not compatible with delegated targets roles, since delegated targets keys
```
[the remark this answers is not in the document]
ethan-lowman-dd: Tracking this here: https://github.com/theupdateframework/go-tuf/issues/266

The input validation on the role name happens in `AddPrivateKeyWithExpires`.

repo.go
```diff
@@ -606,46 +818,45 @@ func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signatur
...
-// getSortedSigningKeys returns available signing keys, sorted by key ID.
+// getSignersInDB returns available signing interfaces, sorted by key ID.
 //
 // Only keys contained in the keys db are returned (i.e. local keys which have
 // been revoked are omitted), except for the root role in which case all local
 // keys are returned (revoked root keys still need to sign new root metadata so
 // clients can verify the new root.json and update their keys db accordingly).
-func (r *Repo) getSortedSigningKeys(name string) ([]keys.Signer, error) {
-	signingKeys, err := r.local.GetSigners(name)
+func (r *Repo) getSignersInDB(roleName string, db *verify.DB) ([]keys.Signer, error) {
```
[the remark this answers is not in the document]
ethan-lowman-dd: I think the name `getSignersInDB` is more clear than `getSigners`, since it distinguishes the functionality from `LocalStore.GetSigners`, which returns an unfiltered list of signers in the local store.  `getSignersInDB` filters the returned list of signers to those present in the given `verify.DB`.

repo.go
ethan-lowman-dd: I changed the method name in https://github.com/theupdateframework/go-tuf/pull/175/commits/0f60bf5964080d24bd5ad7c9847b87aaca6f4d9f, but left the `delegatedRole` name, since to me the parallel names `delegator` and `delegatee` look like they should be the same type, but they are in fact different types.

repo.go
ethan-lowman-dd: See comment above and https://github.com/theupdateframework/go-tuf/pull/175/commits/0f60bf5964080d24bd5ad7c9847b87aaca6f4d9f

repo.go
ethan-lowman-dd: I tried to write a unit test for `Sign` and `AddOrUpdateSignature` using only public methods, but couldn't find a way to create a new unsigned metadata file for a delegated role. I think we would need to make `Repo.setMeta` public to support that, and I'm not sure we want to do that in this same PR. Did you have code that needed to use `Sign` or `AddOrUpdateSignature`? If so, how were you creating the unsigned meta files?

asraa: Yeah! I was able to create unsigned meta files. 

Shouldn't this function do it? https://github.com/theupdateframework/go-tuf/blob/0f60bf5964080d24bd5ad7c9847b87aaca6f4d9f/repo.go#L579

So Init repo, gen top-level keys, add targets and delegated role -- the delegated role addition doesn't include adding the keys to the store, so you'll end up with it unsigned. See example [here](https://github.com/asraa/test-sigstore-root/blob/765a319bd3778a6661e69fc9b816e535b2bf1a0c/cmd/tuf/app/add-delegation.go#L89-L112) (Note: i am using a slightly older API)

ethan-lowman-dd: In unit tests, previously what I've been doing is calling `localStore.SaveSigner` with the delegated role key before I call `repo.AddDelegatedRole`.

I added unit tests in https://github.com/theupdateframework/go-tuf/pull/175/commits/2d5f10dfd0ea3e5a6b766cc13739a93ffef6e1c8 that test `Sign` and `AddOrUpdateSignature` with the usage pattern you described.

client/errors.go
```diff
@@ -58,6 +58,7 @@ func isDecodeFailedWithErrRoleThreshold(err error) bool {
 	return isErrRoleThreshold(e.Err)
 }
 
+//lint:ignore U1000 unused
```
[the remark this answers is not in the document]
ethan-lowman-dd: Yes, but I went ahead and just deleted the unused functions in https://github.com/theupdateframework/go-tuf/pull/175/commits/cbb2528225e3f789caa15d55481deba1b664b796

repo.go
```diff
@@ -477,13 +524,150 @@ func (r *Repo) RevokeKeyWithExpires(keyRole, id string, expires time.Time) error
...
+
+	delegatee := delegatedRole.Name
+	dt, err := r.targets(delegatee)
+	if err != nil {
+		return fmt.Errorf("error getting delegatee (%q) metadata: %w", delegatee, err)
+	}
+	dt.Expires = expires
+
+	delegateeFile := delegatee + ".json"
+	if !r.local.FileIsStaged(delegateeFile) {
```
asraa, line 592: Do we expect this file to possibly already be created? On one hand I suppose this could be used to "update" a delegation, but on the other hand, this function will also clear any targets that were in the delegated file.

Side question: Does the version of the delegateeFile bump if the signers change, but the content (targets and further delegations) do not?

repo.go
```diff
@@ -74,12 +81,15 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	root.ConsistentSnapshot = consistentSnapshot
 	// Set root version to 1 for a new root.
 	root.Version = 1
-	if err = r.setTopLevelMeta("root.json", root); err != nil {
+	if err = r.setMeta("root.json", root); err != nil {
 		return err
 	}
-	if err = r.writeTargetWithExpires(t, data.DefaultExpires("targets")); err != nil {
+
+	t.Version = 1
```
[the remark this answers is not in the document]
ethan-lowman-dd: If `data.NewTargets` or `NewRoot()` set `Version` to 1, that breaks a number of existing tests. Currently, the version is incremented the first time the metadata is staged to the local store. Initializing the version to zero is a bit of a hack that gets the version incremented to 1 when the meta file is staged.

I agree this could be more intuitive, but we'd have to change code unrelated to delegations and it would be a breaking change for `data.NewTargets` and `data.NewRoot`. What do you think about creating an issue for this and handling this in a separate PR?

repo.go
ethan-lowman-dd: > Do we expect this file to possibly already be created?

`delegateeFile` may already be created. For example, if you have a diamond delegation graph with edges (A->B), (A->C), (B->D), (C->D), `D.json` will already exist when you create the delegation from C to D.

> this function will also clear any targets that were in the delegated file

This function should not clear any targets in `delegateeFile`, since the call `r.targets(delegatee)` several lines up returns existing metadata if it exists, or creates empty targets metadata if necessary. I added a test in https://github.com/theupdateframework/go-tuf/pull/175/commits/be25f9324374ffb1c599d3f801150cfe6a6aa344 to verify this.

> Does the version of the delegateeFile bump if the signers change, but the content (targets and further delegations) do not?

The version number should not change unless the content within the `signed` object changes. Adding or removing from the `signatures` section of a `data.Signed` meta file should not change the version number. However, `AddDelegatedRoleWithExpires` does currently refresh the expiration of the delegatee file (defined inside `signed`), which will trigger a version bump (also verified in https://github.com/theupdateframework/go-tuf/pull/175/commits/be25f9324374ffb1c599d3f801150cfe6a6aa344).

repo.go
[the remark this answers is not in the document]
asraa: Sg! https://github.com/theupdateframework/go-tuf/issues/272

asraa (approved):
Let's go! Really want to start getting this code in, and we can tackle the cleanups afterwards. If need be let's make a milestone tracking issue for all the tasks we need to do so we can sort out who can own each.

hosseinsia (approved):
LGTM

repo.go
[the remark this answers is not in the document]
abs007: @ethan-lowman-dd  I'm attempting to solve this issue and I have set `Version` to 1 for `NewRoot()` and `NewTargets()`,.
So I'm wondering if I should remove the code lines that increment the version during staging since `Version` has a value of 1 by default now. Would appreciate your feedback on this.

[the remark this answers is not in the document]
ethan-lowman-dd: Hello @abs007, let's correspond in #272 please. This is a merged PR. Could you please share on that issue the problem you're having, ideally including code to reproduce?

Issue #158, opened 2021-09-28
Multiple key ids for a key.PublicKey


asraa:
Brought up in review of https://github.com/theupdateframework/go-tuf/pull/148#discussion_r717159404

go-tuf holds a list of KEY IDs for a given key, but in practice, this only returns a single string. The python implementation also includes only one key id generated by https://github.com/secure-systems-lab/securesystemslib/blob/9b3ea003ab93ab6909d1327dfb2ab7af0309e7cc/securesystemslib/keys.py#L563

Was there a reason for this? If not, I can handle cleaning this up after #148

@hosseinsia

asraa:
From @trishankatdatadog 

> Yeah, this could be related to the old key_hashing_algorithms kerfuffle (TLDR: we allowed using SHA2-256 and/or SHA2-512 to compute different, collision-resistant keyids for the same key).

znewman01:
@asraa to reroute

rdimitrov:
Closing since the code base changed and so it is no longer valid. 

Thanks for raising this! 👍

PR #181, opened 2021-11-30
wip initial draft for fulcio key tap

asraa:

The test main script produces a root like 
```
{
  "signatures": [
    {
      "cert": "2d2d2d2d2d424547494e2043455254494649434154452d2d2d2d2d0a4d4949436f44434341696167417749424167495541507435784e74316d2b686b324f3146396d362f5772495a6f4e4577436759494b6f5a497a6a304541774d770a4b6a45564d424d474131554543684d4d63326c6e63335276636d55755a4756324d52457744775944565151444577687a6157647a644739795a544165467730790a4d5445784d7a41794d544d794d7a6c61467730794d5445784d7a41794d5455794d7a68614d4141775754415442676371686b6a4f5051494242676771686b6a4f0a50514d4242774e434141516c532b53323263326e30356937614b443178632b474732464244794751766b44573771744b737174414b47683434492f6b48397a640a63394f786737446a444c4367745162775a5269416c43646e2b69664c3357776e6f344942556a43434155347744675944565230504151482f42415144416765410a4d424d47413155644a51514d4d416f47434373474151554642774d444d41774741315564457745422f7751434d414177485159445652304f424259454647746d0a41787a66373469516a77454d4d5756464e675949425944634d42384741315564497751594d426141464d6a46485142426d6951704d6c456b367732755375314b0a427450734d49474e426767724267454642516342415153426744422b4d4877474343734741515546427a4143686e426f644852774f69387663484a70646d46300a5a574e684c574e76626e526c626e51744e6a417a5a6d55335a5463744d4441774d4330794d6a49334c574a6d4e7a55745a6a526d4e5755344d4751794f5455300a4c6e4e3062334a685a3255755a3239765a32786c595842706379356a623230765932457a4e6d45785a546b324d6a5179596a6c6d593249784e445976593245750a59334a304d42344741315564455145422f7751554d424b424547467a636d466851476476623264735a53356a623230774b51594b4b77594242414744767a41420a415151626148523063484d364c79396859324e76645735306379356e6232396e62475575593239744d416f4743437147534d343942414d44413267414d4755430a4d457362334f72726963383369367147664e546a313346334f6756506268397a51746a58733171674f467158586b6845596c4d307065672b6156692f304c73380a59774978414f3456625259344471504666456251517a4d2b7a634462345874554c4a4d4548413067597539326666384565374d4932344b5931427a537a7867410a434664396e673d3d0a2d2d2d2d2d454e442043455254494649434154452d2d2d2d2d0a",
      "keyid": "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f",
      "sig": "3046022100a7394397f720f3400666fbd96a077a6fa83bf4ff90ca687f779223331b25fd020221009e32119365cbd736818647d5482da9a0b4debbf0d767fdda15be83fdd3a7dbc0"
    }
  ],
  "signed": {
    "_type": "root",
    "consistent_snapshot": false,
    "expires": "2021-12-01T21:32:36Z",
    "keys": {
      "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f": {
        "keyid_hash_algorithms": [
          "sha256",
          "sha512"
        ],
        "keytype": "sigstore-oidc",
        "keyval": {
          "identity": "asraa@google.com",
          "issuer": "https://accounts.google.com"
        },
        "scheme": "https://fulcio.sigstore.dev"
      }
    },
    "roles": {
      "root": {
        "keyids": [
          "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f"
        ],
        "threshold": 1
      },
      "snapshot": {
        "keyids": [
          "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f"
        ],
        "threshold": 1
      },
      "targets": {
        "keyids": [
          "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f"
        ],
        "threshold": 1
      },
      "timestamp": {
        "keyids": [
          "e47b8d2a2b634135d827af23abe80312caf36ddfbee96f5da18839858065a68f"
        ],
        "threshold": 1
      }
    },
    "spec_version": "1.0",
    "version": 1
  }
}

```

I haven't implemented verifiation yet! Ahh! tomorrow.


asraa:
Some comments. WDYT?
* I know the spec says this should be PEM encoded, but the JSON marshalling in go really doesn't like newlines in the strings. Any suggestions? Escaping? Is that gross? I had some troubles regarding that. So I HexBytes'd for now.
* Right now when you generate a Fulcio key you can specify the identity and issuer, or, if empty, you will run through an oauth flow to automatically pick it up. I felt that was best for managing adding other Fulcio identities besides your own.
* Every time you sign you have to regenerate the key and get a new cert through the device flow. I could probably store state of the issued cert in case that's preferred, but i think keys should truly be one-time use.

trishankatdatadog:
I think @mnm678 and @joshuagl would definitely be the best reviewers for this PR 🙂

pkg/keys/fulcio.go
```diff
@@ -0,0 +1,280 @@
...
+	fulcioClient "github.com/sigstore/fulcio/pkg/client"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/sigstore/fulcio/pkg/generated/client/operations"
+	"github.com/sigstore/fulcio/pkg/generated/models"
+	"github.com/sigstore/sigstore/pkg/oauthflow"
+	"github.com/theupdateframework/go-tuf/data"
+)
+
+func init() {
```
naveensrinivasan, line 28: Init methods are hard to test https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html

asraa: I think that this is OK. This init function is registering this particular key type in to the registered key map that go-tuf can understand. Unit tests for these individual key types exist independently. 

The alternative in that post is to define them in a main function -- that could be the case, but people also use go-tuf as a library.

pkg/keys/ed25519.go
```diff
@@ -83,8 +83,20 @@ func GenerateEd25519Key() (*ed25519Signer, error) {
...
-func (e *ed25519Signer) SignMessage(message []byte) ([]byte, error) {
-	return e.Sign(rand.Reader, message, crypto.Hash(0))
+func (e *ed25519Signer) SignMessage(message []byte) ([]data.Signature, error) {
+	sig, err := e.Sign(rand.Reader, message, crypto.Hash(0))
+	if err != nil {
+		return nil, err
+	}
+	ids := e.PublicData().IDs()
+	sigs := make([]data.Signature, 0, len(ids))
+	for _, id := range ids {
```
mnm678, line 93: Do all of these ids have the same signature? Or should `sig` be calculated in this loop?

pkg/keys/fulcio.go
```diff
@@ -0,0 +1,280 @@
...
+	Issuer   string `json:"issuer"`
+	key      *data.PublicKey
+}
+
+func (e *fulcioVerifier) Public() string {
+	return fmt.Sprintf("%s:%s", e.Issuer, e.Identity)
+}
+
+func (e *fulcioVerifier) Verify(msg, sig []byte) error {
+	// Check signature verification by extracting public key out of x509 cert
```
mnm678, line 52: As I'm sure you know, this needs to be implemented :)

cmd/tuf/gen_key.go
```diff
@@ -21,6 +21,8 @@ with the addition of the key's ID to the role's list of key IDs.
 Alternatively, passphrases can be set via environment variables in the
 form of TUF_{{ROLE}}_PASSPHRASE
 
+If Fulcio is true, this generates a Fulcio OIDC issued key with the current user's ID.
```
mnm678, line 24: Does this need a code change here? Or is this just from the call to GenKey?

mnm678:
> Some comments. WDYT?
Thanks for working on this @asraa!
> 
>     * I know the spec says the cert should be PEM encoded, but the JSON marshalling in go really doesn't like newlines in the strings. Any suggestions? Escaping? Is that gross? I had some troubles regarding that. So I HexBytes'd for now.

I'll take a closer look to see if we can make the PEM work, but if not (or if it seems like a bad idea), we can go back and update the TAP.
> 
>     * Right now when you generate a Fulcio key you can specify the identity and issuer, or, if empty, you will run through an oauth flow to automatically pick it up. I felt that was best for managing adding other Fulcio identities besides your own.

Seems reasonable
> 
>     * Every time you sign you have to regenerate the key and get a new cert through the device flow. I could probably store state of the issued cert in case that's preferred, but i think keys should truly be one-time use.

The only case I know of where a developer would want to sign multiple targets metadata files at once would be with hashed bin delegations. Those require a fair amount of setup already, so it might be reasonable to just recommend an online key in that case. @trishankatdatadog might have a better sense of what this would look like in practice.
> 
>     * The key data associated with a Fulcio key is truly just the identity and issuer. No real "private" key value.
> 
>     * It is the case that my identity provided through GitHub and through Google are distinct, right? (Distinct issuers)

Yes, this falls under "trusting the delegator to know who they are delegating to". If the delegator was malicious they could do a lot worse than delegating to both your GitHub and your Google identities (like delegating to themself).

pkg/keys/ed25519.go
asraa: Yep, all the same. These are the IDs of the key, e.g. maybe computed by two hashes. We should get rid of this though. https://github.com/theupdateframework/go-tuf/issues/158

pkg/keys/fulcio.go
asraa: Added initial verification! Here's the missing pieces....
* Uploading and then verifying a rekor entry. 
* Verifying the cert chain to the Fulcio Root CA. Is this part pinned by the key scheme pointing to the fulcio "https://fulcio.sigstore.dev" effectively? Fulcio doesn't actually serve it's root certificate (https://github.com/sigstore/fulcio/blob/main/openapi.yaml), in cosign we rely on the TUF root. What do you think? I do think fulcio should serve it, but not sure if that's a risk.

asraa: That kind of brings the question of, how do we "trust" rekor, or the timestamping service that will attest that the sig was created when the cert was valid? should that be pinned as well in root.json?

Should we also avoid the network call to the timestamping service by adding in a signed timestamp (by the trusted service) similar to how cosign does it?

trishankatdatadog:
Could we please mark this PR as draft if it's not yet ready for review? 🙂

asraa:
> Could we please mark this PR as draft if it's not yet ready for review? slightly_smiling_face

Sure! On that note, @mnm678 do you think this should be in a separate branch entirely on go-tuf? I'm not sure it makes sense to pull in the sigstore deps on main.

trishankatdatadog:
> Sure! On that note, @mnm678 do you think this should be in a separate branch entirely on go-tuf? I'm not sure it makes sense to pull in the sigstore deps on main.

Thanks! Does Golang offer conditional compilation? Agree not everyone may want Fulcio deps/support out of the box.

asraa:
> Thanks! Does Golang offer conditional compilation? Agree not everyone may want Fulcio deps/support out of the box.

We could use a tag? like `go build -tags=sigstore ./cmd/tuf` that would only compile and add sigstore keys if the tag is used

mnm678:
> We could use a tag? like `go build -tags=sigstore ./cmd/tuf` that would only compile and add sigstore keys if the tag is used

I like this idea, it should be easier to maintain than a branch/fork, but allows people to avoid the sigstore dependency when they don't need it.

rdimitrov:
Closing this as it's no longer relevant to the code base 👍

PR #183, opened 2021-12-02
Add go vet and staticcheck to CI

ethan-lowman-dd:
We can also make use of GitHub's CodeQL static analysis for free since this is open source, but we can leave that to an other PR.


client/client.go
```diff
@@ -560,56 +559,6 @@ func (c *Client) downloadMetaUnsafe(name string, maxMetaSize int64) ([]byte, err
 	return ioutil.ReadAll(io.LimitReader(r, maxMetaSize))
 }
 
-// getRootAndLocalVersionsUnsafe decodes the versions stored in the local
```
ethan-lowman-dd, line 563: This was unused code, and marked as unsafe, so I'm removing it.

trishankatdatadog (approved):

client/delegations_test.go
```diff
@@ -263,6 +263,9 @@ func initTestDelegationClient(t *testing.T, dirPrefix string) (*Client, func() e
 		TargetsPath:  "targets",
 	}
 	remote, err := HTTPRemoteStore(fmt.Sprintf("http://%s/", addr), opts, nil)
+	if err != nil {
```
asraa, line 266: can this be `assert.Nil(t, err)`?

.github/workflows/build.yml
```diff
@@ -33,3 +33,13 @@ jobs:
       run: |
         GO111MODULE=off go get github.com/mattn/goveralls
         $(go env GOPATH)/bin/goveralls -coverprofile=profile.cov -service=github
+    - name: Vet
+      run: go vet ./...
+    - name: Install staticcheck
+      run: "GO111MODULE=off go get honnef.co/go/tools/cmd/staticcheck"
```
asraa, line 39: should this be pinned to a specific version?

asraa (changes requested):

.github/workflows/build.yml
ethan-lowman-dd: I'm not sure if Go 1.15 supports pinning the version in `go get` -- . Do we actually still need to test against Go 1.15? The `go.mod` specifies 1.16 as the minimum version.

asraa: Yeah I don't see why we should...

ethan-lowman-dd: Oh hm, I must have started this change on a stale branch then rebased, since 1.15 was just removed a few days ago: https://github.com/theupdateframework/go-tuf/pull/179

I'll remove this

asraa (approved):
Thanks! LGTM

trishankatdatadog (approved):

PR #188, opened 2021-12-08
expose `Close()` in `LocalStore` interface

dekkagaijin:
This allows for package consumers to elegantly`Close()` the underlying datastore, avoiding memory leaks.

See also: https://github.com/sigstore/cosign/issues/1160

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -24,6 +24,8 @@ const (
 
 // LocalStore is local storage for downloaded top-level metadata.
 type LocalStore interface {
+	io.Closer
+
 	// GetMeta returns top-level metadata from local storage. The keys are
 	// in the form `ROLE.json`, with ROLE being a valid top-level role.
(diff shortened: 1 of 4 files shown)


asraa (approved):
LGTM, this makes sense.

asraa (approved):

arbll:
Hey folks -- Two comments here:
1. This is a breaking change, do we have any plan to implement some kind of versioning and/or start making sure we don't break the API ? Not a major problem in this case but we are now using this library quite intensively internally and it's been slightly painful
2. I'm not sure I understand this change. `LocalStore` is an interface used to define the methods `go-tuf` wants an external component to define. Adding an unused `Close` method is confusing because anyone reading this will expect `go-tuf` itself to call the `Close` method. If I'm not missing anything you should be manipulating your own struct or own interface extending `LocalStore` in your implementation, not pushing those features onto the lib interface.

PR #192, opened 2021-12-10
[Delegations prereq 2] Add hash bins helpers

ethan-lowman-dd:
Splitting up https://github.com/theupdateframework/go-tuf/pull/175

This PR adds helpers to support hash bin targets delegation. The helpers will be used to generate a list of hash bin manifest names as `leaf_00.json, leaf_01.json, leaf_02.json, ..., leaf_ff.json`, which a targets manifest such as `bins.json` may delegate to using the TUF `path_hash_prefixes` feature ([see spec](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes)).

### DIFF
--- a/internal/targets/hash_bins.go
+++ b/internal/targets/hash_bins.go
@@ -1,0 +1,82 @@
+package targets
+
+import (
+	"strconv"
+	"strings"
+)
+
+// hexEncode formats x as a hex string, left padded with zeros to padWidth.
+func hexEncode(x uint64, padWidth int) string {
+	// Benchmarked to be more than 10x faster than padding with Sprintf.
+	s := strconv.FormatUint(x, 16)
+	if len(s) >= padWidth {
+		return s
+	}
+	return strings.Repeat("0", padWidth-len(s)) + s
+}
+
+// HashBin represents a hex prefix range. First should be less than Last.
+type HashBin struct {
+	First uint64
+	Last  uint64
+}
+
+// Name returns the of the role that signs for the HashBin.
+func (b HashBin) Name(prefix string, padWidth int) string {
+	if b.First == b.Last {
+		return prefix + hexEncode(b.First, padWidth)
+	}
+
+	return prefix + hexEncode(b.First, padWidth) + "-" + hexEncode(b.Last, padWidth)
+}
+
+// Enumerate returns a slice of hash prefixes in the range from First to Last.
+func (b HashBin) Enumerate(padWidth int) []string {
+	n := int(b.Last - b.First + 1)
+	ret := make([]string, int(n))
+
+	x := b.First
+	for i := 0; i < n; i++ {
+		ret[i] = hexEncode(x, padWidth)
+		x++
+	}
+
+	return ret
+}
+
+// HashPrefixLength returns the width of hash prefixes if there are
+// 2^(log2NumBins) hash bins.
+func HashPrefixLength(log2NumBins uint8) int {
+	if log2NumBins == 0 {
+		// Hash prefix of "" is represented equivalently as "0-f".
+		return 1
+	}
+
+	// ceil(log2NumBins / 4.0)
+	return int((log2NumBins-1)/4) + 1
+}
+
+// GenerateHashBins returns a slice of length 2^(log2NumBins) that partitions
+// the space of path hashes into HashBin ranges.
+func GenerateHashBins(log2NumBins uint8) []HashBin {
+	numBins := uint64(1) << log2NumBins
+
+	// numPrefixes = 16^(HashPrefixLength(log2NumBins))
+	numPrefixes := uint64(1) << (4 * HashPrefixLength(log2NumBins))
+
+	p := make([]HashBin, numBins)
+
+	first := uint64(0)
+	interval := numPrefixes / numBins
+	last := first + interval - 1
+	for i := uint64(0); i < numBins; i++ {
+		p[i] = HashBin{
+			First: first,
+			Last:  last,
+		}
+		first += interval
+		last += interval
+	}
+
+	return p
+}
--- a/internal/targets/hash_bins.go
+++ b/internal/targets/hash_bins.go
@@ -22,8 +22,8 @@ func hexEncode(x uint64, padWidth int) string {
 
 const bitsPerHexDigit = 4
 
-// numHexDigits returns is the number of hex digits required to encode
-// the given number of bits.
+// numHexDigits returns the number of hex digits required to encode the given
+// number of bits.
 func numHexDigits(numBits int) int {
 	// ceil(numBits / bitsPerHexDigit)
 	return ((numBits - 1) / bitsPerHexDigit) + 1
--- a/internal/targets/hash_bins.go
+++ b/internal/targets/hash_bins.go
@@ -29,20 +29,19 @@ func numHexDigits(numBits int) int {
 	return ((numBits - 1) / bitsPerHexDigit) + 1
 }
 
+// HashBins represents an ordered list of hash bin target roles, which together
+// partition the space of target path hashes equal-sized buckets based on path
+// has prefix.
 type HashBins struct {
-	rolePrefix string
-	bitLen     int
+	rolePrefix  string
+	bitLen      int
+	hexDigitLen int
 
 	numBins           uint64
 	numPrefixesPerBin uint64
-
-	currIndex uint64
-	curr      *HashBin
 }
 
-// NewHashBins creates an iterator over hash bin targets roles, which together
-// partition the space of target path hashes into 2^bitLen buckets, based on
-// path hash prefix.
+// NewHashBins creates a HashBins partitioning with 2^bitLen buckets.
 func NewHashBins(rolePrefix string, bitLen int) (*HashBins, error) {
 	if bitLen < MinDelegationHashPrefixBitLen || bitLen > MaxDelegationHashPrefixBitLen {
 		return nil, fmt.Errorf("bitLen is out of bounds, should be between %v and %v inclusive", MinDelegationHashPrefixBitLen, MaxDelegationHashPrefixBitLen)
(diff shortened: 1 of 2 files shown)


internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// HashBin represents a hex prefix range. First should be less than Last.
+type HashBin struct {
+	First uint64
+	Last  uint64
+}
+
+// Name returns the of the role that signs for the HashBin.
+func (b HashBin) Name(prefix string, padWidth int) string {
```
mnm678, line 25: Depending on who can call this, you might want to validate that padWidth is positive

mnm678 (approved):
One nit, but LGTM

internal/targets/hash_bins.go
ethan-lowman-dd: I believe the code is safe against negative `padWidth`. If you pass a negative value, then the `len(s) >= padWidth` check in `hexEncode` does not apply any extra padding (same as if `padWidth == 0` or `padWidth < len(s)`).

```go
func hexEncode(x uint64, padWidth int) string {
	// ...
	if len(s) >= padWidth {
		return s
	}
	return strings.Repeat("0", padWidth-len(s)) + s
}
```

hosseinsia:
please provide some description of what this PR does.

ethan-lowman-dd:
@hosseinsia Added a brief description. The unit tests are also a good place to look to see example usage of these functions. You can also find the intended usage in go-tuf APIs in `AddTargetsDelegationsForPathHashBins` in `repo.go` here: https://github.com/theupdateframework/go-tuf/pull/175

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+	return int((log2NumBins-1)/4) + 1
+}
+
+// GenerateHashBins returns a slice of length 2^(log2NumBins) that partitions
+// the space of path hashes into HashBin ranges.
+func GenerateHashBins(log2NumBins uint8) []HashBin {
+	numBins := uint64(1) << log2NumBins
+
+	// numPrefixes = 16^(HashPrefixLength(log2NumBins))
+	numPrefixes := uint64(1) << (4 * HashPrefixLength(log2NumBins))
```
hosseinsia, line 65: why 4*?

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+// HashPrefixLength returns the width of hash prefixes if there are
+// 2^(log2NumBins) hash bins.
+func HashPrefixLength(log2NumBins uint8) int {
+	if log2NumBins == 0 {
+		// Hash prefix of "" is represented equivalently as "0-f".
+		return 1
+	}
+
+	// ceil(log2NumBins / 4.0)
+	return int((log2NumBins-1)/4) + 1
```
hosseinsia, line 56: again the constant 4 is used here. Please define a const and use it.

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+func GenerateHashBins(log2NumBins uint8) []HashBin {
+	numBins := uint64(1) << log2NumBins
+
+	// numPrefixes = 16^(HashPrefixLength(log2NumBins))
+	numPrefixes := uint64(1) << (4 * HashPrefixLength(log2NumBins))
+
+	p := make([]HashBin, numBins)
+
+	first := uint64(0)
+	interval := numPrefixes / numBins
```
hosseinsia, line 70: is the interval = 4?

internal/targets/hash_bins_test.go
```diff
@@ -0,0 +1,95 @@
...
+func BenchmarkHexEncode4(b *testing.B) {
+	for n := 0; n <= b.N; n++ {
+		for x := uint64(0); x <= 0xffff; x += 1 {
+			hexEncode(x, 4)
+		}
+	}
+}
+
+func TestHashBin(t *testing.T) {
+	h := HashBin{
```
hosseinsia, line 26: nit: please use test tables instead.

internal/targets/hash_bins_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	h = HashBin{
+		First: 0x0abc,
+		Last:  0xbcde,
+	}
+	assert.Equal(t, "test_0abc-bcde", h.Name("test_", 4))
+}
+
+func TestHashPrefixLength(t *testing.T) {
+	assert.Equal(t, 1, HashPrefixLength(0))
```
hosseinsia, line 52: nit: please use test tables.

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+		return 1
+	}
+
+	// ceil(log2NumBins / 4.0)
+	return int((log2NumBins-1)/4) + 1
+}
+
+// GenerateHashBins returns a slice of length 2^(log2NumBins) that partitions
+// the space of path hashes into HashBin ranges.
+func GenerateHashBins(log2NumBins uint8) []HashBin {
```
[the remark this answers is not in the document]
ethan-lowman-dd: I think we'll have to duplicate this input validation in the public APIs, but I added a check here just to be safe.

internal/targets/hash_bins_test.go
hosseinsia: please split the tests with different natures (one for `Name` method and other for `Enumerate`)

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"strconv"
+	"strings"
+)
+
+// hexEncode formats x as a hex string, left padded with zeros to padWidth.
+func hexEncode(x uint64, padWidth int) string {
+	// Benchmarked to be more than 10x faster than padding with Sprintf.
+	s := strconv.FormatUint(x, 16)
+	if len(s) >= padWidth {
```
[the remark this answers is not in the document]
hosseinsia: We must return an error in such cases (defensive programming), or provide a comment on top of the method explaining the condition(s).

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,86 @@
...
+		return s
+	}
+	return strings.Repeat("0", padWidth-len(s)) + s
+}
+
+// HashBin represents a hex prefix range. First should be less than Last.
+type HashBin struct {
```
hosseinsia, line 23: since padWidth is used almost in all of the method params, if it does not change for a bin, can you set it as`HashBin` struct member?

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,86 @@
...
+func HashPrefixLen(bitLen uint8) int {
+	// ceil(bitLen / bitsPerHexDigit)
+	return int((bitLen-1)/bitsPerHexDigit) + 1
+}
+
+// GenerateHashBins returns a slice of length 2^(bitLen) that partitions the
+// space of path hashes into HashBin ranges.
+func GenerateHashBins(prefixBitLen uint8) ([]HashBin, error) {
+	if prefixBitLen < MinDelegationHashPrefixBitLen || prefixBitLen > MaxDelegationHashPrefixBitLen {
+		return nil, fmt.Errorf("bitLen is out of bounds, should be between %v and %v inclusive", MinDelegationHashPrefixBitLen, MaxDelegationHashPrefixBitLen)
```
hosseinsia, line 64:
```suggestion
		return nil, fmt.Errorf("bitLen is out of range, should be %v <= bitLen <= %v", MinDelegationHashPrefixBitLen, MaxDelegationHashPrefixBitLen)
```

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,86 @@
...
+const bitsPerHexDigit = 4
+
+// HashPrefixLen returns the width of hash prefixes if there are 2^(bitLen)
+// hash bins.
+func HashPrefixLen(bitLen uint8) int {
+	// ceil(bitLen / bitsPerHexDigit)
+	return int((bitLen-1)/bitsPerHexDigit) + 1
+}
+
+// GenerateHashBins returns a slice of length 2^(bitLen) that partitions the
```
hosseinsia, line 60: make the param name in the comment consistent with param

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,86 @@
...
+// space of path hashes into HashBin ranges.
+func GenerateHashBins(prefixBitLen uint8) ([]HashBin, error) {
```
hosseinsia, line 62: q: is prefixBitLen different from bitLen?

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,86 @@
...
+	return ret
+}
+
+const bitsPerHexDigit = 4
+
+// HashPrefixLen returns the width of hash prefixes if there are 2^(bitLen)
+// hash bins.
+func HashPrefixLen(bitLen uint8) int {
+	// ceil(bitLen / bitsPerHexDigit)
+	return int((bitLen-1)/bitsPerHexDigit) + 1
```
hosseinsia, line 57: What about the case bitLen = 0? -- you had it before.

internal/targets/hash_bins_test.go
ethan-lowman-dd: With the refactor, this is now a table test.

ethan-lowman-dd (commented):
I realized when considering how to alleviate the awkwardness of passing around `padWidth` that we were already wasting a ton of memory by building a complete `HashBin` slice (potentially tens of MB for many bins). I refactored a bit to reduce memory usage while also making the API cleaner.

internal/targets/hash_bins.go
mnm678: Is there a use case for using a single bin rather than just a wildcard delegation? Delegations are generally simpler if they can work.

ethan-lowman-dd: I can't think of any use case for a single bin.

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+type HashBins struct {
+	rolePrefix string
+	bitLen     int
+
+	numBins           uint64
+	numPrefixesPerBin uint64
+
+	currIndex uint64
```
hosseinsia, line 39: Here is my suggestion:
instead of forming this as an iterator style, implement a getBin(index)/bin(index) that returns the bin struct with that index.

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,119 @@
...
+func (b *HashBin) RoleName() string {
+	if b.first == b.last {
+		return b.rolePrefix + hexEncode(b.first, b.hexDigitLen)
+	}
+
+	return b.rolePrefix + hexEncode(b.first, b.hexDigitLen) + "-" + hexEncode(b.last, b.hexDigitLen)
+}
+
+// HashPrefixes returns a slice of all hash prefixes in the bin.
+func (b *HashBin) HashPrefixes() []string {
```
hosseinsia, line 108: I assume we also need a reverse lookup, from a prefix to a bin? -- so then you can get the rolename?

hosseinsia (dismissed):
LGTM!
Excellent PR! Thanks so much!

internal/targets/hash_bins.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// NumBins returns the number of hash bin partitions.
+func (hb *HashBins) NumBins() uint64 {
+	return hb.numBins
+}
+
+// GetBin returns the HashBin at index i, or nil if i is out of bounds.
+func (hb *HashBins) GetBin(i uint64) *HashBin {
+	if i >= hb.numBins {
```
hosseinsia, line 72: nit: please also check for 0<=

internal/targets/hash_bins.go
ethan-lowman-dd: Done. I was curious if this would have a big impact on memory allocs for a large number of hash bins, but a quick profiling shows that if you call `RoleName()` in the `HashBin` loop (as you almost always would), the `strconv.FormatUint` in `hexEncode` dominates allocs regardless of whether we use an iterator or index access. Tuning this is probably not so important right now (especially with TAP 15) -- was just curious.

internal/targets/hash_bins.go
ethan-lowman-dd: We will need a reverse lookup to implement TAP 15, but not for the purposes of implementing the TUF 1.0 spec.

internal/targets/hash_bins.go
ethan-lowman-dd: This is an `uint64`, so it can not be negative, and zero is a valid index.

hosseinsia:
makes sense. Thanks!

hosseinsia (approved):

mnm678 (approved):

PR #196, opened 2021-12-10
[Delegations prereq 6] Use a verify.DB for delegation in client

ethan-lowman-dd:
Splitting up https://github.com/theupdateframework/go-tuf/pull/175

This PR removes the type `verify.DelegationsVerifier` in favor of using a `verify.DB` directly to store sets of keys. This simplifies the types, since a `DelegationsVerifier` just wrapped a `DB` without adding significant functionality, and allows us to use the same type (a `verify.DB`) to hold keys for the top level targets role as for delegated targets roles.

### DIFF
--- a/client/delegations.go
+++ b/client/delegations.go
@@ -20,7 +20,7 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
 	// - filter delegations with paths or path_hash_prefixes matching searched target
 	// - 5.6.7.1 cycles protection
 	// - 5.6.7.2 terminations
-	delegations := targets.NewDelegationsIterator(target)
+	delegations := targets.NewDelegationsIterator(target, c.db)
 	for i := 0; i < c.MaxDelegations; i++ {
 		d, ok := delegations.Next()
 		if !ok {
@@ -28,7 +28,7 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
 		}
 
 		// covers 5.6.{1,2,3,4,5,6}
-		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.Verifier)
+		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.DB)
 		if err != nil {
 			return data.TargetFileMeta{}, err
 		}
@@ -39,11 +39,11 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
 		}
 
 		if targets.Delegations != nil {
-			delegationsVerifier, err := verify.NewDelegationsVerifier(targets.Delegations)
+			delegationsDB, err := verify.NewDBFromDelegations(targets.Delegations)
 			if err != nil {
 				return data.TargetFileMeta{}, err
 			}
-			err = delegations.Add(targets.Delegations.Roles, d.Delegatee.Name, delegationsVerifier)
+			err = delegations.Add(targets.Delegations.Roles, d.Delegatee.Name, delegationsDB)
 			if err != nil {
 				return data.TargetFileMeta{}, err
 			}
@@ -75,7 +75,7 @@ func (c *Client) loadLocalSnapshot() (*data.Snapshot, error) {
 }
 
 // loadDelegatedTargets downloads, decodes, verifies and stores targets
-func (c *Client) loadDelegatedTargets(snapshot *data.Snapshot, role string, verifier verify.DelegationsVerifier) (*data.Targets, error) {
+func (c *Client) loadDelegatedTargets(snapshot *data.Snapshot, role string, db *verify.DB) (*data.Targets, error) {
 	var err error
 	fileName := role + ".json"
 	fileMeta, ok := snapshot.Meta[fileName]
@@ -98,11 +98,7 @@ func (c *Client) loadDelegatedTargets(snapshot *data.Snapshot, role string, veri
 	// 5.6.3 verify signature with parent public keys
 	// 5.6.5 verify that the targets is not expired
 	// role "targets" is a top role verified by root keys loaded in the client db
-	if role == "targets" {
-		err = c.db.Unmarshal(raw, targets, role, fileMeta.Version)
-	} else {
-		err = verifier.Unmarshal(raw, targets, role, fileMeta.Version)
-	}
+	err = db.Unmarshal(raw, targets, role, fileMeta.Version)
 	if err != nil {
 		return nil, ErrDecodeFailed{fileName, err}
 	}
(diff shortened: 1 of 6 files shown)


mnm678 (approved):

internal/targets/delegation.go
```diff
@@ -18,13 +19,23 @@ type delegationsIterator struct {
 }
 
 // NewDelegationsIterator initialises an iterator with a first step
-// on top level targets
-func NewDelegationsIterator(target string) *delegationsIterator {
+// on top level targets.
+func NewDelegationsIterator(target string, topLevelKeysDB *verify.DB) *delegationsIterator {
+	role := topLevelKeysDB.GetRole("targets")
+	keyIDs := []string{}
+	if role != nil {
```
asraa, line 26: Would it be an error or do we never expect that the keys DB never have targets?

hosseinsia:
Would you please add some description to this PR?

client/delegations.go
```diff
@@ -20,15 +20,15 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
...
+	delegations := targets.NewDelegationsIterator(target, c.db)
 	for i := 0; i < c.MaxDelegations; i++ {
 		d, ok := delegations.Next()
 		if !ok {
 			return data.TargetFileMeta{}, ErrUnknownTarget{target, snapshot.Version}
 		}
 
 		// covers 5.6.{1,2,3,4,5,6}
-		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.Verifier)
+		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.DB)
```
hosseinsia, line 31: sorry I do not understand the need for this change. Would you please explain?

client/delegations.go
```diff
@@ -39,11 +39,11 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
 		}
 
 		if targets.Delegations != nil {
-			delegationsVerifier, err := verify.NewDelegationsVerifier(targets.Delegations)
+			delegationsDB, err := verify.NewDBFromDelegations(targets.Delegations)
```
hosseinsia, line 42: probably the old names are still good regardless of the type returned here?

client/delegations.go
```diff
@@ -75,7 +75,7 @@ func (c *Client) loadLocalSnapshot() (*data.Snapshot, error) {
 }
 
 // loadDelegatedTargets downloads, decodes, verifies and stores targets
-func (c *Client) loadDelegatedTargets(snapshot *data.Snapshot, role string, verifier verify.DelegationsVerifier) (*data.Targets, error) {
+func (c *Client) loadDelegatedTargets(snapshot *data.Snapshot, role string, db *verify.DB) (*data.Targets, error) {
```
hosseinsia, line 78: can we somehow hide the implementation details here?

client/delegations.go
ethan-lowman-dd: Does the description I added to the PR help explain the reason for the change?

client/delegations.go
ethan-lowman-dd: True, but `delegationsDB` is both shorter and more accurate.

client/delegations.go
ethan-lowman-dd: I think this is a simplification in implementation. Before, `verify.DelegationsVerifier` was essentially the same type as `verify.DB`, but used only in certain contexts.

internal/targets/delegation.go
```diff
@@ -2,13 +2,14 @@ package targets
...
 	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/internal/sets"
 	"github.com/theupdateframework/go-tuf/verify"
 )
 
 type Delegation struct {
 	Delegator string
-	Verifier  verify.DelegationsVerifier
 	Delegatee data.DelegatedRole
+	DB        *verify.DB
```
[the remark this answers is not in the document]
ethan-lowman-dd: A `DB` is used to hold a set of public keys and a set of roles. This type is used to hold the public keys specified for a delegation to a role.

internal/targets/delegation.go
ethan-lowman-dd: I made this return an error if the top level DB is missing the targets role, and added tests to exercise this functionality.

internal/targets/delegation.go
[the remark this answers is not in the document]
hosseinsia: Seems a bit raw to pass the DB around.

pkg/targets/delegation.go
```diff
@@ -57,7 +72,7 @@ func (d *delegationsIterator) Next() (value Delegation, ok bool) {
 	return delegation, true
 }
 
-func (d *delegationsIterator) Add(roles []data.DelegatedRole, delegator string, verifier verify.DelegationsVerifier) error {
+func (d *delegationsIterator) Add(roles []data.DelegatedRole, delegator string, db *verify.DB) error {
```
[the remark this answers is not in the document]
ethan-lowman-dd: A delegation (an edge in the traversal) is defined in part by a set of keys (see [spec](https://theupdateframework.github.io/specification/latest/#delegations)), so the keys DB is specific to particular edge. It would not be correct to use the same DB for the entire traversal.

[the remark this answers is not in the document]
hosseinsia: I see. I can not make specific suggestion but I wish there was a way this code was more intuitive. Other than that, this looks great!

client/delegations.go
hosseinsia: That should have been even more confusing. But now that this is actually a mapping to keys, using DB does not make sense anymore an confusing, specifically that there is a db per delegation path (is my understanding correct?)

hosseinsia (dismissed):
LGTM. -- I do not want to be a blocker for this PR if someone needs it. However, I believe this should be made more readable.

client/delegations.go
```diff
@@ -20,15 +20,19 @@ func (c *Client) getTargetFileMeta(target string) (data.TargetFileMeta, error) {
...
+
 	for i := 0; i < c.MaxDelegations; i++ {
 		d, ok := delegations.Next()
 		if !ok {
 			return data.TargetFileMeta{}, ErrUnknownTarget{target, snapshot.Version}
 		}
 
 		// covers 5.6.{1,2,3,4,5,6}
-		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.Verifier)
+		targets, err := c.loadDelegatedTargets(snapshot, d.Delegatee.Name, d.DB)
```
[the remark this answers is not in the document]
ethan-lowman-dd: Done in https://github.com/theupdateframework/go-tuf/pull/196/commits/bcf8b66a364a75fbd8885b0969b786d3580b2084.

hosseinsia (approved):
LGTM

mnm678 (approved):

ethan-lowman-dd:
@asraa This has two stamps now -- did you want to take another look before I merge?

asraa (approved):
Let's ship it!

PR #154, opened 2021-09-15
Rename manifest to metadata for consistency with the TUF specification

rdimitrov:
The following PR renames the use of `manifest` to `metadata` so the implementation is consistent with what is used within the TUF specification.

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

PR #198, opened 2021-12-10
[Delegations prereq 8] Make snapshot manifest list dynamic, and remove root.json from snapshot

ethan-lowman-dd:
In https://github.com/theupdateframework/go-tuf/pull/175, we will update the implementation of `snapshotMetadata` to dynamically return delegated targets as well as `targets.json`, so we prepare for that we're putting the list of metadata files that snapshot pins into a function.

We are also opportunistically using this PR to remove `root.json` from snapshot.


mnm678 (approved):

repo.go
```diff
@@ -803,6 +798,12 @@ func (r *Repo) Snapshot() error {
 	return r.SnapshotWithExpires(data.DefaultExpires("snapshot"))
 }
 
+func (r *Repo) snapshotManifests() []string {
+	// Note: root pinning is not supported in Spec 1.0.19.
```
asraa, line 802: Should we parameterize/flag protect this? Using go-tuf as a library causes dependent libraries to print out the root pinning warning. It would be nice to let users decide whether to include root
https://github.com/theupdateframework/go-tuf/issues/203

asraa (approved):

repo.go
asraa: Can this by default remove root.json? If we log warn about it, we should default to avoid the log.

But still think perhaps we should add a parameter to include root when calling `Snapshot`

zwass: What would be the motivation for a user wanting the `root.json` included? IMO it doesn't matter much how it's configured as long as the default is to not include it. But if there's not a clear motivation for providing the option I'd say keep it simple and don't add the option.

ethan-lowman-dd: I removed root.json from the snapshot. I had to change a number of tests to make them pass and also regenerate test data.

As of https://github.com/theupdateframework/go-tuf/pull/143, verification of root.json in snapshot has been removed from the client (replaced with that log line). I'm not sure I see the value in keeping a config option around to write root.json to snapshot if it's not actually verified in the client. The only thing that comes to mind is whether old clients will break against new repos written without root.json in snapshots, but changing the default with configuration via a build flag or by adding an options API is also a breaking change on repo side. I would lean toward making the repo-side breaking change for simplicity and to match the 1.0 spec sooner. What do you think @asraa?

repo.go
```diff
@@ -830,6 +825,12 @@ func (r *Repo) Snapshot() error {
...
```
joshuagl, line 829:
```suggestion
```

repo.go
```diff
@@ -830,6 +825,12 @@ func (r *Repo) Snapshot() error {
...
+	// root.json might need to be removed.
```
joshuagl, line 830:
```suggestion
```

repo.go
```diff
@@ -830,6 +825,12 @@ func (r *Repo) Snapshot() error {
 	return r.SnapshotWithExpires(data.DefaultExpires("snapshot"))
 }
 
+func (r *Repo) snapshotManifests() []string {
```
joshuagl, line 828: We moved away from using the term `manifests` to `metadata` in https://github.com/theupdateframework/go-tuf/pull/154 to better align with specification terminology.

joshuagl (commented):
This series looks good, thanks. Minor suggestions on variable name (manifests->metadata) and removing stale comments.

joshuagl (approved):
LGTM, thanks Ethan!

repo.go
asraa: >  I would lean toward making the repo-side breaking change for simplicity and to match the 1.0 spec sooner. 

This is reasonable to me!

asraa (approved):

PR #204, opened 2022-01-10
Improve CLI experience by avoiding version gaps

asraa:
Improves the CLI experience by avoiding jumping versions between metadata updates.

Prior to this, each CLI command would increment the metadata version since repo state wasn't held between repo/store initialization. This moves the logic into the LocalStore so that versions are only incremented when a metadata files is newly staged, and then resets it on Commit to allow incrementing again.

This avoids the inevitable gap that starts root.json at version 4 after gen-key'ing one key in each role.

cc @haydentherapper


repo.go
```diff
@@ -49,7 +49,15 @@ type LocalStore interface {
 	// If paths is empty, all staged target files will be walked.
 	WalkStagedTargets(paths []string, targetsFn TargetsWalkFunc) error
 
+	// IsStaged determines if a metadata file is currently staged, to avoid incrementing
+	// version numbers repeatedly while staged.
+	IsStaged(string) bool
```
ethan-lowman-dd, line 54: Can we name arguments in the interface?

Additionally, one thing I'd like to gradually clean up is naming of role names vs. manifest/file names. Could we name the arguments or the function name to reflect that this takes in a manifest/file name? e.g. `FileIsStaged(name string)` or `IsStaged(fileName string)` or whatever you think makes the most sense. It looks like this function could take in staged targets files as well as staged manifests.

repo.go
```diff
@@ -111,6 +113,8 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	}
 	root := data.NewRoot()
 	root.ConsistentSnapshot = consistentSnapshot
+	// Start a new root at version 1.
+	root.Version++
```
ethan-lowman-dd, line 117: nit: Should we just modify `data.NewRoot` to set `Version: 1`?

asraa: That would be nice... The one bad part about doing it this way is that if you don't call `repo.Init` and just go from `NewRepo` to `AddKey`, then you'll accidentally increment twice, first from getting `NewRoot` and then since you haven't written to local yet, incrementing once more.

repo.go
asraa: done

repo.go
ethan-lowman-dd: Ah I see. Thanks for explaining

ethan-lowman-dd (approved):

mnm678 (approved):

Issue #208, opened 2022-01-24
Unable to initialize client without fetching from remote


haydentherapper:
Before calling `client.Update` or `client.UpdateRoots`, it is expected that the client has been initialized with `client.Init`. [`Init`](https://github.com/theupdateframework/go-tuf/blob/master/client/client.go#L114) uses a set of root keys, pulls the latest `root.json`, verifies it with those keys, and saves it locally.

The issue is there's no guarantee that the latest root can be verified using the set of root keys. For example, let's say a project is shipped with the first `root.json`. Before a new root signing, one key is rotated. After 5 signings, all keys used to sign the remote `root.json` will have been rotated. `Init` will fail, and there's no other way to directly initialize the client.

The proper way to initialize a client would be to initialize the client with a local root, then walk and verify the chain using `Update` or `UpdateRoots`. I'd like to propose that `Init` take the local `root.json` to initialize the database, and not make any calls to the remote to fetch the latest `root.json`.

Another option would be for `Init` to take the set of root keys and threshold, its current API, and call `UpdateRoots` as part of initialization. I would prefer that this not be the default, for those who don't want to initialize a client with the latest root.

cc @asraa

haydentherapper:
Until a versioned release process is in place, I've added a second method `InitLocal` that resolves this issue. After the process is in place, I'll submit another PR to remove `Init`.

asraa:
Scheduling deprecation for v0.4

Issue #203, opened 2022-01-09
Remove root pinning in snapshot


asraa:
TUF specification does not require root pinning in the snapshot file https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#file-formats-snapshotjson--file-formats-snapshot

Remove https://github.com/theupdateframework/go-tuf/blob/5573c9c8694deb40c802a5d9d62eed5c1c47041a/repo.go#L31

c/f https://twitter.com/trishankkarthik/status/1479894513627930633

zwass:
`go-tuf` now generates a repository that always causes warnings (which cannot be silenced due to use of `fmt.Println`) because of this: https://github.com/theupdateframework/go-tuf/blob/master/client/client.go#L196

Can we either quickly move to fix the output so that it's internally consistent, or roll back that log statement?

We keep hearing from folks who are concerned about this, similar to the Twitter thread: https://github.com/fleetdm/fleet/issues/3699

Is it simply a matter of removing `root.json` from that list?

asraa:
> Is it simply a matter of removing root.json from that list?

Yes! It's handled in https://github.com/theupdateframework/go-tuf/pull/198

@ethan-lowman-dd do you mind if we split that part out?

ethan-lowman-dd:
@asraa Should we try to merge all of #198 or do you want to separate out just part of it?

asraa:
I think the PR stands alone pretty easily (I forgot that #8 means it doesn't need to be #8 merged). Taking another review on it.

trishankatdatadog:
> c/f https://twitter.com/trishankkarthik/status/1479894513627930633

Good lords, how many times do you get a tweet cited in an issue... 😅

zwass:
Looks like this can be closed now that #198 is merged? Thanks y'all!

Issue #200, opened 2021-12-15
Set up versioned releases


ethan-lowman-dd:
We are still pre-1.0, but we need to give users a way to know when breaking changes are made and upgrade accordingly.

As discussed in the TUF community meeting, we intend to version pre-1.0 releases as v0.x.y, where x is incremented for breaking changes.

We will build release notes from a changelog file, updated in each PR as necessary. Changes will be tagged by sub-system, like `[client]`, `[cli]`, or similar.

A PR review checklist may help with the process.

shibumi:
Note: We have a PR for releases via goreleaser and goreleaser can handle pre-releases + generate changelogs. It really makes sense to have a closer look on my PR here and maybe get this done for the first release: 

https://github.com/theupdateframework/go-tuf/pull/157

shibumi:
If needed I can provide a test release in my fork, so that you can have a look at it and  we discuss changes etc

joshuagl:
Cross-referencing an older issue where the need for releases is discussed https://github.com/theupdateframework/go-tuf/issues/114

zwass:
Glad to see that this has been addressed!

Issue #114, opened 2020-02-28
Plan to release go-tuf and roadmap


chuanbozhang-okta:
Is there a plan to release this? Also what is the roadmap planned.

titanous:
Hi, we'd like to get it updated to the latest version of the spec, but haven't had time yet. If you want to help, you could backport changes from https://fuchsia.googlesource.com/third_party/go-tuf/

We consider it to be "released", what are you looking for there?

erickt:
We finally merged in the fuchsia patches, and I've uploaded my outstanding patches in #122, #123, #124. I wrote these some time ago, so I wouldn't be surprised if we had to do some iteration on them to get them ready for landing. After they land, we'll need to implement delegation support, and any other missing features from the TUF spec. At that point we might be in a good spot for doing a go-tuf release.

zwass:
@titanous there is no tagged release on GitHub and therefore (to a naive potential user such as myself) no way to understand the maturity/stability of the project.

@erickt It looks like you have merged some PRs recently, but I don't see recent progress on the PRs you linked above. This is not hugely concerning except for #124 which is going to move to a new major version of the TUF specification. Should we we expect this to cause backwards-incompatible changes for repositories and/or clients using the current `HEAD` of `master`?

When looking at implementing TUF for a new project this implementation seems compelling, but the stability and maturity are unclear.

zwass:
Is there any recommendation as to whether to use this repository vs. https://fuchsia.googlesource.com/third_party/go-tuf/?

znewman01:
>Also what is the roadmap planned.

There is now a [GitHub project](https://github.com/orgs/theupdateframework/projects/1) with a roadmap for the next release of go-tuf. 

> there is no tagged release on GitHub 

There are now version tags (`v0.3.0` is the latest at the time of this comment).

> Is there any recommendation as to whether to use this repository vs. https://fuchsia.googlesource.com/third_party/go-tuf/?

In general, Google `third_party` code is just vendored from some upstream. In this case, the Fuchsia project vendors *this* repository, so you should prefer to use [theupdateframework/go-tuf](https://github.com/theupdateframework/go-tuf).

PR #210, opened 2022-01-25
Add client method for initializing from root metadata

haydentherapper:
This will allow users of go-tuf to initialize a client
without requiring a remote connection. This is useful
for when a root has been updated and the local root
verification keys can no longer verify the the latest
remote root, except by verifying the chain with
client.Update().

Ref #208

Signed-off-by: Hayden Blauzvern <hblauzvern@google.com>

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -111,6 +111,8 @@ func NewClient(local LocalStore, remote RemoteStore) *Client {
 // The latest root.json is fetched from remote storage, verified using rootKeys
 // and threshold, and then saved in local storage. It is expected that rootKeys
 // were securely distributed with the software being updated.
+//
+// Deprecated: Use c.InitLocal and c.Update to initialize a local repository.
 func (c *Client) Init(rootKeys []*data.PublicKey, threshold int) error {
 	if len(rootKeys) < threshold {
 		return ErrInsufficientKeys
@@ -144,6 +146,20 @@ func (c *Client) Init(rootKeys []*data.PublicKey, threshold int) error {
 		return err
 	}
 
+	return c.local.SetMeta("root.json", rootJSON)
+}
+
+// InitLocal initializes a local repository from root metadata.
+//
+// The root's keys are extracted from the root and saved in local storage.
+// Root expiration is not checked.
+// It is expected that rootJSON was securely distributed with the software
+// being updated.
+func (c *Client) InitLocal(rootJSON []byte) error {
+	err := c.loadAndVerifyRootMeta(rootJSON, true /*ignoreExpiredCheck*/)
+	if err != nil {
+		return err
+	}
 	return c.local.SetMeta("root.json", rootJSON)
 }
 
@@ -430,6 +446,12 @@ func (c *Client) loadAndVerifyLocalRootMeta(ignoreExpiredCheck bool) error {
 	if !ok {
 		return ErrNoRootKeys
 	}
+	return c.loadAndVerifyRootMeta(rootJSON, ignoreExpiredCheck)
+}
+
+// loadAndVerifyRootMeta decodes and verifies root metadata and loads the top-level keys.
+// This method first clears the DB for top-level keys and then loads the new keys.
+func (c *Client) loadAndVerifyRootMeta(rootJSON []byte, ignoreExpiredCheck bool) error {
 	// unmarshal root.json without verifying as we need the root
 	// keys first
 	s := &data.Signed{}
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -280,6 +280,50 @@ func (s *ClientSuite) TestInit(c *C) {
 	c.Assert(err, IsNil)
 }
 
+func (s *ClientSuite) TestInitLocalAllowsExpired(c *C) {
+	s.genKeyExpired(c, "targets")
+	c.Assert(s.repo.Snapshot(), IsNil)
+	c.Assert(s.repo.Timestamp(), IsNil)
+	c.Assert(s.repo.Commit(), IsNil)
+	s.syncRemote(c)
+	client := NewClient(MemoryLocalStore(), s.remote)
+	bytes, err := io.ReadAll(s.remote.meta["root.json"])
+	c.Assert(err, IsNil)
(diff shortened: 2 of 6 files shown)


haydentherapper:
cc @asraa 

Open question: Should we remove `client.Init` in favor of this method? Given that go-tuf does not have a 1.x.x release, I'd assume it'd be fine to drop support for `client.Init` without having to worry about deprecation.

joshuagl:
> Open question: Should we remove `client.Init` in favor of this method? Given that go-tuf does not have a 1.x.x release, I'd assume it'd be fine to drop support for `client.Init` without having to worry about deprecation.

The absence of a formal. 1.x.x release does technically absolve us of having to worry about deprecation. However, go-tuf has active users who have been bitten by API changes (i.e. https://github.com/theupdateframework/go-tuf/issues/203). We should be wary about making any breaking API and behaviour changes until we have made a release that can be pinned to and formalised our releases. See https://github.com/theupdateframework/go-tuf/issues/200 and https://github.com/theupdateframework/go-tuf/issues/114

client/client.go
```diff
@@ -147,6 +147,20 @@ func (c *Client) Init(rootKeys []*data.PublicKey, threshold int) error {
 	return c.local.SetMeta("root.json", rootJSON)
 }
 
+// InitWithRoot initializes a local repository from root metadata.
```
asraa, line 150: Very small nit: can we call out that this explicitly does not try to fetch or update from a remote?

asraa (commented):
One comment more about `client.Init` generally: I think, as Hayden suggested from writing this method, that it doesn't always make sense that a client could only init from the **latest** root keys like `client.Init` does. 

In fact, I think this new method which effectively just loads trusted root metadata matches the TUF spec better (https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#load-trusted-root-metadata--load-trusted-root).

hosseinsia:
To the point that @joshuagl made about active users, we are using the existing init ... so removing it would break us ...

trishankatdatadog:
Do we have a release process (which includes making breaking MINOR 0.MINOR.PATCH releases for now) formalized? If so, we should formalize it first before merging this.

haydentherapper:
@joshuagl - Do you know when this release process will be formalized? I agree that it'd be nice to document breaking changes in a changelog. One thing to note is this would be a noticeable breaking change, so users who pull in the latest changes would be able to recognize that `Init`'s params have changed.

+1 to @asraa's comment about this following the spec. It calls out that you should be loading a "trusted root metadata file", not just the root keys. It's worth mentioning again that ideally by default, `Init` would just "load", by initializing state from a known root, and update mechanisms should be kept separate.

@hosseinsia, as part of the refactor, we can change all of our usages of `client.Init`. I see we're using it in the `tuf-client init` [command](https://github.com/theupdateframework/go-tuf/blob/master/cmd/tuf-client/init.go). We can deprecate the usage of the `root-keys-file` flag and switch to a `root-metadata` flag. This should actually make the command easier to use, since the [keys format](https://github.com/theupdateframework/go-tuf/blob/master/data/types.go#L42) has some complexity, and users should always have the root metadata file available. Note that this won't impact `tuf init`, since that uses `repo.Init`.

client/client.go
haydentherapper: Added!

haydentherapper:
I've updated the PR, changing `Init` to only take the root metadata JSON. This commit could be split into its own PR, but I think that we should decide when to make the breaking change to change the parameters of `Init`. My preference would be making the change in this PR.

asraa (dismissed):
ping on this! @hosseinsia 

I would even by happy to keep Init as before and make a new method like `InitLocal`, and then deprecate/rename `Init` on a release?

hosseinsia:
@asraa 
Yes please. That would not break us and makes the transition much smoother. Thank you!

haydentherapper:
@hosseinsia, how should we communicate the differences between `Init` and `InitLocal` to convince users to switch over? And furthermore, if users do start switching over to `InitLocal`, then we'll end up breaking them when we remove the method in the future.

This change shouldn't break anyone who's using a pinned version of go-tuf. Once someone updates to the latest commit in go-tuf, their build will break, but it should be clear what the issue is. Either we can include changelog with this commit, or just assume that the documentation on the function is clear.

It's also worth noting that impact is low according to https://cs.github.com/?scopeName=All+repos&scope=&q=%22github.com%2Ftheupdateframework%2Fgo-tuf%2Fclient%22+AND+%22.Init%22+AND+-path%3Acosign+AND+-repo%3Atheupdateframework%2Fgo-tuf - I've looked through these and mostly see either usage with Sigstore maintainers or go-tuf maintainers.

I'd prefer that we make the breaking change now. This will make Init align to the TUF specification, since Init does not go through the secure root update flow.

hosseinsia:
Potentially, in a release note, we can specify that the `InitLocal` is recommended over the existing method. Also, we can update the comment the `init` and tag it as deprecated to decentivize the usage (see deprecated here https://go.dev/blog/godoc). regardless of whether we keep or remove the `init`, we have to do it in two PR to make the roll-back easier. WDYT?

haydentherapper:
@hosseinsia I've updated the PR and added a deprecation comment to Init. Please take another look! 

I'll leave #208 open until a versioning system is in place, when we can revisit this issue and remove `Init`.

hosseinsia (approved):
LGTM. Thanks!

asraa:
@hosseinsia staticcheck doesn't like seeing usage of `Init` in the codebase now that it's deprecated. Can we leave `Init` as an public method of the client API, and switch over the CLI and tests to use `InitLocal`? That shouldn't break your usage of go-tuf's Client as it's deprecated. 

(Also, this still won't break anyone unless they manually update the pinned sha of go-tuf)

@trishankatdatadog

hosseinsia:
@asraa good resolution. Thanks!

haydentherapper:
Thanks all for the feedback! Updated the client to switch to InitLocal.

asraa (approved):

hosseinsia (approved):
LGTM

PR #213, opened 2022-01-27
make ED25519PrivateKeyValue public, add a way to instantiate ed25519Signer from it

arbll:
# What does this PR do ?

1. Makes `Ed25519PrivateKeyValue` public
2. Adds a way to instantiate ed25519Signer from `Ed25519PrivateKeyValue`

With the change of interface for the keys there is no non-hacky way to use an ed25519 key that is not stored as a `go-tuf` specific marshaled JSON blob without re-implementing the signer interface.

### DIFF
--- a/pkg/keys/ed25519.go
+++ b/pkg/keys/ed25519.go
@@ -54,7 +54,7 @@ func (e *ed25519Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
 	return nil
 }
 
-type ed25519PrivateKeyValue struct {
+type ED25519PrivateKeyValue struct {
 	Public  data.HexBytes `json:"public"`
 	Private data.HexBytes `json:"private"`
 }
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
@hosseinsia @cedricvanrompay-datadog

pkg/keys/ed25519.go
```diff
@@ -83,12 +83,21 @@ func GenerateEd25519Key() (*ed25519Signer, error) {
 	}, nil
 }
 
+func Ed25519Key(keyValue Ed25519PrivateKeyValue) *ed25519Signer {
```
ethan-lowman-dd, line 86: nit: constructor convention
```suggestion
func NewEd25519Signer(keyValue Ed25519PrivateKeyValue) *ed25519Signer {
```

asraa (dismissed):
Nice! I think this is totally valid for a use case where you've done offline key generation.

Generally speaking, do we want some kind of import naming/method/something else? I'm happy with this, maybe the CLI should also allow for key imports at some point.

pkg/keys/ed25519.go
asraa: +1, potentially also `Import`?

ethan-lowman-dd (approved):

asraa (approved):

pkg/keys/ed25519.go
arbll: You mean `Import` instead of `New` @asraa ? Making sure I understand

asraa: Yeah, but I'm not picky with wording :)

PR #214, opened 2022-02-01
feat: add `payload` and `add-signature` commands.

### LINKED ISSUE #205: TUF CLI: Improve experience by allowing offline signature generation
This would help using the TUF CLI in more tightly controlled key environments, e.g. an air-gapped situation.

I propose two methods:

`tuf payload <role>` that outputs the canonicalized, signable content of the role's metadata file
`tuf add-signature <role> --signature <sig_file>` that adds a signature (hex-encoded bytes), generated by an offline tool, to the role. This should verify the signature first.

@haydentherapper

znewman01:
Fixes #205.


znewman01:
ping @asraa (not to nag/rush, just want to make sure it's on your radar)

znewman01:
Oh also CC @haydentherapper

repo.go
```diff
@@ -1092,3 +1092,17 @@ func (r *Repo) timestampFileMeta(roleFilename string) (data.TimestampFileMeta, e
...
 	return util.GenerateTimestampFileMeta(bytes.NewReader(b), r.hashAlgorithms...)
 }
+
+func (r *Repo) Payload(roleFilename string) ([]byte, error) {
+	role := strings.TrimSuffix(roleFilename, ".json")
+	if !roles.IsTopLevelRole(role) {
+		return nil, ErrInvalidRole{role}
+	}
+
+	s, err := r.SignedMeta(roleFilename)
```
rdimitrov, line 1102: It's a small nit, but I noticed the `tuf payload` command implicitly expects a role filename instead of what's in its description - i.e., `root.json` instead of `root`. Adding a test for that would be helpful too 👍

znewman01: Added a test that we get `ErrMissingMetadata` and changed help text to expect `<metadata>` instead of `<role>`

cmd/tuf/add_signature.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/flynn/go-docopt"
+	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
+)
+
+func init() {
+	register("add-signature", cmdAddSignature, `
+usage: tuf add-signature <role> --key-id <key_id> --signature <sig_file>
+
+Adds a signature (as hex-encoded bytes) generated by an offline tool to the given role.
```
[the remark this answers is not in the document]
znewman01: Good call--this is sort-of meaningless without the "offline tool" implemented.

I've added `sign-payload` and verified that the `payload` -> `sign-payload` -> `add-signatures` flow works. I've also clarified the help text here to be agnostic to the format; the requirement is just that it comes from `sign-payload`.

znewman01:
I've pushed a couple commits to fix the issues mentioned; of note, I add a `sign-payload` command which is the "offline tool" we've been dancing around.

repo.go
```diff
@@ -519,36 +520,49 @@ func (r *Repo) setTopLevelMeta(roleFilename string, meta interface{}) error {
 	return r.local.SetMeta(roleFilename, b)
 }
 
-func (r *Repo) Sign(roleFilename string) error {
-	role := strings.TrimSuffix(roleFilename, ".json")
+// Use the keys associated with role to sign the payload in signed.
+func (r *Repo) SignPayload(role string, signed *data.Signed) (int, error) {
 	if !roles.IsTopLevelRole(role) {
-		return ErrInvalidRole{role, "only signing top-level metadata supported"}
+		return -1, ErrInvalidRole{role, "only signing top-level metadata supported"}
```
rdimitrov, line 526: Maybe use something better than -1?

repo.go
```diff
@@ -519,36 +520,49 @@ func (r *Repo) setTopLevelMeta(roleFilename string, meta interface{}) error {
...
 	}
 
-	s, err := r.SignedMeta(roleFilename)
+	keys, err := r.getSortedSigningKeys(role)
 	if err != nil {
-		return err
+		return -1, err
```
rdimitrov, line 531: Same here 👍

repo.go
```diff
@@ -519,36 +520,49 @@ func (r *Repo) setTopLevelMeta(roleFilename string, meta interface{}) error {
...
+	}
+	if len(keys) == 0 {
+		return 0, ErrInsufficientKeys{role}
 	}
+	for _, k := range keys {
+		if err = sign.Sign(signed, k); err != nil {
+			return -1, err
```
rdimitrov, line 538: Same here too 👍

rdimitrov (commented):
I'd also suggest updating the list of commands in the README.md file by adding their descriptions along with examples of how to use them 👍

repo.go
znewman01: Happy to change this, but my understanding is that in Go it's [idiomatic to totally ignore the success-case return value](https://stackoverflow.com/questions/46726832/how-to-return-int-or-nil-in-golang) if an error is passed. In fact, I see [Go by Example](https://gobyexample.com/errors) does exactly this, with a `-1` return value.

What do you propose I do instead?

I've improved the documentation for this function to be clearer about what it returns in these cases.

README.md
```diff
@@ -229,6 +250,46 @@ Enter root keys passphrase:
...
+
+``` bash
+$ tuf payload root.json > root.json.payload
+```
+
+Copy `root.json.payload` to the root box and sign it:
+
+
+``` bash
+$ tuf sign-payload --role=root root.json.payload > root.json.sigs
```
asraa, line 268: Could you point out where the keys should be stored if you are doing this? e.g. this expected a folder located like `keys/root.json` containing key information (output by gen-key)

repo.go
```diff
@@ -1062,3 +1080,17 @@ func (r *Repo) timestampFileMeta(roleFilename string) (data.TimestampFileMeta, e
...
 	return util.GenerateTimestampFileMeta(bytes.NewReader(b), r.hashAlgorithms...)
 }
+
+func (r *Repo) Payload(roleFilename string) ([]byte, error) {
+	role := strings.TrimSuffix(roleFilename, ".json")
+	if !roles.IsTopLevelRole(role) {
+		return nil, ErrInvalidRole{role, "only signing top-level metadata supported"}
+	}
+
+	s, err := r.SignedMeta(roleFilename)
```
asraa, line 1090: should this make sure this is json canonicalized?
it seems like it's just json unmarshalled into data.Signed
https://github.com/theupdateframework/go-tuf/blob/0e25bb50a25b479318f82432dc61ec5461c28322/repo.go#L675-L678

It seems like in most places we set the meta to be the canonicalized form (https://github.com/theupdateframework/go-tuf/blob/0e25bb50a25b479318f82432dc61ec5461c28322/repo.go#L511) but what if someone uses a local store and edits the metadata retrieved?

Maybe I should open a separate issue about this, because I'm not convinced we canonicalize consistently.

asraa (approved):
Just some minor comments!

repo.go
znewman01: > and I wonder if having that number of keys returned is necessary? 

I think it might be useful for programmatic callers. I might want to double check that I've signed with the same number of keys that I expected.

> As far as I see we only use this value for logging a message, so we should be safe to return only the err value and log numKeys again whenever it is necessary from within the SignPayload() function. What are your thoughts on that?

I'd strongly prefer to avoid printing messages that we expect users to see within the library calls. Debug messages etc. are reasonable, but the output of the command should come from the command itself. Otherwise, applications might see our output in the middle of their operation.

So my vote is to leave it as-is. I would also be willing remove the number of keys from the output and the return value if you feel very strongly.

repo.go
znewman01: See my comments in #247. I do think you're broadly right.

That said, I think there's enough in-flight at the moment that we should probably open a separate issue and figure out a consistent strategy for canonicalization there. I'll go ahead and do that.

znewman01: https://github.com/theupdateframework/go-tuf/issues/261

repo.go
rdimitrov: > I think it might be useful for programmatic callers. I might want to double check that I've signed with the same number of keys that I expected.

I'm okay with keeping it then if there's such a use case 👍 

> I'd strongly prefer to avoid printing messages that we expect users to see within the library calls. Debug messages etc. are reasonable, but the output of the command should come from the command itself. Otherwise, applications might see our output in the middle of their operation.
> 
Agree 👍 Makes sense.

rdimitrov (approved):

znewman01:
No substantial changes but looks like I need a re-review, @asraa 

Otherwise I think this is good to go except I need one more approval from a maintainer—@trishankatdatadog any chance you could give a quick review?

trishankatdatadog:
> Otherwise I think this is good to go except I need one more approval from a maintainer—@trishankatdatadog any chance you could give a quick review?

Sorry, no time 😕 Could someone else from DD take over (@ethan-lowman-dd @hosseinsia)?

README.md
```diff
@@ -130,6 +130,28 @@ Changes the passphrase for given role keys file. The CLI supports reading
 both the existing and the new passphrase via the following environment
 variables - `TUF_{{ROLE}}_PASSPHRASE` and respectively `TUF_NEW_{{ROLE}}_PASSPHRASE`
 
+#### `tuf payload <metadata>`
+
+Outputs the metadata file for a role in a ready-to-sign (canonicalized) format.
```
asraa, line 135: FYI sorry about the badgering about the canonicalization again but if https://github.com/theupdateframework/go-tuf/pull/247/files?file-filters%5B%5D=.go&show-viewed-files=true is merged first (trynna get that in...), then this won't hold true -- metadata to the store won't be written in canonical form ahead of time.

Currently the tests use `SignPayload` which will internally canonicalize -- do you think you could add a test demonstrating that Payload() is canonical? Maybe show Canonicalize is no-op to the output

asraa (dismissed):

README.md
znewman01: Ha, fair. Added [ee3e8e](https://github.com/theupdateframework/go-tuf/pull/214/commits/ee3e8ee3c70b0c35316338f53b1ab4afb66e396e) to check. (Turns out that `cjson.EncodeCanonical` is not idempotent!)

znewman01:
Rebased and added a commit [ee3e8e](https://github.com/theupdateframework/go-tuf/pull/214/commits/ee3e8ee3c70b0c35316338f53b1ab4afb66e396e) that adds a test and fixes it in light of #247

rdimitrov (approved):
LGTM 👍

repo.go
```diff
@@ -507,36 +509,53 @@ func (r *Repo) setTopLevelMeta(roleFilename string, meta interface{}) error {
...
 	}
 
-	s, err := r.SignedMeta(roleFilename)
+	keys, err := r.getSortedSigningKeys(role)
 	if err != nil {
-		return err
+		return -1, err
+	}
+	if len(keys) == 0 {
+		return 0, ErrInsufficientKeys{role}
```
cedricvanrompay-datadog, line 527: Nit: I know this was already the case before this PR, but isn't it weird that the error is named "insufficient keys" when it is only returned when *no keys* where found? Shouldn't it be called `ErrNoKeys`? I would expect an "insufficient keys" errors to be returned when the number of keys we found is lesser than the threshold for this role.

cedricvanrompay-datadog (approved):
On behalf of Datadog: looks good to me.

I agree that [it would be better to make sure the data we sign is properly canonicalised](https://github.com/theupdateframework/go-tuf/pull/214#discussion_r840653371), I saw that a dedicated issue has been created.

cedricvanrompay-datadog:
@asraa @trishankatdatadog approved on behalf of Datadog

repo_test.go
```diff
@@ -1867,3 +1875,48 @@ func (rs *RepoSuite) TestSignDigest(c *C) {
...
+	c.Assert(err, IsNil)
+	c.Assert(numKeys, Equals, 1)
+
+	// Add the payload signatures back
+	for _, sig := range signed.Signatures {
+		// This method checks that the signature verifies!
+		err = r.AddOrUpdateSignature("root.json", sig)
+		c.Assert(err, IsNil)
+	}
+}
```
cedricvanrompay-datadog, line 1922: Shouldn't we test that a client would accept the generated output?

repo.go
znewman01: Renamed `ErrInsufficientKeys` to `ErrNoKeys`

README.md
```diff
@@ -130,6 +130,28 @@ Changes the passphrase for given role keys file. The CLI supports reading
...
+
+See also `tuf sign-payload` and `tuf add-signatures`.
+
+#### `tuf sign-payload --role=<role> <path>`
+
+Sign a file (outside of the TUF repo) using keys (in the TUF keys database,
+typically produced by `tuf gen-key`) for the given `role` (from the TUF repo).
+
+Typically, `path` will be the output of `tuf payload`.
```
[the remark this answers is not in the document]
znewman01: Updated the documentatation

znewman01:
Need quick re-reviews again

rdimitrov (dismissed):

trishankatdatadog:
> Need quick re-reviews again

@cedricvanrompay-datadog would you pls mind reviewing?

On that note, I think it's worth nominating @znewman01 and @cedricvanrompay-datadog as additional maintainers. What do the @theupdateframework/go-tuf-maintainers think?

asraa (dismissed):
Thanks for the test that `Payload` is canonicalized!

asraa:
> On that note, I think it's worth nominating @znewman01 and @cedricvanrompay-datadog as additional maintainers. What do the https://github.com/orgs/theupdateframework/teams/go-tuf-maintainers think?

Def +1 to Zack, solid reviews, issue filing, and PRs.

znewman01:
@cedricvanrompay-datadog it won't let me re-request a review for you for some reason so tagging you in manually

znewman01:
Gentle ping @cedricvanrompay-datadog @asraa

asraa (approved):

PR #230, opened 2022-03-04
Rename assertNotNil to assertNoError, since the former name is incorrect

ethan-lowman-dd:
The `assertNotNil` helper functions were actually asserting that the error *was* nil. This PR renames the functions to reflect their actual behavior.

### DIFF
--- a/client/testdata/go-tuf-transition-M3/generate.go
+++ b/client/testdata/go-tuf-transition-M3/generate.go
@@ -22,7 +22,7 @@ type persistedKeys struct {
 	Data      []*data.PrivateKey `json:"data"`
 }
 
-func assertNotNil(err error) {
+func assertNoError(err error) {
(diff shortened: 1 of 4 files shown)


cedricvanrompay-datadog (approved):

joshuagl (approved):
LGTM, thanks @ethan-lowman-dd.

I assume we're carrying four copies of the same function to avoid the overhead of defining some kind of test utils module (package?)?

ethan-lowman-dd:
@joshuagl I assume so. I'd rather not add another testing module, since the codebase already uses at least two (https://pkg.go.dev/github.com/stretchr/testify and https://pkg.go.dev/gopkg.in/check.v1)

trishankatdatadog (approved):

asraa (approved):

PR #226, opened 2022-02-24
A first cut of automated releases using goreleaser

trishankatdatadog:
Thanks to #157 for inspiration (dropped SBOMs and signing now for simplicity).

Checklist:
- [x] Release client vs all
- [ ] Changelog (including a convention for marking PRs with backwards-incompatible changes)
- [ ] Prereleases
- [ ] Define convention to mark backwards-incompatible changes (despite using a MAJOR SemVer version of 0)
- [ ] Test with pushing a tag

Deferred:
* SBOMs
* Sign and add to Sigstore (besides gosumdb)

PR #190, opened 2021-12-09
docs: Add contributing guidelines for go-tuf

asraa:
Include PR process, review process, and other communication details.

Signed-off-by: Asra Ali <asraa@google.com>

PR #234, opened 2022-03-15
Add automatic releases using goreleaser

rdimitrov:
The following PR is a continuation of the effort in #226 by @trishankatdatadog with some additional changes. If this one gets merged, #226 can be closed.

**Details:**

1. The configuration is refactored so it leverages reusable workflows - i.e for releasing the CLIs

- `Tests` run on push and pull requests while `Release` runs only on tags being pushed. SemVer is assumed, but not enforced. These configs were tested in another private repo.

3. There are separate groups created based on keywords in the commit messages - breaking, fix, feat, and other.

- There's an explicit `Braking changes` group created every time there’s a breaking change introduced in the given release (has to be stated either by the author of the PR or by the maintainer merging it)

4. Builds a changelog in `GitHub`-style manner
- Lists commit hashes and messages but also tags their authors in the changelog resulting in a nice-looking chapter with everyone’s photos. This appreciates the people who contributed to the given release and helps promote other to contribute too to the project. It looks nice too.

**Related topics to discuss:**
Before merging it would be nice if we have the following decided and merged in the guideline too.

Define the convention in our contributing guidelines for:
- Marking changes as `fix`, `feature`, or `breaking` in our contributing guidelines in #190 
- How to tag prereleases (I'd suggest `v0.1.0-rc1`)
- How to tag breaking changes (I'd suggest incrementing the MAJOR version)


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          distribution: goreleaser
+          version: "v1.2.5"
```
shibumi, line 27: You want to bump the version

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: runner.os == 'Linux'
+        run: test -z $(go fmt ./...)
+      - name: Test
+        run: go test -race -covermode atomic -coverprofile='profile.cov' ./...
+      - name: Send coverage
+        if: runner.os == 'Linux'
+        env:
+          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          GO111MODULE=off go get github.com/mattn/goveralls
```
shibumi, line 35: Is there a specific reason why you disable Go Modules? :thinking:

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+          $(go env GOPATH)/bin/goveralls -coverprofile=profile.cov -service=github
+      - name: Vet
+        run: go vet ./...
+      - name: Install staticcheck
+        run: "go install honnef.co/go/tools/cmd/staticcheck@v0.2.2"
```
shibumi, line 40: Wonder if it would make sense to use golangci-lint instead of staticcheck. It comes with staticchecks but a ton of useful other linters (including security related linters)

.goreleaser/tuf-client.yml
```diff
@@ -0,0 +1,33 @@
...
+      regexp: "^.*breaking[(\\w)]*:+.*$"
+      order: 0
+    - title: Features
+      regexp: "^.*feat[(\\w)]*:+.*$"
+      order: 1
+    - title: "Bug fixes"
+      regexp: "^.*fix[(\\w)]*:+.*$"
+      order: 2
+    - title: Others
+      order: 999
```
shibumi, line 33: What happened to the SBOM generation and ephemeral signature creation?

shibumi:
@rdimitrov left a few comments. Can you elaborate why you removed the SBOM and signature creation? I am not sure if releasing binaries without signatures matches the security requirements of TUF, but @trishankatdatadog  might have more to say about this :D 

IMO, as long as the signature creation is not properly discussed it would still make sense to create SBOMs for the projects.
Having SBOMs doesn't block anything and I can only see benefits of generating SBOMs :thinking:

trishankatdatadog:
> @rdimitrov left a few comments. Can you elaborate why you removed the SBOM and signature creation? I am not sure if releasing binaries without signatures matches the security requirements of TUF, but @trishankatdatadog might have more to say about this :D

Not a priority right now: we need to nail down a release process advertising backwards-incompatible changes first, then we can do all else. I also don't see the value of signing go-tuf releases on Rekor when Go modules get added to the gosumdb anyway, and go tooling doesn't support cosign (yet) but they do support gosumdb.

.goreleaser/tuf.yml
```diff
@@ -0,0 +1,33 @@
+project_name: tuf
+builds:
+  - ldflags:
+      - "-s -w"
+      - "-extldflags=-zrelro"
+      - "-extldflags=-znow"
+    env:
+      - "CGO_ENABLED=0"
+      - "GO111MODULE=on"
+      - "GOFLAGS=-mod=readonly -trimpath"
```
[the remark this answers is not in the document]
rdimitrov: @developer-guy - Thanks for suggesting that 👍 Applied it!

.github/workflows/release.yml
rdimitrov: @shibumi - Thanks! 👍 Bumped it to v1.6.3

.github/workflows/tests.yml
rdimitrov: @shibumi - I think it was copied from the initial PR. Anyway, I switched to using the goveralls GitHub Action instead. Shouldn't make a difference, but it looks cleaner the least 👍

.github/workflows/tests.yml
rdimitrov: @shibumi - Actually, I was planning to add that in another PR as I was afraid it was going to introduce a lot of additional changes to this one, but now that it was suggested - I tried it and it turned out to be okay. I kept the same linters we had so far and added a few others. At some point, we can enable more linters in a separate PR along with fixes for the issues they raise.

trishankatdatadog:
Thanks @rdimitrov and friends! Have we decided how to answer these Qs?

> Define the convention in our contributing guidelines for:
> 
> * Marking changes as `fix`, `feature`, or `breaking` in our contributing guidelines in [docs: Add contributing guidelines for go-tuf #190](https://github.com/theupdateframework/go-tuf/pull/190)
> * How to tag prereleases (I'd suggest `v0.1.0-rc1`)
> * How to tag breaking changes (I'd suggest incrementing the MAJOR version)

rdimitrov:
> Thanks @rdimitrov and friends! Have we decided how to answer these Qs?
> 
> > Define the convention in our contributing guidelines for:
> > 
> > * Marking changes as `fix`, `feature`, or `breaking` in our contributing guidelines in [docs: Add contributing guidelines for go-tuf #190](https://github.com/theupdateframework/go-tuf/pull/190)
> > * How to tag prereleases (I'd suggest `v0.1.0-rc1`)
> > * How to tag breaking changes (I'd suggest incrementing the MAJOR version)

@trishankatdatadog - I've just added a comment to @asraa in #190 if we can address these in the same PR, otherwise we can merge it and I'll prepare a separate one. As a preliminary check, do we have any comments on the suggested conventions, or we can agree on them?

shibumi:
> Marking changes as fix, feature, or breaking in our contributing guidelines in https://github.com/theupdateframework/go-tuf/pull/190

The goreleaser configuration in this PR is covering this already: https://github.com/theupdateframework/go-tuf/pull/234/files#diff-2e9100addf89d1dc73418f538ce4f3d8fcea7d07ab9379f7309e6ce4f4c2cd83R25

It is just key that every dev uses conventional commits in their commit messages.... I would suggest adding some sort of linter that checks git commit messages for conventional commits.

Because, if you don't use conventional commits, all commits will be sorted into the "Others" section and this is not what you want when you try to generate the changelog via Goreleaser.



> How to tag prereleases (I'd suggest v0.1.0-rc1)

GoReleaserer covers this as well via a `prerelease` flag in the configuration. The default flag value is `false`. You might want to add `prerelease: auto` to your GoReleaser configuration. This will automatically identify a semver version with suffix as a prerelease: https://goreleaser.com/customization/release/ 


> How to tag breaking changes (I'd suggest incrementing the MAJOR version)

Conventional commits and Semver are the documents you want to have a look at. Conventional commits suggest adding a `Breaking:` line to the commit message and Semver suggests that breaking changes should always lead to a MAJOR version increase.

See also: https://www.conventionalcommits.org/en/v1.0.0/#:~:text=in%20Semantic%20Versioning).-,breaking%20change%3A,-a%20commit%20that

And: https://semver.org/

trishankatdatadog:
Agree with both of your recommendations about marking commits and preleases.

> Conventional commits and Semver are the documents you want to have a look at. Conventional commits suggest adding a `Breaking:` line to the commit message and Semver suggests that breaking changes should always lead to a MAJOR version increase.

Also agree with this, but want to make sure we can make breaking changes now without bumping MAJOR version from 0, which I understand is possible with SemVer.

rdimitrov:
> Agree with both of your recommendations about marking commits and preleases.
> 
> > Conventional commits and Semver are the documents you want to have a look at. Conventional commits suggest adding a `Breaking:` line to the commit message and Semver suggests that breaking changes should always lead to a MAJOR version increase.
> 
> Also agree with this, but want to make sure we can make breaking changes now without bumping MAJOR version from 0, which I understand is possible with SemVer.

@trishankatdatadog - Agree 👍 As per the [spec](https://semver.org/#spec-item-4), everything should be considered as breakable until we have a MAJOR version of 1. 

@shibumi - I was imagining a less strict process of using conventional commits, but you have a point that it may easily become useless unless we use it as it should. 

I've already added the flag for automatic prereleases and I'll see a bit later today what are the available linters for enforcing conventional commit messages 👍

shibumi:
Small note:

> Also agree with this, but want to make sure we can make breaking changes now without bumping MAJOR version from 0, which I understand is possible with SemVer.

The goreleaser conventional commit changes nor the semver would prevent us from not bumping MAJOR. It's absolutely up to us how we "interpret" the semver definition and how we release it...


What GoReleaser does with conventional commits and SemVer are just two things:

1. It will create a prerelease if the version tag has a suffix. That means everything in SUFFIX of  MAJOR.MINOR.PATCH-SUFFIX qualifies the version to be prerelease. Prereleases are treated differently on GitHub (they are explicitly marked as prerelease and not shown as "latest release").

2. GoReleaser creates a changelog via reading the Git commit message headers. So for example a commit like: "fix: bug in go-tuf" would land in the "Bugfixes" section of the GoReleaser changelog.

If you want an example how this look like. Have a look on the Testifysec/Witness repository, where I have implemented such a pipeline already: https://github.com/testifysec/witness/releases/tag/v0.1.2

local_store.go
```diff
@@ -315,11 +315,8 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 			return err
 		}
 		if !info.IsDir() && isTarget(rel) && needsRemoval(rel) {
-			//lint:ignore SA9003 empty branch
-			if err := os.Remove(path); err != nil {
-				// TODO: log / handle error
-			}
-			// TODO: remove empty directory
+			// TODO: log / handle error or remove empty directory
+			_ = os.Remove(path)
```
trishankatdatadog, line 319: This looks dangerous. Can we restore the original code but return the err?

trishankatdatadog (changes requested):

local_store.go
rdimitrov: In that case, should we keep or disable the linters raising these issues (for ref. when enabled this will cause the golangci-lint job to fail for this PR)?

shibumi: @rdimitrov I would fix this in a separate PR.. merge this before this one.. and then merge this one on top.

rdimitrov: Reverted part of the commit that introduced these changes and addressed them in https://github.com/theupdateframework/go-tuf/pull/236

rdimitrov:
FYI golangci-lint now fails because of https://github.com/theupdateframework/go-tuf/pull/236

trishankatdatadog (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
```
joshuagl, line 22: 1.18 is the latest, how do we choose which version to use here? Is the default the latest release?
If we pin to a specific version here, how do we monitor for new versions?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,61 @@
+on:
+  workflow_call:
+name: Tests
+jobs:
+  run:
+    strategy:
+      matrix:
+        go-version: [1.16.x, 1.17.x]
```
joshuagl, line 8: same here, how do we monitor these versions and keep them update (1.18.x is now out).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade iso8601 requests securesystemslib six tuf
```
joshuagl, line 27: Should we be pinning the python dependencies, so that a release of one of these python packages doesn't cause a change in CI behaviour? (counter-argument – if we pin them, we need to monitor and update them).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: shogo82148/actions-goveralls@v1
+        with:
+          path-to-profile: profile.cov
+          flag-name: Go-${{ matrix.go-version }}
+          parallel: true
+
+  golangci:
+    strategy:
+      matrix:
+        go-version: [1.16.x, 1.17.x]
```
joshuagl, line 42: Similar question here on choosing and monitoring go versions.

joshuagl (commented):
This ended up being a larger set of changes than I anticipated. Thanks for the contributions @shibumi @trishankatdatadog @rdimitrov !

Can we pin actions by digest, rather than tag (release name)? Tags are mutable, so we could theoretically get different versions. We should also have dependabot monitor the actions for updates (see i.e. https://github.com/theupdateframework/python-tuf/blob/b7b035aea12048596f62116f29e95f2551efa9be/.github/dependabot.yml#L11)

.github/workflows/release.yml
rdimitrov: Go doesn't release that frequently so we should be on the safe side when we pin to a relatively new version. 

Anyway, figured out how to run the CI against all available go versions newer than the one we specify in the `go.mod` file, so that should significantly reduce the number of places where we pin go versions and thus always give a sense if we are okay against the latest one 👍

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          distribution: goreleaser
+          version: "v1.6.3"
```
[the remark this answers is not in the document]
rdimitrov: Apparently, goreleaser releases frequently, because when I opened the PR 1.6.3 was the latest one. I'm open to suggestions 👍 

Latest is the default, but I'd suggest pinning to a specific version rather than using "latest" as it improves the consistency of the CI behavior.

.github/workflows/tests.yml
rdimitrov: Addressed it 👍

.github/workflows/tests.yml
rdimitrov: Addressed it via a requirements file which is also monitored by dependabot 👍

.github/workflows/tests.yml
rdimitrov: Addressed it 👍

.github/workflows/release.yml
rdimitrov: The result is:

- Tests will run against every go version starting from the one we say we support in the `go.mod` file. 
- Releases will use the one we specify in the `go.mod` file

.github/dependabot.yml
```diff
@@ -1,7 +1,17 @@
...
+      time: "10:00"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
```
joshuagl, line 18: nice! should we have these commit-message settings for "gomod" too?

.github/workflows/tests.yml
```diff
@@ -2,21 +2,34 @@ on:
...
 name: Tests
 jobs:
+  get-go-versions:
+    name: Collect available Go versions
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.versions.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - uses: arnested/go-version-action@v1
```
joshuagl, line 12: could you pin this one by digest? :-)

.github/workflows/tests.yml
```diff
@@ -35,11 +35,11 @@ jobs:
         with:
           python-version: 3.6
           cache: "pip"
-          cache-dependency-path: "requirements-test.txt"
```
joshuagl: Is requirements-test.txt cached by default?

.github/dependabot.yml
joshuagl: Let's do this in a follow-up PR.

.github/workflows/tests.yml
rdimitrov: Done 👍

.github/workflows/tests.yml
rdimitrov: Oh, right, forgot that I changed the file name. Reverting that 👍

joshuagl (approved):
This is huge, let's get it merged. Looks like an excellent starting point. Thank you everyone!

trishankatdatadog (approved):

trishankatdatadog:
@joshuagl before we merge this, we first need to merge #236, no?

joshuagl:
Absolutely right @trishankatdatadog

trishankatdatadog:
> Absolutely right @trishankatdatadog

If you would please help fix and approve it, I'll approve that also and then we'll get both merged, thx

trishankatdatadog:
Let's also close #226 and #157 after this gets merged

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_call:
+    inputs:
+      cli-name:
+        required: true
+        type: string
+name: Release
+jobs:
+  release:
+    permissions:
+      id-token: write
```
shibumi, line 11: I think this line is not necessary anymore, because you don't use the OIDC issuer for signing in this PR.

It might make sense to remove this line to reduce blast radius. Right now it exposes the Github OIDC token for nothing.

CC: @trishankatdatadog 

Sorry for coming up with this after the merge. I spent the last week with finishing my master degree in CS :partying_face:

trishankatdatadog: > It might make sense to remove this line to reduce blast radius. Right now it exposes the Github OIDC token for nothing.

Good catch! @rdimitrov can we fix this ASAP?

> Sorry for coming up with this after the merge. I spent the last week with finishing my master degree in CS 🥳

Congratulations! ❤️ 🚀

rdimitrov: @trishankatdatadog - Got it - https://github.com/theupdateframework/go-tuf/pull/255 👍

ethan-lowman-dd:
One of the effects of this PR is that you have to format every commit message using conventional commits, even though we squash merge PRs. I don't think commits that make up a PR need to be that rigorously named, since the history is squashed on merge and work-in-progress commits probably don't belong in a changelog.

Could we instead only check that the PR titles follow conventional commit format?

shibumi:
@ethan-lowman-dd your information is partly incorrect. Developers can decide if they use conventional commits.. if they do the commit will get sorted into the correct section in the Changelog. If they do not follow, the commit will **still** land in the Changelog, it will just be a in a  "general" section.

Using PR titles is not possible, because goreleaser only reads the commit message titles.

ethan-lowman-dd:
See the CI failures https://github.com/theupdateframework/go-tuf/runs/5841476493?check_suite_focus=true on PR https://github.com/theupdateframework/go-tuf/pull/175 -- ths appears to enforce the format for all commit messages.

Is the intention to merge PRs despite failure of this CI check?

ethan-lowman-dd:
> Using PR titles is not possible, because goreleaser only reads the commit message titles.

PR titles become part of the squashed commit message upon merge. Isn't goreleaser only run from `main` or a version branch with cherry picked PR commits, not a PR feature branch with unsquashed commits?

trishankatdatadog:
Hmm, let's sort this out one way or another.

rdimitrov:
Having several commits like "test", "test1" and so on is definitely not a best practice, so it should be something we aim to deprecate from ending up in the squashed commit message anyway.

I'll have a look at the linter we are using now and see if it's possible to check for the PR title only and ignore the commits in it. Considering that we only do squash and merge, this is indeed the only thing that goreleaser takes into account while generating the changelog.

ethan-lowman-dd:
Please see my suggested change here: https://github.com/theupdateframework/go-tuf/pull/264/

PR #236, opened 2022-03-18
Fix linter errors raised by staticcheck

rdimitrov:
This is a small PR separating a few fixes for linter errors that would otherwise be raised by golangci-lint/staticcheck in #234. 

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

### DIFF
--- a/local_store.go
+++ b/local_store.go
@@ -431,11 +431,9 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 			return err
 		}
 		if !info.IsDir() && isTarget(rel) && needsRemoval(rel) {
-			//lint:ignore SA9003 empty branch
 			if err := os.Remove(path); err != nil {
-				// TODO: log / handle error
+				return err
 			}
-			// TODO: remove empty directory
 		}
 		return nil
 	}
--- a/local_store.go
+++ b/local_store.go
@@ -410,6 +410,7 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 		}
 		return nil
 	}
+	// Checks if target file should be deleted
 	needsRemoval := func(path string) bool {
 		if consistentSnapshot {
 			// strip out the hash
@@ -422,6 +423,19 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 		_, ok := hashes[path]
 		return !ok
 	}
+	// Checks if folder is empty
+	folderNeedsRemoval := func(path string) bool {
+		f, err := os.Open(path)
+		if err != nil {
+			return false
+		}
+		defer f.Close()
+		_, err = f.Readdirnames(1)
+		if err == io.EOF {
+			return true
+		}
+		return false
+	}
 	removeFile := func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


local_store.go
```diff
@@ -431,11 +431,9 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
 			return err
 		}
 		if !info.IsDir() && isTarget(rel) && needsRemoval(rel) {
-			//lint:ignore SA9003 empty branch
 			if err := os.Remove(path); err != nil {
-				// TODO: log / handle error
+				return err
```
joshuagl, line 435: We were ignoring file removal failures and now, with this change, we are not? That seems like a reasonable change, are there likely to be any unexpected consequences of this change?

local_store.go
```diff
@@ -431,11 +431,9 @@ func (f *fileSystemStore) Commit(consistentSnapshot bool, versions map[string]in
...
 			}
-			// TODO: remove empty directory
```
joshuagl, line 438: This seems worth implementing, rather than just removing the comment?

trishankatdatadog (approved):

joshuagl (commented):
Thanks for working on fixing inter failures. I'd like a second set of eyes on the `LocalStore` changes.

local_store.go
trishankatdatadog: Good question. Maybe we should add a test case, if it's missing?

trishankatdatadog (changes requested):
ditto Joshua's comments

local_store.go
rdimitrov: Addressed it 👍

local_store.go
rdimitrov: I tested several use cases and I didn't find a reason to break something. It's good to leave this check though and address it again in case someone raises a valid reason to change it back 👍

rdimitrov:
@trishankatdatadog @joshuagl - Thanks again! I've addressed the feedback 👍

trishankatdatadog:
> @trishankatdatadog @joshuagl - Thanks again! I've addressed the feedback 👍

I'll wait for Joshua's approval

rdimitrov:
@trishankatdatadog @joshuagl - The tests caught some side effect of this that it deletes the "targets" folder in case there are no targets at all left in the repository. Do you think that can be a problem? If so, I can update the implementation to skip deleting folders named "targets".

joshuagl:
> @trishankatdatadog @joshuagl - The tests caught some side effect of this that it deletes the "targets" folder in case there are no targets at all left in the repository. Do you think that can be a problem? If so, I can update the implementation to skip deleting folders named "targets".

@trishankatdatadog @ethan-lowman-dd do you anticipate this behaviour causing problems, based on your experience deploying go-tuf?

trishankatdatadog:
> @trishankatdatadog @ethan-lowman-dd do you anticipate this behaviour causing problems, based on your experience deploying go-tuf?

Is it really more dangerous to delete an empty folder than a target file? BTW, what folder is it: is it specific to that target, or is it for all targets?

rdimitrov:
> > @trishankatdatadog @ethan-lowman-dd do you anticipate this behaviour causing problems, based on your experience deploying go-tuf?
> 
> Is it really more dangerous to delete an empty folder than a target file? BTW, what folder is it: is it specific to that target, or is it for all targets?

Previously, go-tuf ignored deleting folders without any targets in it. Meaning if we have a single target "foo/bar" and we delete it, the "foo" folder will remain even though there's nothing left to be served in it.

Now, this change checks if the parent folder of the target we delete will be left empty, and if yes - deletes it too. In the example above where we delete a target "foo/bar" with location "./repository/targets/foo/bar" - deleting "bar" will delete the "foo" folder too (in case there are no more targets in it). So to your question - it is specific to that target.

The side effect that was caught by the tests is about when we have only 1 target "bar" in the repository with location "./repository/targets/bar" - deleting it will also delete its parent directory, which in this case is the "./repository/targets" folder. We need to decide if this is the expected behavior and if not, what is the correct way to handle that?

joshuagl (approved):
This change seems reasonable, it's hard to imagine all scenarios where this might cause breakage. I'm inclined to merge and handle error reports if they are filed.

trishankatdatadog (approved):

trishankatdatadog:
> This change seems reasonable, it's hard to imagine all scenarios where this might cause breakage. I'm inclined to merge and handle error reports if they are filed.

Approve again as you see fit?

PR #238, opened 2022-03-20
Allow commit without adding targets

znewman01:
By making `init` write an empty `targets.json`.

The following didn't work before but now does:

```shell
tuf init

tuf gen-key root
tuf gen-key targets
tuf gen-key snapshot
tuf gen-key timestamp

tuf sign root.json
# you used to need a `tuf add` here
tuf sign targets.json

tuf snapshot
tuf timestamp

tuf commit
```

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -112,11 +112,14 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	root.ConsistentSnapshot = consistentSnapshot
 	// Set root version to 1 for a new root.
 	root.Version = 1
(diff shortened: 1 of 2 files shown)


asraa (approved):
+1 to this change -- this bothered me!

joshuagl (commented):
Thanks @znewman01 for this. The change looks good to me, but I'd love to hear from a maintainer who's more actively using go-tuf cc @trishankatdatadog @ethan-lowman-dd @hosseinsia

trishankatdatadog (approved):
LGTM

PR #240, opened 2022-03-26
use int64 instead of int for metadata version

arbll:
Release Notes: Metadata version now uses int64

**Types of changes**:
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

We have a high write throughput on our TUF repositories and there's a good chance we'll reach 2B+ snapshots versions in the future which would break our clients on 32bit archs. Let's standardize to `int64` to avoid any surprises.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -69,10 +69,10 @@ type Client struct {
 
 	// The following four fields represent the versions of metatdata either
 	// from local storage or from recently downloaded metadata
-	rootVer      int
-	targetsVer   int
-	snapshotVer  int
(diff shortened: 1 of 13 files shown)


arbll:
@asraa @ethan-lowman-dd Anything blocking this ?

arbll:
@asraa @ethan-lowman-dd Gentle bump, let me know if there's anything blocking this. Our backend can wait but we'd like to include it asap in our clients

asraa (approved):

PR #247, opened 2022-03-29
fix: Fix JSON canonicalisation

### LINKED ISSUE #246: Canonicalised JSON format produces invalid JSON
The go-tuf library appears to mis-use JSON canonicalisation when serialising  metadata.

The JSON canonicalisation format (specified here: https://wiki.laptop.org/go/Canonical_JSON) does not escape of control sequences in string values, however the JSON standard (https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) requires that all control characters (defined as U+0000 to U+001F) MUST be escaped in strings.

As a result new line characters in strings are not escaped and this produces invalid JSON and go-tuf will fail to marshal metadata containing these sequences. This is particularly problematic for RSA and ECDSA keys where the specification states they are encoded in PEM format, which will inevitably contain newline characters.

I believe the cause of the error is that go-tuf does not just use canonicalisation to calculate the message digest, and also uses it to serialise the metadata for output.

toby-jn:
Fixes #246

Release Notes:
- Fixes error in JSON canonicalisation that stops RSA/ECDSA keys from being marshalled correctly.

**Types of changes**:

I don't believe this breaks any compatibility, a valid client should still canonicalise the message before hashing.

The JSON canonicalisation format (specified here: https://wiki.laptop.org/go/Canonical_JSON) does not escape of control sequences in string values, however the JSON standard (https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) requires that all control characters (defined as U+0000 to U+001F) MUST be escaped in strings.

As a result new line characters in strings are not escaped and this produces invalid JSON and go-tuf will fail to marshal metadata containing these sequences. This is particularly problematic for RSA and ECDSA keys where the specification states they are encoded in PEM format, which will inevitably contain newline characters.

This change only changes repo to only used canonicalised JSON format for calculating the signature of the message. The serialised message just uses the standard library encoding/json, which correctly handles all control sequences to produce a valid JSON encoding.

The changes to tests are required as they use the `github.com/theupdateframework/go-tuf/pkg/keys.Signer.SignMessage` method directly, this method accepts a `[]byte` instead of a `json.RawMessage`, which implies that it expects a canonicalised message and will not perform canonicalisation.

### DIFF
--- a/client/testdata/go-tuf/consistent-snapshot-false/0/repository/1.root.json
+++ b/client/testdata/go-tuf/consistent-snapshot-false/0/repository/1.root.json
@@ -1,58 +1,53 @@
 {
-	"signatures": [
-		{
-			"keyid": "ce72db3f938914205461a415c9b7b91267a2079df991fd6283aa8461988c1add",
-			"sig": "afab8b34f93c99e200505cc6f0d6a9c4e757a7229f96bd31ab618f8bc4c3abe491b372bfec9d231136de46c1e183df194d372360ee9d32652f8c04d7574e2608"
-		}
-	],
 	"signed": {
 		"_type": "root",
-		"consistent_snapshot": false,
+		"spec_version": "1.0",
+		"version": 1,
 		"expires": "2100-01-01T00:00:00Z",
 		"keys": {
 			"289e5a9e71afd7909326aa4caea92f7557ee0e2283d8c31f0c3401ce67248a45": {
+				"keytype": "ed25519",
+				"scheme": "ed25519",
 				"keyid_hash_algorithms": [
 					"sha256",
 					"sha512"
 				],
-				"keytype": "ed25519",
 				"keyval": {
 					"public": "d482fa4805a50870aa1356ace6b764f7ab47ed4dc38f49b1a189afa25f179e94"
-				},
-				"scheme": "ed25519"
+				}
 			},
 			"808aa256a8172bb0cb961767c6768e55ccf732c99afccc6145752d7a328b7937": {
+				"keytype": "ed25519",
+				"scheme": "ed25519",
 				"keyid_hash_algorithms": [
 					"sha256",
 					"sha512"
 				],
-				"keytype": "ed25519",
 				"keyval": {
 					"public": "62f71f99c788f16bcdc8bb252455e3a690350e4ddea5a6aab1f9a3aaabcf369a"
-				},
-				"scheme": "ed25519"
+				}
 			},
 			"aa3255b4e8e17e566d2bdbea0e5842978f9fa1d2fa9ec76ae76b146164acbfc8": {
+				"keytype": "ed25519",
+				"scheme": "ed25519",
 				"keyid_hash_algorithms": [
 					"sha256",
 					"sha512"
 				],
-				"keytype": "ed25519",
 				"keyval": {
 					"public": "8656ad20568cd3534c405e4d9a84b0c6e6163f7f66434df77416502835b9b160"
-				},
-				"scheme": "ed25519"
+				}
 			},
 			"ce72db3f938914205461a415c9b7b91267a2079df991fd6283aa8461988c1add": {
+				"keytype": "ed25519",
+				"scheme": "ed25519",
 				"keyid_hash_algorithms": [
 					"sha256",
 					"sha512"
 				],
-				"keytype": "ed25519",
 				"keyval": {
 					"public": "bca152214dbfd3e7a1b7a7b3e2cc179fb00520fd7212c8a60ba99f14dfc0e1ca"
-				},
-				"scheme": "ed25519"
+				}
 			}
 		},
 		"roles": {
@@ -81,7 +76,12 @@
 				"threshold": 1
 			}
 		},
-		"spec_version": "1.0",
-		"version": 1
-	}
+		"consistent_snapshot": false
+	},
+	"signatures": [
+		{
+			"keyid": "ce72db3f938914205461a415c9b7b91267a2079df991fd6283aa8461988c1add",
+			"sig": "afab8b34f93c99e200505cc6f0d6a9c4e757a7229f96bd31ab618f8bc4c3abe491b372bfec9d231136de46c1e183df194d372360ee9d32652f8c04d7574e2608"
+		}
+	]
 }
--- a/client/testdata/go-tuf/consistent-snapshot-false/0/repository/root.json
+++ b/client/testdata/go-tuf/consistent-snapshot-false/0/repository/root.json
@@ -1,58 +1,53 @@
 {
-	"signatures": [
-		{
-			"keyid": "ce72db3f938914205461a415c9b7b91267a2079df991fd6283aa8461988c1add",
-			"sig": "afab8b34f93c99e200505cc6f0d6a9c4e757a7229f96bd31ab618f8bc4c3abe491b372bfec9d231136de46c1e183df194d372360ee9d32652f8c04d7574e2608"
-		}
-	],
 	"signed": {
 		"_type": "root",
-		"consistent_snapshot": false,
+		"spec_version": "1.0",
+		"version": 1,
 		"expires": "2100-01-01T00:00:00Z",
 		"keys": {
 			"289e5a9e71afd7909326aa4caea92f7557ee0e2283d8c31f0c3401ce67248a45": {
+				"keytype": "ed25519",
+				"scheme": "ed25519",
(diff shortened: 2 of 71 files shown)


trishankatdatadog:
Hrm, is this also a python-tuf issue? @lukpueh

toby-jn:
The client tests were failing due to the testdata being out of date. After regenerating  they are now passing, however it has raised an interesting issue with the spec.. for the snapshot metadata, it's not entirely clear from the spec whether the hashes metadata applies to the serialised metadata file or whether it should be the hash of the canonicalised version.

As I interpret the spec, it's the serialised form, which is what's implemented here.. however this does seem a strange decision as it means the serialised metadata must be consistently serialised, but it's conceivable that an intermediate mirror or proxy could reformat the JSON. This is not directly related to this change, however as there is a change in the serialised format it is highlighting the issue.

asraa (approved):
+10000!!! I have had this issue for a very long time! THANK YOU!


FYI we may want `repo.Payload(rolename)` method to export the canonicalized bytes that should be signed, and maybe some private function that takes data.Signed -> canonicalized bytes that repo.Payload can call. @znewman01

rdimitrov:
fyi @jku

toby-jn:
> Do we want to canonicalize on the way _out_ (via `tuf payload`) or the way in (`tuf sign-payload`), or both.

I think I'd suggest neither in fact.. keeping all serialised representations un-canonicalised first means they're valid json, and makes it possible to modify the payload before doing sign-payload, without having to mess around with canonicalisation.

Fundamentally the only things that need to ever consider canonicalisation are sign.Sign and verifier.Verify, and all these are really doing is canonicalising the data, passing it to the specific key implementation (that operates on raw []byte) and then attaching the signature in the result.

Following this, the cli sign-payload command is really just the cli analogue to sign.Sign, and so it should accept un-canonicalised data.

> that canonicalizes slightly differently for the signing

The canonicalised format is the specification to ensure this doesn't happen. If a client canonicalises differently, then it's not following the specification and would be unable to verify signatures correctly.

toby-jn:
> This change makes the "idiomatic" usage in repo_test.go really error-prone.

Yep certainly quite messy. I did look at how to do it better, and would perhaps question what these tests are trying to do vs using sign.Sign (they seem to just re-implement a lot of that logic in the test). With the changes in #214 to split things out separately into payload and sign, I think it makes it a bit easier to improve these tests.

znewman01:
Thanks for thinking hard about these questions! I think we're going to wind up with a better interface for users because of it.

> > Do we want to canonicalize on the way _out_ (via `tuf payload`) or the way in (`tuf sign-payload`), or both.
> 
> I think I'd suggest neither in fact.. keeping all serialised representations un-canonicalised first means they're valid json, and makes it possible to modify the payload before doing sign-payload, without having to mess around with canonicalisation.

The way I'm thinking about it, that counts as "on the way in, but not the way out". 

I'd actually lean towards "the way out but *not necessarily* the way in" to enable a use case where non-TUF tools (or a different TUF implementation) can be used to sign a blob. This workflow isn't available at the moment, but you could imagine a `add-signatures` knowing about some standardized signature formats. (I know this description is a little at-odds with #214 as it exists currently).

I guess the question is: "what is a *payload*?" If it's (1) a blob you can sign as-is, then we probably *do* want it to canonicalize. If it's just (2) a representation of the metadata that you need to sign, then I agree with your interpretation.

> Following this, the cli sign-payload command is really just the cli analogue to sign.Sign, and so it should accept un-canonicalised data.

Maybe this is a case where the CLI and API *shouldn't* correspond one-to-one. The way I've been thinking about it:

- `tuf payload`: grab the metadata and output in a canonicalized form
- `tuf sign-payload`: just sign the bytes. In theory, should be able to be replaced by a magic OpenSSL CLI incantation (though we need better `import-key` functionality first).
- `tuf add-signatures`: take a signature/signatures, and attach them to a given metadata file as-is.

This doesn't allow any flows where the metadata is modified before signing; perhaps we really want:

- `tuf export-metadata [--canonical]`
- `tuf sign-blob <blob>`
- `tuf import-metadata`
- `tuf import-signatures`

But it seems like we're really running up against the fact that the go-tuf CLI is more evolved than designed here :)

> The canonicalised format is the specification to ensure this doesn't happen. If a client canonicalises differently, then it's not following the specification and would be unable to verify signatures correctly.

Strictly speaking, the specification [defers format details](https://theupdateframework.github.io/specification/latest/#metaformat) to the [POUF of the implementation](https://theupdateframework.github.io/specification/latest/#pouf-documents). I can imagine wanting to sign payloads in very offline environments that don't have the full go-tuf implementation available (and consequently doesn't know how to canonicalize), but does have a *different* TUF implementation or just has the capacity to sign binary blobs in standard formats. This isn't a client (so there's no need to verify signatures) nor is it a repo (so there's no expectation of a full TUF implementation).

Maybe there's no actual demand for this use case, though.

toby-jn:
> I guess the question is: "what is a payload?" If it's (1) a blob you can sign as-is, then we probably do want it to canonicalise. If it's just (2) a representation of the metadata that you need to sign, then I agree with your interpretation.

Because the signature needs to be in a custom format for the `add-signatures` command to work, I think the external tool would need to be somewhat aware of TUF to make this work well. Doing something with openssl would inevitably require some post-processing step to transform the raw asn.1 signature into a `data.Signature`, so I'm not sure I see completely external signing tools as being viable, and more likely you'd need a `tuf-openssl-signer` that would drive openssl.

In this design, having that tool do the canonicalisation seems optimal, as it means it cannot produce invalid signatures due to user error of providing invalid canonicalised json.

> Strictly speaking, the specification [defers format details](https://theupdateframework.github.io/specification/latest/#metaformat) to the [POUF of the implementation](https://theupdateframework.github.io/specification/latest/#pouf-documents).

I'd misinterpreted this when reading the spec, I'd thought of this more like an http accept header, where a client could request the metadata in different formats, and all signing information would remain intact regardless of the serialization... but yes, it seems this is not the case, which is a bit disappointing and makes the choice of canonical json a

znewman01:
> Because the signature needs to be in a custom format for the `add-signatures` command to work, I think the external tool would need to be somewhat aware of TUF to make this work well. Doing something with openssl would inevitably require some post-processing step to transform the raw asn.1 signature into a `data.Signature`, so I'm not sure I see completely external signing tools as being viable, and more likely you'd need a `tuf-openssl-signer` that would drive openssl.
> 
> In this design, having that tool do the canonicalisation seems optimal, as it means it cannot produce invalid signatures due to user error of providing invalid canonicalised json.

If you're right that  we want a `tuf-openssl-signer` driver, I agree: *it* should do canonicalization.

I'm not 100% convinced: I think there's a small enough number of tools that could plausibly be used to sign (and their signatures are usually in standard formats) that we can special case the `import-signature` command (and possibly the `payload` command) for each one. I just really think it's nice to preserve the option of signing things from environments without TUF installed (though I may be imagining the need for this).

I don't want to block this PR on solving the whole problem, though. So I guess the question is: does moving canonicalization to `sign.Sign` paint us into a corner?

I think the answer is: *only* if we think that we might want to use the go-tuf client to sign payloads from other implementations. I'm okay with giving up on that hope, and then longer-run having a coherent strategy around where canonicalization should happen. WDYT?

znewman01:
I've filed an issue to figure out a coherent canonicalization strategy: https://github.com/theupdateframework/go-tuf/issues/261

For now, this PR fixes a real issue and doesn't seem to commit us to any decisions we're not comfortable with so I'm okay with merging it.

asraa (dismissed):
I agree, let's merge this. I will put out a fix for CI, since it seems broken on go 1.17 and golangci

asraa:
@mnm678 @ethan-lowman-dd could one of you PTAL?

ethan-lowman-dd (dismissed):

asraa:
Could you please rebase?

toby-jn:
rebased onto master

ethan-lowman-dd (approved):

ethan-lowman-dd:
Rebasing again should fix the failed conventional commit check.

toby-jn:
Bit confused by the failing security scan, this is not related to my changes.. should I just ignore it?

asraa:
> Bit confused by the failing security scan, this is not related to my changes.. should I just ignore it?

Yeah, I'll dismiss! Thanks

asraa (approved):

PR #248, opened 2022-03-30
add import-key cli command to import an existing key

toby-jn:
Release Notes:
- Added `import-key` sub-command to the tuf cli to import an existing private key into the local key store and add an entry in the root metadata.

**Types of changes**:

Adds an import-key sub-command to import an existing private key into the local key store and add a corresponding entry in the root metadata.

The reason for this change is that I'm separately working on adding support for storing keys in cloud KMS and Yubikeys. In these cases the key is generated externally and through separate tooling it can then generate a `data.PrivateKey` (in these cases the key type will be some custom type, and the key value will be the path to the key in KMS/Yubikey Serial+Slot), and so this command is necessary to be able to add these keys.

- [x] Docs have been added for the bug fix or new feature

### DIFF
--- a/cmd/tuf/import_key.go
+++ b/cmd/tuf/import_key.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/flynn/go-docopt"
+	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+)
+
+func init() {
+	register("import-key", cmdImportKey, `
+usage: tuf import-key [--expires=<days>] <role> <file>
+
+Import an existing signing key for the given role.
+
+The key will be imported from the file and added to the "keys" directory with
+filename pattern "ROLE-KEYID.json". The root metadata file will also be staged
+with the addition of the key's ID to the role's list of key IDs.
+
+Options:
+  --expires=<days>   Set the root metadata file to expire <days> days from now.
+`)
+}
+
+func cmdImportKey(args *docopt.Args, repo *tuf.Repo) error {
+	role := args.String["<role>"]
+	file := args.String["<file>"]
+	var err error
+
+	var privateKey data.PrivateKey
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return err
+	}
+	if err := json.Unmarshal(data, &privateKey); err != nil {
+		return fmt.Errorf("failed to unmarshal key file: %w", err)
+	}
+
+	signer, err := keys.GetSigner(&privateKey)
+	if err != nil {
+		return fmt.Errorf("failed to get signer for key file: %w", err)
+	}
+
+	if arg := args.String["--expires"]; arg != "" {
+		var expires time.Time
+		expires, err = parseExpires(arg)
+		if err != nil {
+			return err
+		}
+		err = repo.AddPrivateKeyWithExpires(role, signer, expires)
+	} else {
+		err = repo.AddPrivateKey(role, signer)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to add key to repository: %w", err)
+	}
+
+	keyids := signer.PublicData().IDs()
+	for _, id := range keyids {
+		fmt.Println("Imported", role, "key with ID", id)
+	}
+	return nil
+}
--- a/cmd/tuf/import_key.go
+++ b/cmd/tuf/import_key.go
@@ -14,15 +14,16 @@ import (
 
 func init() {
 	register("import-key", cmdImportKey, `
-usage: tuf import-key [--expires=<days>] <role> <file>
+usage: tuf import-key [--expires=<days>] --role=<role> <file>
 
 Import an existing signing key for the given role.
 
-The key will be imported from the file and added to the "keys" directory with
-filename pattern "ROLE-KEYID.json". The root metadata file will also be staged
-with the addition of the key's ID to the role's list of key IDs.
+The private key will be read from the file and added key store.
+The root metadata file will also be staged with the addition of the key's ID to
+the role's list of key IDs.
 
 Options:
+  --role=<role>      The role to import the key for (required).
   --expires=<days>   Set the root metadata file to expire <days> days from now.
 `)
 }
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
@znewman01 do you mind taking a look? 🙂

cmd/tuf/import_key.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/flynn/go-docopt"
+	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+)
+
+func init() {
+	register("import-key", cmdImportKey, `
+usage: tuf import-key [--expires=<days>] <role> <file>
```
znewman01, line 17: Strong preference for `--role`, I don't believe in commands with >1 positional argument after using `ln` for over a decade and still never getting it right on the first try.

cmd/tuf/import_key.go
```diff
@@ -0,0 +1,68 @@
...
+
+Import an existing signing key for the given role.
+
+The key will be imported from the file and added to the "keys" directory with
```
znewman01, line 21: The bit about the "keys" directory is an implementation detail IMO and we don't want to commit to that in the public docs. I think the important points to hit are:

- the key will be imported to this repo's list of keys
- the key will be added to the role's keys

cmd/tuf/import_key.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+)
+
+func init() {
+	register("import-key", cmdImportKey, `
+usage: tuf import-key [--expires=<days>] <role> <file>
+
+Import an existing signing key for the given role.
```
znewman01, line 19: What format(s)?

znewman01: It looks like right now this is "the exact format from the keys DB for another go-tuf repo".  I think that's a lot less useful than e.g. PKCS 8 keys. Also, I think we want to be able to import both public/private keys.

cmd/tuf/import_key.go
```diff
@@ -0,0 +1,68 @@
...
+The key will be imported from the file and added to the "keys" directory with
+filename pattern "ROLE-KEYID.json". The root metadata file will also be staged
+with the addition of the key's ID to the role's list of key IDs.
+
+Options:
+  --expires=<days>   Set the root metadata file to expire <days> days from now.
+`)
+}
+
+func cmdImportKey(args *docopt.Args, repo *tuf.Repo) error {
```
znewman01, line 30: Would prefer to see a `repo.ImportKey` that took a `PublicKey` for the key and `role` for the role so then we could test it a little better (long-run I'd like to have actual CLI tests but short-run we do what we can :) ).

cmd/tuf/import_key.go
znewman01: This is doing both: importing a key into the key database, and adding it to a role. But often we're going to only want to do one or the other.

znewman01 (commented):
We may want to look to the [python-tuf API](https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.metadata.root.html#tuf.api.metadata.Root.add_key) here (CC @jku who has thought pretty hard about the APIs for TUF libraries).


Of course we don't need to match exactly, but I think there's a need to (1) expose key import in the repo API, and (2) have a good interface for users to get `PublicKey`s to pass that API.

There's less precedent for the "keys DB" in python-tuf AFAICT.

jku:
> We may want to look to the [python-tuf API](https://theupdateframework.readthedocs.io/en/latest/api/tuf.api.metadata.root.html#tuf.api.metadata.Root.add_key) here (CC @jku who has thought pretty hard about the APIs for TUF libraries).

Sorry not much help there: we've avoided the CLI design so far and do not have a higher level repository API that would handle details like this.

cmd/tuf/import_key.go
toby-jn: Have switched it over to an option argument for `--role` if that's more consistent with the rest of the tool.

cmd/tuf/import_key.go
toby-jn: So perhaps some more context on why I'm trying to add this would help answer this and the comment below. The specific use case is for adding a private key. The way I envisage this working is having a separate tool like `tuf-yubikey` that would do something like:
```
$ tuf-yubikey generate-key --serial=12345678 --slot=9c --output=yubi-snapshot.json
<prompt for management key>
New key written to yubi-snapshot.json
```
(this would be a bit more interactive to make it easier to derive slot and serial numbers)

This would generate a file like:
```
{
	"keytype": "yubi-rsa",
	"scheme": "rsassa-pss-sha256",
	"keyid_hash_algorithms": [
		"sha256",
		"sha512"
	],
	"keyval": {
		"serial": "12345678",
		"slot": "9c",
		"public": "...."
	}
}
```
But we now need a way to import this into the keystore (adding it to the private key store and then adding it to the root metadata). So you'd now just be able to do `tuf import --role=snapshot yubi-snapshot.json.

As far as I'm aware, unlike public keys the private key format is not part of the TUF specification and are effectively an implementation detail of go-tuf (but perhaps inherited from python-tuf??). Could certainly implement something like pkcs8 fairly easily, but think this is for solving a different problem (this is building up to support external key management, so you won't be able to export the private key).

cmd/tuf/import_key.go
toby-jn: Improved the documentation to remove implementation details.

cmd/tuf/import_key.go
znewman01: > The way I envisage this working is having a separate tool like `tuf-yubikey`

Hmm, my preference is to build as few wrappers as possible for external tools, and instead teach go-tuf about specific formats to import.

> But we now need a way to import this into the keystore (adding it to the private key store and then adding it to the root metadata). So you'd now just be able to do `tuf import --role=snapshot yubi-snapshot.json`.

Perhaps we want a more "plumbing and porcelain" approach here: we provide separate commands for manipulating the keys database and associating keys with roles. Then, you'd be able to write an ugly shell script for generating a key pair and attaching it to a role. After that, you could turn that shell script into its own command so we can do this process in one step.

toby-jn:
Closing this in favour of moving signing with externally managed keys out of go-tuf, and using an export-payload, import-signature workflow instead.

PR #264, opened 2022-04-08
ci: Check PR title instead of commits for conventional format

ethan-lowman-dd:
Release Notes: PR titles must now use [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format

**Types of changes**:

One of the effects of https://github.com/theupdateframework/go-tuf/pull/234 is that we have to format every commit message using the Conventional Commit format, even though we squash merge PRs, and that commit history is lost. Since the PR title (or in some cases a single commit message) becomes the squashed commit message, that is the place we should check for proper formatting. I swapped in a GitHub action that performs this check.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,18 @@
+name: "Changelog checks"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  conventional-commits-lint-check:
+    name: "Lint PR title"
+    runs-on: "ubuntu-latest"
+    steps:
+      - name:
+        uses: amannn/action-semantic-pull-request@0eb081bc9c35210408951834a444794406eff6f8
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,13 +66,3 @@ jobs:
         uses: golangci/golangci-lint-action@5c56cd6c9dc07901af25baab6f2b0d9f3b7c3018
         with:
           version: v1.45.2
-
-  conventional-commits-lint-check:
-    runs-on: "ubuntu-latest"
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
-        with:
-          fetch-depth: 0
-      - name: Commitsar Action
-        uses: aevea/commitsar@159cec82966ca402a09ae3c185524a5256affa22


rdimitrov (approved):
I _think_ the action we are currently using has support too for checking only the PR title via: 
```bash
pull_request:
  conventional: true
```
, but I haven't tried it yet (ref. https://commitsar.aevea.ee/configuration/config-file/)

Anyway, I find the action you suggested to be better and it's also more widely adopted, so I think it make sense for us to switch to it 👍 

A few nits: 

- Not sure if it should be of a concern, but I do like that the one we have doesn't rely on having the GITHUB_TOKEN exposed.
- Should we configure the set of allowed change types ([here](https://github.com/amannn/action-semantic-pull-request#configuration)) to be in line with the changelog groups goreleaser builds ([here](https://github.com/theupdateframework/go-tuf/blob/0e889ad1c804fe0aaae895a8642a664f53847e23/.goreleaser/tuf-client.yml#L24))?

trishankatdatadog (approved):

ethan-lowman-dd:
> Not sure if it should be of a concern, but I do like that the one we have doesn't rely on having the GITHUB_TOKEN exposed.

By default the token has only read permissions, unless there's a different default setting a the repo or org level. I don't have the permissions to verify this.

> Should we configure the set of allowed change types ([here](https://github.com/amannn/action-semantic-pull-request#configuration)) to be in line with the changelog groups goreleaser builds ([here](https://github.com/theupdateframework/go-tuf/blob/0e889ad1c804fe0aaae895a8642a664f53847e23/.goreleaser/tuf-client.yml#L24))?

In that file, it doesn't look like we're restricting the set of allowed change types, just configuring their order in the changelog.

rdimitrov (approved):
> > Should we configure the set of allowed change types ([here](https://github.com/amannn/action-semantic-pull-request#configuration)) to be in line with the changelog groups goreleaser builds ([here](https://github.com/theupdateframework/go-tuf/blob/0e889ad1c804fe0aaae895a8642a664f53847e23/.goreleaser/tuf-client.yml#L24))?
> 
> In that file, it doesn't look like we're restricting the set of allowed change types, just configuring their order in the changelog.

The order is set in the goreleaser config file whereas in the first one I think we can specify which types are allowed - 
```bash
        with:
          # Configure which types are allowed.
          # Default: https://github.com/commitizen/conventional-commit-types
          types: |
            fix
            feat
```
Currently, the changelog will have breaking, fix and feat and everything else will go to others. So my thought was do we want to have separate ones for things like docs, ci, etc and on other side enforce the types we want to allow?

ethan-lowman-dd:
@rdimitrov Personally, I think we should leave it flexible, and then if/when well-defined categories emerge, we can formalize it.

asraa (approved):

hosseinsia (approved):
LGTM

Issue #223, opened 2022-02-14
ecdsa keyval.public format is hex-encoded DER, should be PEM


jku:
It looks like the `ecdsa-sha2-nistp256` keytypes `keyval.public` value is incorrectly the hex encoded raw bytes of the key, instead of PEM. Spec says:

> The ["ecdsa-sha2-nistp256"](https://theupdateframework.github.io/specification/latest/#keytype-ecdsa-sha2-nistp256) format is:
> ```
> {
>   "keytype" : "ecdsa-sha2-nistp256",
>   "scheme" : "ecdsa-sha2-nistp256",
>   "keyval" : {
>     "public" : PUBLIC
>   }
> }
> ```
> **PUBLIC**: PEM format and a string.

I'm looking at e.g the sigstore metadata https://github.com/sigstore/root-signing/blob/main/repository/repository/2.root.json#L36 where the value does not seem to be PEM.

Related:
* Spec issue to clarify key format https://github.com/theupdateframework/specification/issues/201
* python-tuf interop issue: https://github.com/theupdateframework/python-tuf/issues/1859
* maybe this issue is a duplicate of #127 ?

asraa:
I think the big problem was that newlines inside JSON couldn't be handled well with go. I pulled a PEM encoded public key string from python repository testadata and you'll usually get invalid chars trying to unmarshal:
https://go.dev/play/p/WikoJBIBMRo

jku:
I realize the following is not useful to solve the problem that we have now but I'll say this to any future developers thinking of a similar workaround: _Redefining an existing keytype is not the best solution to this problem. The specification supports new keytypes just fine._

I don't have ideas to fix the issue at hand: I'm not very familiar with the go deployment styles but I suppose we are now in a situation where:
* current clients only accept hex ecdsa keys
* even if client code is updated to support both, there is no guarantee that clients are updated
* even if the metadata generating code is fixed, repositories currently serving hex ecdsa keys might not want to switch to PEM keys because of above point

What the project can definitely do is document the issue: _"go-tuf does X in a non-compliant way: it works fine but may prevent using other TUF implementations with the same metadata"_

toby-jn:
Implemented a fix for this in: https://github.com/theupdateframework/go-tuf/pull/270

PR #270, opened 2022-04-16
feat: Implement full support for ECDSA and RSA keys

toby-jn:
Release Notes:

- Implements full support for ECDSA and RSA key types.
- When generating a key with the `gen-key` sub command, the `--type` flag can be specified to select a key type (according the the types in https://theupdateframework.github.io/specification/latest/#keytype). Defaults to `ed25519` keys.

**Types of changes**:

Previously support for ECDSA and RSA keys was only partially implemented:
- Only the verifier/public key type was implemented for ECDSA.
- ECDSA public keys were parsed as the hex encoding of the raw point serialised as ASN.1 . The TUF specification states that they should be encoded as PEM https://theupdateframework.github.io/specification/latest/#keyval-ecdsa-public.
- The RSA signer was not fully implemented, the marshal and unmarshal methods returned not implemented errors.
- The RSA public key mixed up the PEM encoding, it encoded as PKIX, but then set the header to `RSA PUBLIC KEY` which is used for PKCS1 encoding. While the spec doesn't specify which encoding should be used (and just specifies "PEM"), the python reference implementation uses PKIX for both ECDSA and RSA keys.
- The RSA verifier attempted to parse the public key as PKCS1 and PKIX. While the spec does not make this clear, the python implementation only uses PKIX.
- The `gen-key` subcommand does not allow key types to be specified, added a `--type` flag to allow the user to select the key type to be generated.

As well as the above fixes, some additional improvements to the library were made:
- Defined new go types for `KeyType`, `KeyScheme` and `HashAlgorithm`, rather than just using strings, and fixed a few places where the key type and scheme had been mixed up. 
- Added a new `data.PKIXPublicKey` type which abstracts away a lot of the boiler plate for marshalling/unmarshalling public keys.

I opted not to maintain backwards compatibility as there is currently no implementation to generate ECDSA or RSA keys, and the go-tuf implementation wouldn't have worked with keys serialised according the the TUF specification, so there didn't seem any benefit to maintaining this functionality.

- [x] Docs have been added for the bug fix or new feature

### DIFF
--- a/pkg/keys/ecdsa.go
+++ b/pkg/keys/ecdsa.go
@@ -3,69 +3,139 @@ package keys
 import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"crypto/rand"
 	"crypto/sha256"
-	"encoding/asn1"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
-	"math/big"
+	"fmt"
 
 	"github.com/theupdateframework/go-tuf/data"
 )
 
 func init() {
-	VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, NewEcdsaVerifier)
+	VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, newEcdsaVerifier)
+	SignerMap.Store(data.KeyTypeECDSA_SHA2_P256, newEcdsaSigner)
 }
 
-func NewEcdsaVerifier() Verifier {
-	return &p256Verifier{}
+func newEcdsaVerifier() Verifier {
+	return &ecdsaVerifier{}
 }
 
-type ecdsaSignature struct {
-	R, S *big.Int
+func newEcdsaSigner() Signer {
+	return &ecdsaSigner{}
 }
 
-type p256Verifier struct {
-	PublicKey data.HexBytes `json:"public"`
+type ecdsaVerifier struct {
+	PublicKey *data.PKIXPublicKey `json:"public"`
+	ecdsaKey  *ecdsa.PublicKey
 	key       *data.PublicKey
 }
 
-func (p *p256Verifier) Public() string {
-	return p.PublicKey.String()
+func (p *ecdsaVerifier) Public() string {
+	r, _ := x509.MarshalPKIXPublicKey(p.ecdsaKey)
+	return string(r)
 }
 
-func (p *p256Verifier) Verify(msg, sigBytes []byte) error {
-	x, y := elliptic.Unmarshal(elliptic.P256(), p.PublicKey)
-	k := &ecdsa.PublicKey{
-		Curve: elliptic.P256(),
-		X:     x,
-		Y:     y,
-	}
-
-	var sig ecdsaSignature
-	if _, err := asn1.Unmarshal(sigBytes, &sig); err != nil {
-		return err
-	}
-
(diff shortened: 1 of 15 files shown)


toby-jn:
any chance of getting a review of this?

mnm678 (commented):
This looks great. My only concern is the backwards compatibility issue. In the long term I think this is what we need to do (use the spec compliant key types). We could just do this patch as part of a major release. What do others think? @asraa @trishankatdatadog

toby-jn:
Making it backwards compatible with what go-tuf was doing before wouldn't be that difficult.. however the old implementation doesn't follow the spec, and is incompatible with py-tuf and it also doesn't have the necessary functions to actually generate ecdsa/rsa keys, so for someone to have a compatibility issue here they'd have need to have done quite a lot of work outside of go-tuf to make keys in the right format.

asraa (commented):
One problem I see with the ECDSA keys is that if you're using an existing TUF root with ecdsa keys, your root key IDs won't match on update. This is a big problem when you're trying to sign the next root. I think we have to make this backwards compatible, maybe with a flag or an option

trishankatdatadog:
Fixes issue #223

pkg/keys/ecdsa.go
```diff
@@ -3,69 +3,139 @@ package keys
...
 	return nil
 }
 
-func (p *p256Verifier) MarshalPublicKey() *data.PublicKey {
+func (p *ecdsaVerifier) MarshalPublicKey() *data.PublicKey {
 	return p.key
 }
 
-func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
+func (p *ecdsaVerifier) UnmarshalPublicKey(key *data.PublicKey) error {
```
asraa, line 54: Question: To keep backwards compatibility with existing roots (at least for 1-2 releases), could we do something like rename the old `ecdsaVerifier` to `deprecatedEcdsaVerifier`, and attempt to `json.Unmarshal` into that one? 

I'm hoping to be able to still support old verifiers.

For example, if TUF attempts to `UnmarshalPublicKey` on an old ecdsa key, the resulting object would return the deprecatd `Public()` and `Marshal()` as the received data.

pkg/keys/ecdsa_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/theupdateframework/go-tuf/data"
+	. "gopkg.in/check.v1"
+)
+
+type EcdsaSuite struct{}
+
+var _ = Suite(EcdsaSuite{})
+
+func (EcdsaSuite) TestSignVerify(c *C) {
```
asraa, line 12: if you go with backwards compatibility, could you add a test with hard-coded deprecated key material, and make sure that properties like `PublicData()` and `Public()` and `Verify()` still work?

asraa (commented):
Thanks! Hoping to revive this so we can get go-tuf python compliant in the next release.

I just have one proposition on supporting the old key types for a short period of time so users can have a migration plan away from it.

Let me know if you need any help! I'm happy to contribute.

trishankatdatadog:
Damn, I got busy, sorry. Let me look ASAP, maybe tonight.

znewman01:
I'm closing in favor of #357

Thanks for kicking this off, @toby-jn  😄

Issue #292, opened 2022-05-17
Snapshotting doesn't verify the root.json manifest.


asraa:
Currently `Snapshot` will verify signatures on all snapshotted manifests, but does not require verifying the root.json because root is no longer included in the Snapshot. (see https://github.com/theupdateframework/go-tuf/issues/203)

Root.json was removed from snapshot.json here: https://github.com/theupdateframework/taps/blob/master/tap5.md#security-analysis

Should repo managers verify that root.json is valid before snapshotting? As a matter of robustness? Or should we handle this in our own client.

See https://github.com/sigstore/root-signing/issues/238

JustinCappos:
I would think it should verify this.  I'm not sure how you can check that
the snapshot key is current / valid otherwise.

joshuagl:
Agreed that repository tooling should be verifying metadata as the tools proceeds through repository operations.

asraa:
Awesome, the PR opened should do so!

PR #293, opened 2022-05-17
Repo: Check root metadata verification before snapshotting

asraa:

Details in https://github.com/theupdateframework/go-tuf/issues/292
See https://github.com/sigstore/root-signing/issues/238

Because `root.json` is not pinned in `snapshot.json`, it's metadata is not verified before Snapshotting. For robustness, we should verify the root.json is signed correctly, because we're using DB role data from root to verify other manifests that are pinned, e.g. targets.

I'm open to having this just be a warning, because I don't see this mentioned in the spec, but overall repo management lacks in the spec.

Fixes #<Issue>
Release Notes: 

**Types of changes**:

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -1280,6 +1280,11 @@ func (r *Repo) SnapshotWithExpires(expires time.Time) error {
 		return err
 	}
 
+	// Verify root metadata before verifying signatures on role metadata.
+	if err := r.verifySignatures("root.json"); err != nil {
+		return err
+	}
+
 	for _, metaName := range r.snapshotMetadata() {
 		if err := r.verifySignatures(metaName); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


JustinCappos:
I'm open to having the specification make it clear that one should
re-validate the root and the current snapshot key before signing a new
snapshot.

JustinCappos:
To be clear, one need not do this if the root has not changed in the
interim.  But if the root has been updated, this step is important.

joshuagl (approved):

joshuagl:
> I'm open to having the specification make it clear that one should re-validate the root and the current snapshot key before signing a new snapshot.
> […](#)

Indeed, we should probably add sections on generating snapshot, timestamp, and root metadata to the [repository operations](https://theupdateframework.github.io/specification/latest/#repository-operations) section of the spec, it currently only describes [adding and updating targets](https://theupdateframework.github.io/specification/latest/#adding-updating-targets).

znewman01 (approved):

Issue #308, opened 2022-06-08
Contributor licensing


znewman01:
#190 adds a DCO requirement. This was somewhat controversial.

We should:

1. Make a decision on contributor licensing.
2. Document and enforce it.

Summary of available options.

# [Developer Certificate of Origin](https://wiki.linuxfoundation.org/dco) (DCO)

Pros:

- It's what python-tuf does.
- Lighter-weight, legally speaking, than a CLA: many developers are able to sign a DCO but *not* a CLA
- Easy to enforce: there's an off-the-shelf GitHub action.

Cons:

- Can be annoying to type `-s` all the time.
- Makes GitHub web UI changes difficult (you can manually add the `Signed-off-by:` line).

# [Contributor License Agreement](https://en.wikipedia.org/wiki/Contributor_License_Agreement) (CLA)

Pros:

- Set-and-forget: you just have to sign it once.

Cons:

- Some contributors are unable (often due to their employer) or unwilling (due to the legal requirements) to sign a CLA.
- Enforcement is more of a pain. We also have to pick a CLA.

# Do nothing

Pros:

- Status quo; don't have to do anything. Seems to be working so far.
- Other TUF projects seem to be okay with this.
- Nobody has ever convinced me that these things matter.

Cons:

- Legal risks (???)

joshuagl:
DCO is now easier to manage through the GitHub web UI, admins can mark it as required and the UI will fill it in automagically: https://github.blog/changelog/2022-06-08-admins-can-require-sign-off-on-web-based-commits/

I should add that python-tuf requires DCO: https://github.com/theupdateframework/python-tuf/blob/develop/docs/CONTRIBUTING.rst#dco

mnm678:
I think DCO is easier for most people. It looks like the cncf wants projects to have [some kind of contributor licence](https://github.com/cncf/foundation/issues/130)

trishankatdatadog:
> Pros:
>
> * Nobody has ever convinced me that these things matter.
> 
> Cons:
> 
> * Legal risks (???)

Exactly. So let's just go with the simplest thing. DCO is fine by me considering recent changes.

znewman01:
It appears that I don't have the permissions to set this up: https://github.com/dcoapp/app#usage

Someone with "write access" to the `go-tuf` repo needs to configure the repo in Settings > Branches > Branch protection for `master` > Require status checks to pass before merging > DCO (check it!).

[...]

PR #310, opened 2022-06-08
feat: Implement TAP-12 support

### LINKED ISSUE #232: Implement TAP-12 for future-proofing
With the python-tuf implementation, I think that we can move TAP 12 to accepted pretty quickly. Unless there are strong objections to the TAP that will lead to major changes (which I haven't seen), I'd argue for future-proofing go-tuf by implementing TAP 12 now. It may make the implementation temporarily out of sync with the specification, but not in a major way, and not in a way that affects security (as described in more detail in [the TAP](https://github.com/theupdateframework/taps/blob/master/tap12.md)).

_Originally posted by @mnm678 in https://github.com/theupdateframework/go-tuf/issues/228#issuecomment-1059332867_

@mnm678: I think I agree with you, but I want to clarify what that would mean:

- [X] We can just keep *producing* keyIDs using SHA2, no change needed there
- [ ] We need to double-check that we don't add multiple distinct keys with the same keyID (this is done, but we should allow adding multiple *identical* keys to prevent errors)
- [ ] Make sure keys are managed locally to each metadata file, not globally (I think this is the case but need to double-check)
- [ ] During signature verification, need to enforce the uniqueness of each *key*

znewman01:

Fixes #232
Release Notes: Add support for repositories that implement TAP-12 (allowing arbitrary key IDs). go-tuf itself will continue to emit pre-TAP-12-compliant key IDs.

**Types of changes**:

- [X] New feature (non-breaking change which adds functionality)

Allows arbitrary key IDs in repos that we read, per [TAP-12](https://github.com/theupdateframework/taps/blob/master/tap12.md). (We will still only produce key IDs that are the SHA2 of the public key.)

- [X] Tests have been added for the bug fix or new feature
- [ ] (N/A)  Docs have been added for the bug fix or new feature

### DIFF
--- a/verify/db.go
+++ b/verify/db.go
@@ -67,7 +67,7 @@ func (db *DB) AddKey(id string, k *data.PublicKey) error {
 	//
 	// TAP-12: https://github.com/theupdateframework/taps/blob/master/tap12.md
 	if oldVerifier, exists := db.verifiers[id]; exists && oldVerifier.Public() != verifier.Public() {
-		return ErrRepeatID{}
+		return ErrRepeatID{id}
 	}
 
 	db.verifiers[id] = verifier
--- a/verify/db.go
+++ b/verify/db.go
@@ -53,13 +53,23 @@ func NewDBFromDelegations(d *data.Delegations) (*DB, error) {
 }
 
 func (db *DB) AddKey(id string, k *data.PublicKey) error {
-	if !k.ContainsID(id) {
-		return ErrWrongID{}
-	}
 	verifier, err := keys.GetVerifier(k)
 	if err != nil {
 		return ErrInvalidKey
 	}
+
+	// TUF is considering in TAP-12 removing the
+	// requirement that the keyid hash algorithm be derived
+	// from the public key. So to be forwards compatible,
+	// we allow any key ID, rather than checking k.ContainsID(id)
+	//
+	// AddKey should be idempotent, so we allow re-adding the same PublicKey.
+	//
+	// TAP-12: https://github.com/theupdateframework/taps/blob/master/tap12.md
+	if oldVerifier, exists := db.verifiers[id]; exists && oldVerifier.Public() != verifier.Public() {
+		return ErrRepeatID{}
+	}
+
 	db.verifiers[id] = verifier
 	return nil
 }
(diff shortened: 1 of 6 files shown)


verify/db.go
```diff
@@ -53,13 +53,23 @@ func NewDBFromDelegations(d *data.Delegations) (*DB, error) {
...
+	// TUF is considering in TAP-12 removing the
+	// requirement that the keyid hash algorithm be derived
+	// from the public key. So to be forwards compatible,
+	// we allow any key ID, rather than checking k.ContainsID(id)
+	//
+	// AddKey should be idempotent, so we allow re-adding the same PublicKey.
+	//
+	// TAP-12: https://github.com/theupdateframework/taps/blob/master/tap12.md
+	if oldVerifier, exists := db.verifiers[id]; exists && oldVerifier.Public() != verifier.Public() {
+		return ErrRepeatID{}
```
asraa, line 70: Can we add the repeated ID to this error?

verify/db_test.go
```diff
@@ -33,13 +34,6 @@ func TestDelegationsDB(t *testing.T) {
 			}},
 			initErr: ErrInvalidThreshold,
 		},
-		{
-			testName: "invalid keys",
-			delegations: &data.Delegations{Keys: map[string]*data.PublicKey{
-				"a": {Type: data.KeySchemeEd25519},
-			}},
-			initErr: ErrWrongID{},
```
asraa, line 41: re-add this test to test non-sha key IDs and expect no error?

znewman01: Oof, it's a whole lot of work to expect no error because you need to have things signed and so on. I can get to a `ErrNoSignatures` with a lot less work, which *would have been* a `ErrWrongID` before. Is that okay?

verify/db.go
znewman01: Great idea!

verify/db_test.go
asraa: Ah, let's close, we have the test coverage from the other tests you had. thanks!

asraa: Ah but you had added those! already, can you add a quick comment explaining that those tests are testing that we don't fail on non-sha key IDs just in case those tests break later?

asraa (dismissed):

asraa (dismissed):
Thanks!

mnm678 (dismissed):

znewman01:
Hoist by my own petard (#308)!

No changes other than DCO and rebasing, but I need fresh approvals.

mnm678 (approved):

asraa:
@znewman01 @rdimitrov I'm having trouble updating go-tuf in my library, because of a transitive dependency from rekor to v0.3.0 here. This removed `verify.ErrWrongID` -- can we cut a new minor release soon?

PR #325, opened 2022-06-21
fix: verify length and hashes of fetched bytes before parsing

### LINKED ISSUE #233: metadata hashes used in a way that negates security benefits
There seems to be a logic mistake in the way hash checks are done WRT parsing metadata: the point of having hashes for metadata is that parsing completely unverified data can be avoided. The go-tuf client implementation parses the metadata before checking the content hash. This makes using metadata hashes a bit pointless: Once the metadata has been parsed, we have access to signatures...

Details: `downloadMetaFromTimestamp()` downloads bytes, then calls `GenerateTimestampFileMeta()` that parses bytes to metadata. Then it calls `TimestampFileMetaEqual()` to verify byte hashes. I did not look at the same process for snapshot but I assume same is true there.

The logical order for these operations is:
* download bytes 
* check hashes of the bytes
* data is now known to be "vouched for" by delegating metadata: parse/unmarshal bytes into metadata
* do the checks that require parsed metadata (signature, version, etc, ...)

this order is documented in the spec as well, although it does not explain the reason for this order of events and does not specify that data should not be parsed before hash checks.

see also #229

joshuagl:
Fixes #233
Release Notes: 
Check hashes and length of downloaded snapshot and targets metadata before parsing.

**Types of changes**:

Better match the specification by performing checks against length and hashes on the downloaded bytes before parsing the bytes into a FileMeta.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -92,24 +92,23 @@ func BytesMatchLenAndHashes(fetched []byte, length int64, hashes data.Hashes) er
 		return ErrWrongLength{length, flen}
 	}
 
-	if len(hashes) != 0 {
-		for alg, expected := range hashes {
-			var h hash.Hash
-			switch alg {
-			case "sha256":
-				h = sha256.New()
-			case "sha512":
-				h = sha512.New()
-			default:
-				return ErrUnknownHashAlgorithm{alg}
-			}
-			h.Write(fetched)
-			hash := h.Sum(nil)
-			if !hmac.Equal(hash, expected) {
-				return ErrWrongHash{alg, expected, hash}
-			}
+	for alg, expected := range hashes {
+		var h hash.Hash
+		switch alg {
+		case "sha256":
+			h = sha256.New()
+		case "sha512":
+			h = sha512.New()
+		default:
+			return ErrUnknownHashAlgorithm{alg}
+		}
+		h.Write(fetched)
+		hash := h.Sum(nil)
+		if !hmac.Equal(hash, expected) {
+			return ErrWrongHash{alg, expected, hash}
 		}
 	}
+
 	return nil
 }
 
--- a/util/util.go
+++ b/util/util.go
@@ -129,7 +129,7 @@ func hashEqual(actual data.Hashes, expected data.Hashes) error {
 	return nil
 }
 
-func versionEqual(actual int64, expected int64) error {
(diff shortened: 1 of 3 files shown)


util/util.go
```diff
@@ -86,6 +86,33 @@ func FileMetaEqual(actual data.FileMeta, expected data.FileMeta) error {
 	return nil
 }
 
+func BytesMatchLenAndHashes(fetched []byte, length int64, hashes data.Hashes) error {
+	flen := int64(len(fetched))
+	if length != 0 && flen != length {
+		return ErrWrongLength{length, flen}
+	}
+
+	if len(hashes) != 0 {
```
mnm678, line 95: Should this error if there are no hashes?

joshuagl: I don't think so, length and hashes are both optional for [`METAFILES`](https://theupdateframework.github.io/specification/latest/#metafiles).

mnm678 (dismissed):

util/util.go
znewman01, line 95: Why do we need the `!= 0` check? (This is probably my Go inexperience showing)

znewman01 (dismissed):

util/util.go
rdimitrov: That's right, the `len` check can be omitted since it will skip the `for` loop if there are no `hashes` entries 👍

rdimitrov (changes requested):
I'd suggest we add a few simple unit tests for `BytesMatchLenAndHashes` and `VersionEqual` in `util_tests.go`, so we know in case they break at some point 👍 Other than that, it's great 👍 🙏

util/util.go
joshuagl: Well spotted both, thanks! This was my Go inexperience showing :-) Removed the `len()` check in [39b9b0f](https://github.com/theupdateframework/go-tuf/pull/325/commits/39b9b0f1c495d2be5ff14317f0248dce5327e05d)

joshuagl:
I've add some tests in af46f5d and removed the unnecessary `len()` check in 39b9b0f. I've also rebased on the latest master branch and force pushed.

PTAL @mnm678 @znewman01  @rdimitrov 🙏

rdimitrov (approved):
Thanks! 👍 LGTM! 💯

znewman01 (approved):

PR #341, opened 2022-07-13
fix: remove extraneous snapshotted metadata

asraa:

Fixes #<Issue>
Release Notes: 

**Types of changes**:

Before, go-tuf would add `root.json` metadata to the snapshot manifest. Later, it was removed https://github.com/theupdateframework/go-tuf/issues/203. However, snapshotting didn't reset the snapshotmeta, so this left old metadata on the `root.json`: see https://github.com/sigstore/root-signing/pull/320#discussion_r920174081

- [ x] Tests have been added for the bug fix or new feature

### DIFF
--- a/repo.go
+++ b/repo.go
@@ -1277,6 +1277,7 @@ func (r *Repo) SnapshotWithExpires(expires time.Time) error {
 		return err
 	}
 
+	snapshot.Meta = data.SnapshotFiles{}
 	for _, metaName := range r.snapshotMetadata() {
 		if err := r.verifySignatures(metaName); err != nil {
 			return err
--- a/repo_test.go
+++ b/repo_test.go
@@ -2631,3 +2631,48 @@ func (rs *RepoSuite) TestSnapshotWithInvalidRoot(c *C) {
 	c.Assert(r.Timestamp(), IsNil)
 	c.Assert(r.Commit(), IsNil)
 }
+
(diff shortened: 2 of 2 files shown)


asraa:
per @trishankatdatadog on slack, i think this PR may lead to bugs:
```
Check for a rollback attack. The version number of the targets metadata file, and all delegated targets metadata files, if any, in the trusted snapshot metadata file, if any, MUST be less than or equal to its version number in the new snapshot metadata file. **Furthermore, any targets metadata filename that was listed in the trusted snapshot metadata file, if any, MUST continue to be listed in the new snapshot metadata file.** If any of these conditions are not met, discard the new snapshot metadata file, abort the update cycle, and report the failure.
```

even if targets metadata is removed, it must remain in the snapshotted metadata.

here i'm only trying to remove the `root.json` that old versions of tuf clients used to add snapshotted metadata. should i just manually remove this in my impl? or should we special case remove `root.json` from `Snapshot()`?

mnm678:
> here i'm only trying to remove the `root.json` that old versions of tuf clients used to add snapshotted metadata. should i just manually remove this in my impl? or should we special case remove `root.json` from `Snapshot()`?

I'd say just manually remove it. Root updates should be rare, and I worry this functionality could be a footgun if folks don't know how to use it properly. Alternately we can add it for just `root.json` and include a detailed warning

PR #345, opened 2022-07-18
fix: require length and hashes for target metadata

asraa:

Fixes #<Issue>
Release Notes: 
```
* breaking change: `data.FileMeta` no longer contains a `Custom` field
```

Rust's `tough` requires length in the JSON, and go-tuf omits if empty. always marshal a length, even if `0`. Some files need to be signed as empty because there is explicitly no content there.

cc @raulcabello

**Types of changes**:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -669,7 +669,7 @@ func (c *Client) downloadMeta(name string, version int64, m data.FileMeta) ([]by
 }
 
 func (c *Client) downloadMetaFromSnapshot(name string, m data.SnapshotFileMeta) ([]byte, error) {
-	b, err := c.downloadMeta(name, m.Version, m.FileMeta)
+	b, err := c.downloadMeta(name, m.Version, data.FileMeta{m.Length, m.Hashes})
 	if err != nil {
 		return nil, err
 	}
@@ -679,7 +679,7 @@ func (c *Client) downloadMetaFromSnapshot(name string, m data.SnapshotFileMeta) 
 		return nil, ErrDownloadFailed{name, err}
 	}
 
-	meta, err := util.GenerateSnapshotFileMeta(bytes.NewReader(b), m.HashAlgorithms()...)
+	meta, err := util.GenerateSnapshotFileMeta(bytes.NewReader(b), m.Hashes.HashAlgorithms()...)
 	if err != nil {
 		return nil, err
 	}
@@ -693,7 +693,7 @@ func (c *Client) downloadMetaFromSnapshot(name string, m data.SnapshotFileMeta) 
 }
 
 func (c *Client) downloadMetaFromTimestamp(name string, m data.TimestampFileMeta) ([]byte, error) {
-	b, err := c.downloadMeta(name, m.Version, m.FileMeta)
+	b, err := c.downloadMeta(name, m.Version, data.FileMeta{m.Length, m.Hashes})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


asraa:
@trishankatdatadog @mnm678 @rdimitrov could you please review this change? it's breaking consumption of sigstore metadata from `tuftool` because a target file had length 0.

mnm678 (approved):

rdimitrov (commented):
As per the spec, both length and hashes should be optional - [ref](https://theupdateframework.github.io/specification/latest/#metapath-length). I wonder though if having the field present with zero can end up being a problem in case there's some client logic that triggers if the field is present without checking the value too.

rdimitrov:
> > As per the spec, both length and hashes should be optional - [ref](https://theupdateframework.github.io/specification/latest/#metapath-length).
> 
> I don't think that's the correct reference. That's for a metapath length like snapshotted metadata. Target files https://theupdateframework.github.io/specification/latest/#target-files do not mention length being optional.

Yeah, sorry, I was about to edit my comment. I think the problem is that we use the same `FileMeta` object to describe both metafiles and targets. As you said, it's optional for things like snapshot meta (which is what I had in mind initially), but shouldn't be the same for targets.

asraa:
> As you said, it's optional for things like snapshot meta (which is what I had in mind initially), but shouldn't be the same for targets.

I'll update this PR to be a larger fix then to separate these two metas. I'll add a test for a 0 length target and an optionally omitted  lenghth on snapshot meta hopefully.

rdimitrov:
> > As you said, it's optional for things like snapshot meta (which is what I had in mind initially), but shouldn't be the same for targets.
> 
> I'll update this PR to be a larger fix then to separate these two metas. I'll add a test for a 0 length target and an optionally omitted lenghth on snapshot meta hopefully.

Sounds great! 👍

joshuagl (dismissed):
Thank you, this fix looks correct. [`TARGETS`](https://theupdateframework.github.io/specification/latest/#targets-obj-targets) should always have `length` and `hashes`, with an optional `custom` field. [`METAFILES`](https://theupdateframework.github.io/specification/latest/#metafiles) should always have `version`, with optional `length` and `hashes` fields. 

In python-tuf's metadata API we have have a different class to represent each file type: [`TargetFile`](https://theupdateframework.readthedocs.io/en/stable/api/tuf.api.metadata.supporting.html#tuf.api.metadata.TargetFile) and [`MetaFile`](https://theupdateframework.readthedocs.io/en/stable/api/tuf.api.metadata.supporting.html#tuf.api.metadata.MetaFile). The names of the types feel a little less identifiable here, as this is already a break change (right?) could we rename `TargetFileMeta` to `TargetFile`?

asraa:
> The names of the types feel a little less identifiable here, as this is already a break change (right?) could we rename TargetFileMeta to TargetFile?

The only breaking change here is that `FileMeta` (the base FileMeta type) no longer contains `Custom`: added this to the release note. All other names are preserved! I introduced `MetapathFileMeta`, analogous to `TargetFileMeta`, which `SnapshotFileMeta` and `TimestampFileMeta` alias to.

I'd rather keep the wording `XXXXFileMeta` since that's how `(Targets/Snapshot/Timestamp)FileMeta` are all worded in go-tuf. Should I make `MetapathFileMeta` private and only expose `(Targets/Snapshot/Timestamp)FileMeta`? I'd rather not break API and remove `(Snapshot/Timestamp)FileMeta` in favor or something like `MetaFile`

rdimitrov (dismissed):
Thanks, @asraa! 💯 I also appreciate that you made sure it's not introducing any breaking changes 👍

asraa:
thanks! all set with @joshua's comment on `metapathFileMeta` privatized.

mnm678 (approved):

rdimitrov (approved):

joshuagl (approved):

PR #347, opened 2022-07-29
feat: Adds a new raw file metadata storage for clients

kommendorkapten:
This came up during a discussion related to new features for the sigstore TUF client, as there is a desire to make implementations in different languages sharing the target and metadata cache.

Release Notes: Possibility to store metadata files as raw JSON files to allow for interoperability with other TUF implementations.

**Types of changes**:

This pull request introduces a new implementation of the client's `LocalStore` interface. This storage stores the metadata files are regular files in a provided directory. This would improve compatibility to share metadata cache between different language implementations.


client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+
+	"gopkg.in/check.v1"
+)
+
+type RawJSONStoreSuite struct{}
+
+var _ = check.Suite(&RawJSONStoreSuite{})
+
+// Hook up gocheck into the "go test" runnder
```
joshuagl, line 18: Nit: typo
```suggestion
// Hook up gocheck into the "go test" runner
```

joshuagl (approved):
Thanks, this LGTM. 

It would be great to get a review from a maintainer who is more fluent in Go than I am.

asraa (commented):
The change looks good to me as a start, but I wonder a few things and would love feedback:

 Do we need a lock mechanism, esp for SetMeta()? If there are more than one client accessing the store.

kommendorkapten:
Yeah, let me add version that's safe for concurrent use!

kommendorkapten:
I added a new wrapper, 'ConcurrentLocalStore' which can be used to protect any `LocalStore` (e.g. the in memory one and the raw JSON one) for concurrent access. 
This would let the client decide what use depending on their access pattern.

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+
+	meta := map[string]json.RawMessage{}
+	for _, name := range names {
+		p := filepath.FromSlash(filepath.Join(f.baseDir, name.Name()))
+		if ok, err := util.IsMetaFile(name); !ok {
+			continue
+		} else if err != nil {
+			return nil, err
```
ethan-lowman-dd, line 79: This line will never run. Let's check the error before we check `ok`.

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func newImpl(baseDir string, recurse bool) (*FileJSONStore, error) {
+	var f = FileJSONStore{
+		baseDir: baseDir,
+	}
+	var err error
+
+	// Does the directory exist?
+	fi, err := os.Stat(baseDir)
```
ethan-lowman-dd, line 41: Perhaps we should check permissions on the directory and the files inside (in `GetMeta`) if it already exists, similar to how SSH validates keys on disk.

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,121 @@
...
+			baseDir)
+	}
+
+	return &f, nil
+}
+
+// GetMeta returns the currently cached set of metadata files.
+func (f *FileJSONStore) GetMeta() (map[string]json.RawMessage, error) {
+	names, err := os.ReadDir(f.baseDir)
+
```
ethan-lowman-dd, line 68:
```suggestion
```

client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,175 @@
...
+		err := s.SetMeta(k, v)
+		c.Assert(err, check.IsNil)
+	}
+
+	md, err := s.GetMeta()
+	c.Assert(err, check.IsNil)
+	c.Assert(md, check.HasLen, 3)
+	c.Assert(md, check.DeepEquals, expected)
+
+	// Nuke items
```
ethan-lowman-dd, line 91:
```suggestion
	// Delete all items
```

client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,175 @@
...
+func (RawJSONStoreSuite) TestGetNoJSON(c *check.C) {
+	tmp := c.MkDir()
+	p := filepath.Join(tmp, "tuf_raw.db")
+	s, err := NewFileJSONStore(p)
+	c.Assert(s, check.NotNil)
+	c.Assert(err, check.IsNil)
+
+	// Create a file which does not end with '.json'
+	fp := filepath.FromSlash(filepath.Join(p, "meta.xml"))
+	os.WriteFile(fp, []byte{}, 0644)
```
ethan-lowman-dd, line 116: Can we check the returned error?

client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,175 @@
...
+	files := []string{
+		"file.xml",
+		"file",
+		"",
+	}
+	for _, f := range files {
+		err := s.SetMeta(f, []byte{})
+		c.Assert(err, check.Equals, ErrNotJSON)
+	}
+
```
ethan-lowman-dd, line 139:
```suggestion
```

client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,175 @@
...
+	c.Assert(errors.Is(err, os.ErrNotExist), check.Equals, true)
+}
+
+func (RawJSONStoreSuite) TestInterface(c *check.C) {
+	var localStore client.LocalStore
+	var err error
+
+	tmp := c.MkDir()
+	p := filepath.Join(tmp, "tuf_raw.db")
+	localStore, err = NewFileJSONStore(p)
```
ethan-lowman-dd, line 172:
```suggestion
	tmp := c.MkDir()
	p := filepath.Join(tmp, "tuf_raw.db")
	localStore, err := NewFileJSONStore(p)
```

client/filejsonstore/filejsonstore_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"gopkg.in/check.v1"
+
+	"github.com/theupdateframework/go-tuf/client"
+)
+
+type RawJSONStoreSuite struct{}
```
ethan-lowman-dd, line 16: Can we name these tests consistently? (`RawJSONStoreSuite -> FileJSONStoreSuite`)

client/concurrentlocalstore.go
```diff
@@ -0,0 +1,49 @@
+package client
+
+import (
+	"encoding/json"
+	"sync"
+)
+
+// ConcurrentLocalStore provides a LocalStore over an existing LocalStore
+// implementation that is safe for concurrent access.
+type ConcurrentLocalStore struct {
```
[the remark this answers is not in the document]
kommendorkapten: No particular reason except it's flexible. Any `LocalStore` can be protected with this construct, and the existing ones did not perform any concurrency control, so I figured it can be solved for them all with this. And It _can_ simplify development as the locking is managed by a different layer. I'm fine to refactor if it's desired.

There can definitely be some scenarios that does not require concurrency control, but I hardly think there is a measurable performance benefit from omitting it :)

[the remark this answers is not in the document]
ethan-lowman-dd: Yeah, intuitively, I would think that if no concurrency control is actually needed, there would be zero lock contention, and thus minimal overhead for the default safety.

client/filejsonstore/filejsonstore_test.go
kommendorkapten: The reason here for being upfront with the type (`client.LocalStore`) is to make sure the returned type is matching this interface, so I rather not change it.

util/util.go
```diff
@@ -331,3 +331,17 @@ func AtomicallyWriteFile(filename string, data []byte, perm os.FileMode) error {
 
 	return nil
 }
+
+// IsMetaFile tests wheter a DirEntry appears to be a metaddata file or not.
+func IsMetaFile(e os.DirEntry) (bool, error) {
```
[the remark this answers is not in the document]
kommendorkapten: 👍

client/filejsonstore/filejsonstore_test.go
kommendorkapten: 👍

client/filejsonstore/filejsonstore_test.go
kommendorkapten: 👍

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,121 @@
...
+	var f = FileJSONStore{
+		baseDir: baseDir,
+	}
+	var err error
+
+	// Does the directory exist?
+	fi, err := os.Stat(baseDir)
+	if err != nil {
+		pe, ok := err.(*os.PathError)
+		if ok && errors.Is(pe.Err, os.ErrNotExist) && recurse {
```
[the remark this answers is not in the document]
kommendorkapten: Yes, you are correct. I have a faint memory of this being required in an earlier version, but now as you said it's not needed.

client/filejsonstore/filejsonstore.go
kommendorkapten: Good catch!

client/filejsonstore/filejsonstore.go
kommendorkapten: Sure, I'll add checks to make sure only `user` (owner) have write permission to the directory and any files in it.

kommendorkapten:
Windows tests are failing, I will look into this tomorrow.

kommendorkapten:
The failing tests are due to filesystem permission, which on Windows is very different from UNIX-like operating system.
I found this post on the topic: https://medium.com/@MichalPristas/go-and-file-perms-on-windows-3c944d55dd44
The summary is that in windows you can only create a file with either:
* 0666 (Read/writeable by all users)
* 0444 (Readable by all users)

And this seems to be correct according the source code: https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L307
The provided permission is only used to determine if a file is intended to be create as read-only. 
The same is valid for `Mkdir`, the permission mode bits are ignored on Windows: https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L518
And for reference the `Chmod` call: https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L646 (only the read-only flag is manipulated).

It seems that there are some ACL concept in Windows that we can rely on. But I'm not qualified to speak if the normal permission modes set (all-read or all-write) actually are system-wide or if the effective permission applies to the user only. I found this ACL implementation in go: https://github.com/hectane/go-acl
It haven't had any updates in three years, and I would not be comfortable relying on this, as I lack understanding on Windows filesystem access semantics.

As this feature of verifying the filesystem permission bits, is new for this implementation (LevelDB implementation does not care), nor is there any guidance for implementors of `LocalStore` to consider this my suggestion is _to skip file permissions for Windows_.

Please advise on how to proceed.

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,150 @@
...
+// as retrieved from the remote repository.
+type FileJSONStore struct {
+	mtx     sync.RWMutex
+	baseDir string
+}
+
+// NewFileJSONStore returns a new metadata cache, implemented using raw JSON
+// files, stored in a directory provided by the client.
+// If the provided directory does not exist on disk, it will be created.
+// The provided metadata cache is afe for concurrent access
```
patflynn, line 45: (s)afe

kommendorkapten:
Any updates on how to proceed with this (file permission on Windows host)? My comments and proposal are in the comment above: https://github.com/theupdateframework/go-tuf/pull/347#issuecomment-1211615181
cc @trishankatdatadog @ethan-lowman-dd

trishankatdatadog:
> Any updates on how to proceed with this (file permission on Windows host)? My comments and proposal are in the comment above: [#347 (comment)](https://github.com/theupdateframework/go-tuf/pull/347#issuecomment-1211615181) cc @trishankatdatadog @ethan-lowman-dd

I'm going to make a point of looking at this tonight.

trishankatdatadog:
> Please advise on how to proceed.

Sorry for the late reply.

Let's take a look at the python-tuf [reference implementation](https://github.com/theupdateframework/python-tuf/blob/7210487281ee24d14ebfcdc7c1543d2751113c5f/tuf/ngclient/updater.py): there, the permissions of the metadata/targets cache directories is not checked, and metadata/targets files appear to be RW-able by the user but not others (using both `tempfile.NamedTemporaryFile()` and `open()`).

I like your idea for checking the permissions on both the directories and files, but technically, this is not a requirement in the TUF specification. At least the directory permissions can probably be left to the application calling go-tuf, but the files inside them are trickier. If the golang stdlib doesn't give control beyond either RW or RO by all, then there is not much we can do for now.

Is not supporting Windows for now an option? That seems the safest to me.

trishankatdatadog:
@kommendorkapten do you need another review? If so, I recommend pinging those who have reviewed before. Thanks!

kommendorkapten:
@ethan-lowman-dd would you mind taking another look? Addressed your concerns and added filesystem permission check for the metadata files/directories on UNIX-like systems (see discussion above on why Windows is exempted).

trishankatdatadog:
@kommendorkapten if there isn't already a GitHub issue to track the Windows file permissions issue, would you mind making one for it? We should track it. Thanks!

kommendorkapten:
@trishankatdatadog Of course! Created this> https://github.com/theupdateframework/go-tuf/issues/360

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// ErrNotJSON is returned when a metadata operation is attempted
+// against a file that does not seem to be a JSON file
+// (e.g. does not end in .json, case sensitive).
+var ErrNotJSON = errors.New("file is not in JSON format")
+
+// ErrToPermissive is returned when the metadata directory, or a metadata
+// file has too permissive file mode bits set
+var ErrTooPermissive = errors.New("permissions are too permissive")
```
ethan-lowman-dd, line 23: Can we make these not sentinel errors and use `fmt.Sprintf` with context for debugging instead? See `perm.go` for more discussion.

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+var _ client.LocalStore = (*FileJSONStore)(nil)
+
+// NewFileJSONStore returns a new metadata cache, implemented using raw JSON
+// files, stored in a directory provided by the client.
+// If the provided directory does not exist on disk, it will be created.
+// The provided metadata cache is safe for concurrent access.
+func NewFileJSONStore(baseDir string) (*FileJSONStore, error) {
+	var f = FileJSONStore{
```
ethan-lowman-dd, line 50: Can we initialize as a pointer? Now that there is a mutex field, the only valid way to use this type is via a pointer type, since otherwise the mutex is copied.

```suggestion
	f := &FileJSONStore{
```

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,153 @@
...
+		baseDir: baseDir,
+	}
+	var err error
```
ethan-lowman-dd, line 53: This is unnecessary
```suggestion
```

client/filejsonstore/filejsonstore.go
```diff
@@ -0,0 +1,153 @@
...
+	defer f.mtx.RUnlock()
+
+	names, err := os.ReadDir(f.baseDir)
+	if err != nil {
+		return nil, err
+	}
+
+	meta := map[string]json.RawMessage{}
+	for _, name := range names {
+		p := filepath.FromSlash(filepath.Join(f.baseDir, name.Name()))
```
ethan-lowman-dd, line 94: nit: can we put `p` closer to where it's used?

Also, is `filepath.FromSlash` necessary, given that `filepath.Join` already uses an OS-specific filepath separator?

internal/fsutil/perm.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"io/fs"
+	"os"
+)
+
+// EnsurePemissions tests the provided file info to make sure the
+// permission bits matches the provided.
+func EnsurePermission(fi os.FileInfo, perm os.FileMode) error {
+	// Clear all bits which are note related to the permission.
+	mode := fi.Mode() & fs.ModePerm
```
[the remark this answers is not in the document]
kommendorkapten: Ah, nice.

client/filejsonstore/filejsonstore.go
kommendorkapten: Yeah. The result are the same though as the pointer escapes this function, and all methods receives the target by pointer. But making it a pointer in the first place could be more clear.

client/filejsonstore/filejsonstore.go
kommendorkapten: The issue is that we don't know what's inside `f.baseDir`, but that could be fixed in `NewFileJSONStore`. But as I read through the code of `Join` I found that prior to returning the value, `Clean` is called, and `Clean` calls `FromSlash`: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/path/filepath/path.go;l=89

So yes, this `FromSlash` is redundant.

kommendorkapten:
Good feedback @ethan-lowman-dd, all your comments should be addressed now 👍

ethan-lowman-dd (dismissed):

ethan-lowman-dd:
@asraa or @joshuagl Could you please provide a followup review when you get the chance?

trishankatdatadog:
> Good feedback @ethan-lowman-dd, all your comments should be addressed now 👍

Nice work, thanks Ethan

joshuagl (approved):
Lovely work here @kommendorkapten and @ethan-lowman-dd , thank you!

trishankatdatadog:
@kommendorkapten would you please rebase and resolve conflicts? thx!

kommendorkapten:
Was off on PTO, so didn't see you message @trishankatdatadog until now. PR is updated now.

trishankatdatadog:
> Was off on PTO, so didn't see you message @trishankatdatadog until now. PR is updated now.

Thanks! Would you pls fix the linting errors?

kommendorkapten:
> Thanks! Would you pls fix the linting errors?

Sorry, was not aware of them, fixed now.

ethan-lowman-dd (approved):

trishankatdatadog (approved):

PR #353, opened 2022-08-03
chore!: Remove deprecated client Init() function

znewman01:
Has been deprecated for a release or so.

This fixes some tests and removes others; the removals are all
justified:

- TestInitRootExpired: testing of InitLocal checks the behavior on
  expired roots (which is *different*; it allows expired roots)
- TestInitRootTooLarge: covered now in TestUpdate

We also add a "GetMeta" function, used to get at the root json. 

BREAKING CHANGE: remove client.Init() in favor of InitLocal()

Fixes #<Issue>
Release Notes: 

**Types of changes**:


znewman01:
Fixes #208 

Hang on, let me see what's up with the tests....

asraa (changes requested):
thank you very much! so prompt

client/client_test.go
```diff
@@ -174,17 +174,16 @@ func (s *ClientSuite) addRemoteTarget(c *C, name string) {
 	s.syncRemote(c)
 }
 
-func (s *ClientSuite) rootKeys(c *C) []*data.PublicKey {
-	rootKeys, err := s.repo.RootKeys()
+func (s *ClientSuite) rootMeta(c *C) []byte {
+	meta, err := s.repo.RawMeta("root.json")
```
asraa, line 178: do we really need this method? tbh i would enjoy access to the store through the Repo, but otherwise client suite has access to the store.

would like to avoid adding unnecessary methods for test-only. in a separate PR i would wonder if its warranted to access raw metadata from the Repo struct though.

znewman01: Yeah, I don't like test only code in general. However, I think this is actually a useful thing to have: you're quite often going to want to pull the `root.json` bytes out of the repository to stuff in your client repo or whatever. (maybe just a specific `RawRoot()` would be better?)

WDYT?

asraa: >  However, I think this is actually a useful thing to have: you're quite often going to want to pull the root.json bytes out of the repository to stuff in your client repo or whatever.

I completely agree! I think though that it's better to access through the `Store` though. Through the `GetMeta()`.

asraa: If we do go this route, I would prefer `Repo.GetMeta()` that calls the underlying `r.store.GetMeta()`?

znewman01: Oh I think I see what you're saying...1 sec

znewman01: Look now?

asraa: Nice, I'm happy with that change. It simplifies code for me a lot!

I hated having to hold on to a `Store` AND `Repo` object.

anyone else who is reviewing I hope agrees :)

asraa (dismissed):

joshuagl (commented):
https://github.com/theupdateframework/go-tuf/issues/208#issuecomment-1034476029 talks about renaming `InitLocal()` --> `Init()`, is there a reason we shouldn't do that?

This is already a breaking change and the function signatures are different, so it should be relatively easy to figure out how to change `Init()` calls.

@haydentherapper as the author of `InitLocal()`, wdyt?

@hosseinsia @ethan-lowman-dd any thoughts on this?

znewman01:
Good call! I might prefer to do that in a follow-up PR, which will be very very easy to write/review.

trishankatdatadog:
> This is already a breaking change and the function signatures are different, so it should be relatively easy to figure out how to change `Init()` calls.

Agreed

joshuagl:
Easy to review follow-up PR works for me, thanks.

joshuagl (approved):
Thanks for handling this so quickly.

znewman01:
Had to rebase so I need another round of approvals, sorry!

joshuagl (approved):

repo.go
```diff
@@ -1591,3 +1591,8 @@ func (r *Repo) CheckRoleUnexpired(role string, validAt time.Time) error {
 	}
 	return nil
 }
+
+// GetMeta returns the underlying meta file map from the store.
+func (r *Repo) GetMeta() (map[string]json.RawMessage, error) {
```
trishankatdatadog, line 1596: What happened to the `Repo.RootKeys()` function? Is it not there now? Why do we need this?

znewman01: `RootKeys()` is still there, but it returns `data.PublicKey`. When we give the root metadata to the client, it's of type `[]byte`. We *could* format the `data.PublicKey` into a root metadata file ourselves, then serialize. However, we'd have to pull out details like `consistent_snapshot`, thresholds, etc. very manually. 

To my mind, it seems to make sense for the repo to expose an interface to get the *exact* bytes in the `root.json` file, since that's what clients will be downloading, without having to drop down to the store itself.

trishankatdatadog (approved):

znewman01:
FYI, follow-up here to rename "InitLocal" to "Init": https://github.com/theupdateframework/go-tuf/pull/354

PR #357, opened 2022-08-17
feat: Support ecdsa and RSA keys (#270 with backwards compatibility)

asraa:

Fixes https://github.com/theupdateframework/go-tuf/issues/223
Release Notes: 
* breaking change: ECDSA public key verification now use TUF compliant PEM-encoding keys. For backwards compatibility with hex-encoded keys verification, client libraries must import `pkg/deprecated`
* feat: Adds support for ECDSA and RSA signers 
* feat: Repository CLI command allows a `--scheme` flag to specify key scheme.

**Types of changes**:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

* Defined new go types for KeyType, KeyScheme and HashAlgorithm, rather than just using strings, and fixed a few places where the key type and scheme had been mixed up.
* Added a new data.PKIXPublicKey type which abstracts away a lot of the boiler plate for marshalling/unmarshalling public keys.
* Completes support for the RSA signer and fixes public key encoding.

cc @toby-jn: The only changes I made was adding deprecation support and updating your branch!

### DIFF
--- a/cmd/tuf/gen_key.go
+++ b/cmd/tuf/gen_key.go
@@ -6,11 +6,12 @@ import (
 
 	"github.com/flynn/go-docopt"
 	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
 )
 
 func init() {
 	register("gen-key", cmdGenKey, `
-usage: tuf gen-key [--expires=<days>] <role>
+usage: tuf gen-key [--expires=<days>] [--scheme=<scheme>] <role>
 
 Generate a new signing key for the given role.
 
@@ -23,28 +24,40 @@ form of TUF_{{ROLE}}_PASSPHRASE
 
 Options:
   --expires=<days>   Set the root metadata file to expire <days> days from now.
+  --scheme=<scheme>      Set the key scheme to use [default: ed25519].
 `)
 }
 
 func cmdGenKey(args *docopt.Args, repo *tuf.Repo) error {
 	role := args.String["<role>"]
 	var keyids []string
+
+	keyScheme := data.KeySchemeEd25519
+	switch t := args.String["--scheme"]; t {
+	case string(data.KeySchemeEd25519),
+		string(data.KeySchemeECDSA_SHA2_P256),
+		string(data.KeySchemeRSASSA_PSS_SHA256):
+		keyScheme = data.KeyScheme(t)
+	default:
+		fmt.Println("Using default key scheme", keyScheme)
+	}
+
 	var err error
+	var expires time.Time
 	if arg := args.String["--expires"]; arg != "" {
-		var expires time.Time
 		expires, err = parseExpires(arg)
 		if err != nil {
 			return err
 		}
-		keyids, err = repo.GenKeyWithExpires(role, expires)
 	} else {
-		keyids, err = repo.GenKey(role)
+		expires = data.DefaultExpires(role)
 	}
+	keyids, err = repo.GenKeyWithSchemeAndExpires(role, expires, keyScheme)
 	if err != nil {
 		return err
 	}
 	for _, id := range keyids {
-		fmt.Println("Generated", role, "key with ID", id)
+		fmt.Println("Generated", role, keyScheme, "key with ID", id)
 	}
 	return nil
 }
--- a/pkg/deprecated/deprecated_repo_test.go
+++ b/pkg/deprecated/deprecated_repo_test.go
@@ -1,0 +1,82 @@
+package deprecated
+
+import (
+	"crypto"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/json"
+	"testing"
+
+	"github.com/secure-systems-lab/go-securesystemslib/cjson"
+	repo "github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
+	_ "github.com/theupdateframework/go-tuf/pkg/deprecated/set_ecdsa"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+	. "gopkg.in/check.v1"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type RepoSuite struct{}
+
+var _ = Suite(&RepoSuite{})
+
+func genKey(c *C, r *repo.Repo, role string) []string {
+	keyids, err := r.GenKey(role)
+	c.Assert(err, IsNil)
+	c.Assert(len(keyids) > 0, Equals, true)
+	return keyids
+}
+
+// Deprecated ecdsa key support: Support verification against roots that were
+// signed with hex-encoded ecdsa keys.
+func (rs *RepoSuite) TestDeprecatedHexEncodedKeysSucceed(c *C) {
+	files := map[string][]byte{"foo.txt": []byte("foo")}
+	local := repo.MemoryStore(make(map[string]json.RawMessage), files)
+	r, err := repo.NewRepo(local)
+	c.Assert(err, IsNil)
+
+	r.Init(false)
+	// Add a root key with hex-encoded ecdsa format
+	signer, err := keys.GenerateEcdsaKey()
+	c.Assert(err, IsNil)
+	type deprecatedP256Verifier struct {
+		PublicKey data.HexBytes `json:"public"`
+	}
+	pub := signer.PublicKey
+	keyValBytes, err := json.Marshal(&deprecatedP256Verifier{PublicKey: elliptic.Marshal(pub.Curve, pub.X, pub.Y)})
+	c.Assert(err, IsNil)
+	publicData := &data.PublicKey{
+		Type:       data.KeyTypeECDSA_SHA2_P256,
+		Scheme:     data.KeySchemeECDSA_SHA2_P256,
+		Algorithms: data.HashAlgorithms,
+		Value:      keyValBytes,
+	}
+	err = r.AddVerificationKey("root", publicData)
+	c.Assert(err, IsNil)
+	// Add other keys as normal
+	genKey(c, r, "targets")
+	genKey(c, r, "snapshot")
+	genKey(c, r, "timestamp")
+	c.Assert(r.AddTarget("foo.txt", nil), IsNil)
+
+	// Sign the root role manually
+	rootMeta, err := r.SignedMeta("root.json")
+	c.Assert(err, IsNil)
+	rootCanonical, err := cjson.EncodeCanonical(rootMeta.Signed)
+	c.Assert(err, IsNil)
+	hash := sha256.Sum256(rootCanonical)
+	rootSig, err := signer.PrivateKey.Sign(rand.Reader, hash[:], crypto.SHA256)
+	c.Assert(err, IsNil)
+	for _, id := range publicData.IDs() {
+		c.Assert(r.AddOrUpdateSignature("root.json", data.Signature{
+			KeyID:     id,
+			Signature: rootSig}), IsNil)
+	}
+
+	// Committing should succeed because the deprecated key pkg is added.
+	c.Assert(r.Snapshot(), IsNil)
+	c.Assert(r.Timestamp(), IsNil)
+	c.Assert(r.Commit(), IsNil)
+}
--- a/pkg/keys/ecdsa.go
+++ b/pkg/keys/ecdsa.go
@@ -4,64 +4,62 @@ import (
 	"bytes"
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"crypto/rand"
 	"crypto/sha256"
-	"encoding/asn1"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
-	"math/big"
 
 	"github.com/theupdateframework/go-tuf/data"
 )
 
 func init() {
 	VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, NewEcdsaVerifier)
+	SignerMap.Store(data.KeyTypeECDSA_SHA2_P256, newEcdsaSigner)
 }
 
 func NewEcdsaVerifier() Verifier {
-	return &p256Verifier{}
+	return &EcdsaVerifier{}
 }
 
-type ecdsaSignature struct {
-	R, S *big.Int
+func newEcdsaSigner() Signer {
+	return &ecdsaSigner{}
 }
 
-type p256Verifier struct {
-	PublicKey data.HexBytes `json:"public"`
+type EcdsaVerifier struct {
+	PublicKey *PKIXPublicKey `json:"public"`
+	ecdsaKey  *ecdsa.PublicKey
 	key       *data.PublicKey
 }
 
-func (p *p256Verifier) Public() string {
-	return p.PublicKey.String()
+func (p *EcdsaVerifier) Public() string {
+	// This is already verified to succeed when unmarshalling a public key.
+	r, err := x509.MarshalPKIXPublicKey(p.ecdsaKey)
+	if err != nil {
+		// TODO: Gracefully handle these errors.
+		// See https://github.com/theupdateframework/go-tuf/issues/363
+		panic(err)
+	}
+	return string(r)
 }
 
-func (p *p256Verifier) Verify(msg, sigBytes []byte) error {
-	x, y := elliptic.Unmarshal(elliptic.P256(), p.PublicKey)
-	k := &ecdsa.PublicKey{
-		Curve: elliptic.P256(),
-		X:     x,
-		Y:     y,
-	}
-
-	var sig ecdsaSignature
-	if _, err := asn1.Unmarshal(sigBytes, &sig); err != nil {
-		return err
-	}
-
+func (p *EcdsaVerifier) Verify(msg, sigBytes []byte) error {
 	hash := sha256.Sum256(msg)
 
-	if !ecdsa.Verify(k, hash[:], sig.R, sig.S) {
+	if !ecdsa.VerifyASN1(p.ecdsaKey, hash[:], sigBytes) {
 		return errors.New("tuf: ecdsa signature verification failed")
 	}
 	return nil
 }
 
-func (p *p256Verifier) MarshalPublicKey() *data.PublicKey {
+func (p *EcdsaVerifier) MarshalPublicKey() *data.PublicKey {
 	return p.key
 }
 
-func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
+func (p *EcdsaVerifier) UnmarshalPublicKey(key *data.PublicKey) error {
 	// Prepare decoder limited to 512Kb
 	dec := json.NewDecoder(io.LimitReader(bytes.NewReader(key.Value), MaxJSONKeySize))
 
@@ -73,14 +71,98 @@ func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
 		return err
 	}
 
-	curve := elliptic.P256()
-
-	// Parse as uncompressed marshalled point.
-	x, _ := elliptic.Unmarshal(curve, p.PublicKey)
-	if x == nil {
-		return errors.New("tuf: invalid ecdsa public key point")
+	ecdsaKey, ok := p.PublicKey.PublicKey.(*ecdsa.PublicKey)
+	if !ok {
+		return fmt.Errorf("invalid public key")
 	}
 
+	if _, err := x509.MarshalPKIXPublicKey(ecdsaKey); err != nil {
+		return fmt.Errorf("marshalling to PKIX key: invalid public key")
+	}
+
+	p.ecdsaKey = ecdsaKey
 	p.key = key
 	return nil
+}
+
+type ecdsaSigner struct {
+	*ecdsa.PrivateKey
+}
+
+type ecdsaPrivateKeyValue struct {
+	Private string         `json:"private"`
+	Public  *PKIXPublicKey `json:"public"`
+}
+
+func (s *ecdsaSigner) PublicData() *data.PublicKey {
+	// This uses a trusted public key JSON format with a trusted Public value.
+	keyValBytes, _ := json.Marshal(EcdsaVerifier{PublicKey: &PKIXPublicKey{PublicKey: s.Public()}})
+	return &data.PublicKey{
+		Type:       data.KeyTypeECDSA_SHA2_P256,
+		Scheme:     data.KeySchemeECDSA_SHA2_P256,
+		Algorithms: data.HashAlgorithms,
+		Value:      keyValBytes,
+	}
+}
+
+func (s *ecdsaSigner) SignMessage(message []byte) ([]byte, error) {
+	hash := sha256.Sum256(message)
+	return ecdsa.SignASN1(rand.Reader, s.PrivateKey, hash[:])
+}
+
+func (s *ecdsaSigner) MarshalPrivateKey() (*data.PrivateKey, error) {
+	priv, err := x509.MarshalECPrivateKey(s.PrivateKey)
+	if err != nil {
+		return nil, err
+	}
+	pemKey := pem.EncodeToMemory(&pem.Block{Type: "EC PRIVATE KEY", Bytes: priv})
+	val, err := json.Marshal(ecdsaPrivateKeyValue{
+		Private: string(pemKey),
+		Public:  &PKIXPublicKey{PublicKey: s.Public()},
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &data.PrivateKey{
+		Type:       data.KeyTypeECDSA_SHA2_P256,
+		Scheme:     data.KeySchemeECDSA_SHA2_P256,
+		Algorithms: data.HashAlgorithms,
+		Value:      val,
+	}, nil
+}
+
+func (s *ecdsaSigner) UnmarshalPrivateKey(key *data.PrivateKey) error {
+	val := ecdsaPrivateKeyValue{}
+	if err := json.Unmarshal(key.Value, &val); err != nil {
+		return err
+	}
+	block, _ := pem.Decode([]byte(val.Private))
+	if block == nil {
+		return errors.New("invalid PEM value")
+	}
+	if block.Type != "EC PRIVATE KEY" {
+		return fmt.Errorf("invalid block type: %s", block.Type)
+	}
+	k, err := x509.ParseECPrivateKey(block.Bytes)
+	if err != nil {
+		return err
+	}
+	if k.Curve != elliptic.P256() {
+		return errors.New("unsupported ecdsa curve")
+	}
+	if _, err := json.Marshal(EcdsaVerifier{
+		PublicKey: &PKIXPublicKey{PublicKey: k.Public()}}); err != nil {
+		return fmt.Errorf("invalid public key: %s", err)
+	}
+
+	s.PrivateKey = k
+	return nil
+}
+
+func GenerateEcdsaKey() (*ecdsaSigner, error) {
+	privkey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, err
+	}
+	return &ecdsaSigner{privkey}, nil
 }
--- a/pkg/keys/ecdsa.go
+++ b/pkg/keys/ecdsa.go
@@ -17,7 +17,10 @@ import (
 )
 
 func init() {
-	VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, NewEcdsaVerifier)
+	// Note: we use LoadOrStore here to prevent accidentally overriding the
+	// an explicit deprecated ECDSA verifier.
+	// TODO: When deprecated ECDSA is removed, this can switch back to Store.
+	VerifierMap.LoadOrStore(data.KeyTypeECDSA_SHA2_P256, NewEcdsaVerifier)
 	SignerMap.Store(data.KeyTypeECDSA_SHA2_P256, newEcdsaSigner)
 }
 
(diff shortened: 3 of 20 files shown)


trishankatdatadog:
So does this supersede #270 ?

asraa:
> So does this supersede https://github.com/theupdateframework/go-tuf/pull/270 ?

Yes! It's all the changes there with backwards compatibility for just ECDSA verification: i.e. if a current root (sigstore) is using hex-encoded keys, then they will still be able to verify (no hex-encoded signers supported). Adds testing for making sure deprecation still verifies at the key and repo testing level.

asraa:
I can also split out just the ECDSA changes. Hoping to get that in ASAP since the next sigstore TUF root to rotate those out relies on it.

toby-jn:
Great thanks, LGTM. I did start trying to rebase my changes onto master over the weekend to make these updates, but there were some conflicts that I needed to go read up on and then real life got in the way, so thanks for taking this over.

trishankatdatadog:
Thanks very much for all your hard work, Toby! Sorry I didn't get around to reviewing it in time: swamped by work, too...

pkg/keys/ecdsa.go
```diff
@@ -82,5 +155,82 @@ func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
...
+	return &data.PublicKey{
+		Type:       data.KeyTypeECDSA_SHA2_P256,
+		Scheme:     data.KeySchemeECDSA_SHA2_P256,
+		Algorithms: data.HashAlgorithms,
+		Value:      keyValBytes,
+	}
+}
+
+func (s *ecdsaSigner) SignMessage(message []byte) ([]byte, error) {
+	hash := sha256.Sum256(message)
```
trishankatdatadog, line 182: It's not likely to be a large metadata file, but should we watch out for it?

pkg/keys/ecdsa.go
```diff
@@ -82,5 +155,82 @@ func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
...
+	}
+	if block.Type != "EC PRIVATE KEY" {
+		return fmt.Errorf("invalid block type: %s", block.Type)
+	}
+	k, err := x509.ParseECPrivateKey(block.Bytes)
+	if err != nil {
+		return err
+	}
+	if k.Curve != elliptic.P256() {
+		return errors.New("invalid ecdsa curve")
```
trishankatdatadog, line 224:
```suggestion
		return errors.New("unsupported ecdsa curve")
```

pkg/keys/ecdsa_test.go
```diff
@@ -17,28 +17,74 @@ import (
...
+	c.Assert(pubKey.Verify(msg, sig), IsNil)
+}
 
-	// Marshall as non compressed point
-	pub := elliptic.Marshal(elliptic.P256(), priv.X, priv.Y)
-
-	publicKey, err := json.Marshal(map[string]string{
-		"public": hex.EncodeToString(pub),
-	})
+func (ECDSASuite) TestDeprecatedFormatVerify(c *C) {
```
trishankatdatadog, line 34:
```suggestion
func (ECDSASuite) TestSignVerifyDeprecatedFormat(c *C) {
```

pkg/keys/rsa.go
```diff
@@ -57,53 +52,26 @@ func (p *rsaVerifier) UnmarshalPublicKey(key *data.PublicKey) error {
...
+	if !ok {
+		return fmt.Errorf("invalid public key")
 	}
+	p.rsaKey = rsaKey
 	p.key = key
 	return nil
 }
 
-// parseKey tries to parse a PEM []byte slice by attempting PKCS1 and PKIX in order.
-func parseKey(data string) (*rsa.PublicKey, error) {
```
trishankatdatadog, line 70: Good riddance!

pkg/keys/rsa_test.go
```diff
@@ -28,3 +29,16 @@ func (RsaSuite) TestMarshalUnmarshal(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(pubKey.MarshalPublicKey(), DeepEquals, publicData)
 }
+
```
trishankatdatadog, line 32: Missing a few more tests, no, like signing and verifying? Would be nice to see the tests match the ecdsa ones.

repo.go
```diff
@@ -309,19 +309,29 @@ func (r *Repo) GenKey(role string) ([]string, error) {
...
-		return []string{}, err
+func (r *Repo) GenKeyWithTypeAndExpires(role string, expires time.Time, keyType data.KeyType) ([]string, error) {
+	var signer keys.Signer
+	var err error
+	switch keyType {
+	case data.KeyTypeEd25519:
+		signer, err = keys.GenerateEd25519Key()
+	case data.KeyTypeECDSA_SHA2_P256:
+		signer, err = keys.GenerateEcdsaKey()
+	case data.KeyTypeRSASSA_PSS_SHA256:
```
trishankatdatadog, line 323: This is where I think things get confusing. I understand that key types and schemes are distinct, but if we are only going to support "rsa-pss-sha256", then we should use schemes instead of types here to make things sparkling clear.

trishankatdatadog (commented):
Good stuff, thanks @asraa (and @toby-jn)!

Sorry for my late review, but a few comments...

pkg/keys/ecdsa.go
```diff
@@ -4,64 +4,137 @@ import (
...
+	key *data.PublicKey
+	// This will switch based on whether this is a PEM-encoded key
+	// or a deprecated hex-encoded key.
+	Verifier
+}
+
+func (p *ecdsaVerifierWithDeprecatedSupport) UnmarshalPublicKey(key *data.PublicKey) error {
+	p.key = key
+	pemVerifier := &ecdsaVerifier{}
+	if err := pemVerifier.UnmarshalPublicKey(key); err != nil {
```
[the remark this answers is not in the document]
asraa: Guarding behind a flag would be great! I could imagine a build tag too that only adds the deprecated if the tag is enabled. But this API can be called as a library, and not as a compiled package. So maybe some options are:
* An environment variable, go-tuf specific: `GO_TUF_ENABLE_DEPRECATED_ECDSA`
* Adding an `init` function for deprecated keys into a separate package, and not including that into the default binary. Then only libraries would be able to use it on import.

cmd/tuf/gen_key.go
```diff
@@ -23,23 +25,35 @@ form of TUF_{{ROLE}}_PASSPHRASE
...
+		keyType = data.KeyType(t)
+	default:
+		return errors.New("invalid key type")
+	}
+
+	var expires time.Time
 	if arg := args.String["--expires"]; arg != "" {
-		var expires time.Time
-		expires, err = parseExpires(arg)
+		exp, err := parseExpires(arg)
```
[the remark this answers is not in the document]
asraa: Fixed, good catch! likely because err was not defined then.

repo.go
asraa: I agree, changing to scheme.

pkg/keys/ecdsa.go
asraa: Since all signers implement this, I wonder if it should be implemented at a higher level, before the json becomes canonicalized bytes. I can file a follow-up issue. I'm hand-wavy on whether it matters: signing is never done by the client so it's not as much of a threat.

pkg/keys/ecdsa.go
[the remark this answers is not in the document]
trishankatdatadog: I like the second one, if it's not too much work.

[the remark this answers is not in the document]
asraa: Me too! All set. I have to add a separate package `pkg/deprecated` so that the init function only runs when someone pull this in (as opposed to the keys package). Added a repo level test in there so that you can verify repository metadata with hex-encoded keys. Modified the `repo_test.go` so that it *fails* with the hex-encoded keys (same test, just failure)

asraa:
@theupdateframework/go-tuf-maintainers could I please get one more review on this? really need this

trishankatdatadog:
> @theupdateframework/go-tuf-maintainers could I please get one more review on this? really need this

Yes, sorry, was on PTO, will look tonight

asraa:
No worries at all -- didn't realize that! Feel free to do it tomorrow, I hope you had some rest :)

pkg/deprecated/deprecated_repo_test.go
```diff
@@ -0,0 +1,81 @@
...
+	hash := sha256.Sum256(rootCanonical)
+	rootSig, err := signer.PrivateKey.Sign(rand.Reader, hash[:], crypto.SHA256)
+	c.Assert(err, IsNil)
+	for _, id := range publicData.IDs() {
+		c.Assert(r.AddOrUpdateSignature("root.json", data.Signature{
+			KeyID:     id,
+			Signature: rootSig}), IsNil)
+	}
+
+	// Committing should fail initially.
```
trishankatdatadog, line 77: I'm confused. These don't look like they're failing?

pkg/keys/ecdsa.go
```diff
@@ -4,64 +4,56 @@ import (
...
+type EcdsaVerifier struct {
+	PublicKey *PKIXPublicKey `json:"public"`
+	ecdsaKey  *ecdsa.PublicKey
 	key       *data.PublicKey
 }
 
-func (p *p256Verifier) Public() string {
-	return p.PublicKey.String()
+func (p *EcdsaVerifier) Public() string {
+	r, _ := x509.MarshalPKIXPublicKey(p.ecdsaKey)
```
trishankatdatadog, line 39: Why are we not checking for an error here?

pkg/keys/ecdsa.go
```diff
@@ -73,14 +65,87 @@ func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
...
+	Private string         `json:"private"`
+	Public  *PKIXPublicKey `json:"public"`
+}
 
-	// Parse as uncompressed marshalled point.
-	x, _ := elliptic.Unmarshal(curve, p.PublicKey)
-	if x == nil {
-		return errors.New("tuf: invalid ecdsa public key point")
+func (s *ecdsaSigner) PublicData() *data.PublicKey {
+	keyValBytes, _ := json.Marshal(EcdsaVerifier{PublicKey: &PKIXPublicKey{PublicKey: s.Public()}})
```
trishankatdatadog, line 87: No checking for errors?

pkg/keys/rsa.go
```diff
@@ -9,37 +10,33 @@ import (
...
 
 func (p *rsaVerifier) Public() string {
-	// Unique public key identifier, use a uniform encodng
-	r, err := x509.MarshalPKIXPublicKey(p.rsaKey)
-	if err != nil {
-		// This shouldn't happen with a valid rsa key, but fallback on the
-		// JSON public key string
-		return string(p.PublicKey)
-	}
+	r, _ := x509.MarshalPKIXPublicKey(p.rsaKey)
```
trishankatdatadog, line 39: No error checking?

pkg/keys/rsa.go
```diff
@@ -54,56 +51,37 @@ func (p *rsaVerifier) MarshalPublicKey() *data.PublicKey {
...
 
 func (s *rsaSigner) PublicData() *data.PublicKey {
-	pub, _ := x509.MarshalPKIXPublicKey(s.Public().(*rsa.PublicKey))
-	pubBytes := pem.EncodeToMemory(&pem.Block{
-		Type:  "RSA PUBLIC KEY",
-		Bytes: pub,
-	})
-
-	keyValBytes, _ := json.Marshal(rsaPublic{PublicKey: string(pubBytes)})
+	keyValBytes, _ := json.Marshal(rsaVerifier{PublicKey: &PKIXPublicKey{PublicKey: s.Public()}})
```
trishankatdatadog, line 84: No error checking?

pkg/keys/ecdsa.go
```diff
@@ -73,14 +65,87 @@ func (p *p256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
...
+	return &data.PublicKey{
```
trishankatdatadog, line 88: (BTW, side note: I'm beginning to think the TUF spec should also specify the encoding expect for public keys ahead of time, but that's a different story for another time.) @mnm678

pkg/keys/rsa.go
```diff
@@ -54,56 +51,37 @@ func (p *rsaVerifier) MarshalPublicKey() *data.PublicKey {
...
-}
-
 type rsaSigner struct {
 	*rsa.PrivateKey
 }
 
-type rsaPublic struct {
-	// PEM encoded public key.
-	PublicKey string `json:"public"`
+type rsaPrivateKeyValue struct {
```
trishankatdatadog, line 78: (Side note: no need to fix this now, and sticking to auto salt lengths for signing/verifying makes sense, but at some point, may make sense to encode this as part of the public key. See my related [Vault PR](https://github.com/hashicorp/vault/pull/16549).)

trishankatdatadog (commented):
Looks great to me, thanks very much! Just a few comments...

cmd/tuf/gen_key.go
```diff
@@ -25,35 +24,33 @@ form of TUF_{{ROLE}}_PASSPHRASE
...
-	var keyType data.KeyType
-	switch t := args.String["--type"]; t {
-	case string(data.KeyTypeEd25519),
-		string(data.KeyTypeECDSA_SHA2_P256),
-		string(data.KeyTypeRSASSA_PSS_SHA256):
-		keyType = data.KeyType(t)
-	default:
-		return errors.New("invalid key type")
+	keyScheme := data.KeySchemeEd25519
+	switch t := args.String["--scheme"]; t {
```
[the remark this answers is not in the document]
asraa: Ohhh, I was operating under if `--scheme` isn't explicitly provided, then default ot Ed25519 (like existing behavior). I don't think people should always have to specify `--scheme`, same as how they don't have to specify `--expires`.

A compromise could be logging `Generated root ed25519 key with id xyz` just to inform users? Added.

pkg/deprecated/deprecated_repo_test.go
asraa: Nice catch! this was copy-pasted. This succeeds, since the deprecated keys are added as imports.

pkg/keys/ecdsa.go
asraa: Hmm this was done because this assumes by the time a verifier is created (either done through unmarshalling or explcit construction) that it is well-formed. It looks like in `UnmarhsalPublicKey` that is not verified. Adding that, and adding a comment on the error handling here.

pkg/keys/ecdsa.go
asraa: Same as above, but in fact the entire key interface should hanlde errors in `Public()` and `PublicData()` where they currently return no errors.

pkg/keys/rsa.go
asraa: Going to file an issue about this. It would be nice to clean this up along with the change about multiple Key IDs.

pkg/keys/rsa.go
asraa: For now I've added checking that this will succeed when creating an rsa signer through unmarshalling, but filed https://github.com/theupdateframework/go-tuf/issues/363

pkg/keys/ecdsa.go
asraa: Agreed! There currently is consistency of repo key storage across impls.

cmd/tuf/gen_key.go
```diff
@@ -6,11 +6,12 @@ import (
 
 	"github.com/flynn/go-docopt"
 	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
 )
 
 func init() {
 	register("gen-key", cmdGenKey, `
-usage: tuf gen-key [--expires=<days>] <role>
+usage: tuf gen-key [--expires=<days>] [--type=<type>] <role>
```
trishankatdatadog, line 14:
```suggestion
usage: tuf gen-key [--expires=<days>] [--scheme=<scheme>] <role>
```

trishankatdatadog (commented):
Thanks very much, Asra!

So just to double-check: what's the worst that can happen re:#363? We are not able to accept malformed/unusable public keys after unmarshaling, correct?

asraa:
> So just to double-check: what's the worst that can happen re:https://github.com/theupdateframework/go-tuf/issues/363? We are not able to accept malformed/unusable public keys after unmarshaling, correct?

Good question, and I'll attach my reply to the issue as well. Currently, we use those functions after we create a `verifier` in functionality like `keys.GetVerifier`
or a `signer` through `keys.GetSigner`

https://github.com/theupdateframework/go-tuf/blob/355e39cb2df220fc3961396a6d0e30bcf2c9ac12/pkg/keys/keys.go#L57-L69

So as long as they're verified during `UnmarshalPublicKey` or `UnmarshalPrivateKey` (which now they are), then we will always have a "well-formed" enough public key to `Marshal` or get the `PublicData()` from. 

In case someone accidentally corrupts those keys in the key store or TUF metadata, then we would return errors like `tuf: error unmarshalling key:` on loading them in as `Verifier` or `Signer`. 

So worst case, yes, the key becomes un-usable on load.

That being said, it would always be good practice to re-validate the keys upon marshalling even if they were loaded in a good state, just in case. (for the issue)

trishankatdatadog (dismissed):
Well-done, @asraa (and @toby-jn), thank you very much! Let's go 🚀

znewman01:
ACK can look today (sorry, thought review was blocked for some reason)

repo.go
```diff
@@ -309,19 +309,33 @@ func (r *Repo) GenKey(role string) ([]string, error) {
 }
 
 func (r *Repo) GenKeyWithExpires(keyRole string, expires time.Time) (keyids []string, err error) {
-	// Not compatible with delegated targets roles, since delegated targets keys
-	// are associated with a delegation (edge), not a role (node).
+	return r.GenKeyWithSchemeAndExpires(keyRole, expires, data.KeySchemeEd25519)
+}
 
-	signer, err := keys.GenerateEd25519Key()
+func (r *Repo) GenKeyWithSchemeAndExpires(role string, expires time.Time, keyScheme data.KeyScheme) ([]string, error) {
```
znewman01, line 315: I'm sensing combinatorial explosion...what if I want a custom scheme but the default expiration time?

What do Go developers do in this case? (TBH I would believe you if you told me "write 2^(# options) functions 😛 )

pkg/keys/ecdsa.go
znewman01: If that's the case, I'd like to see an assertion that the error is nil.

znewman01 (dismissed):
Do you need to check "breaking change" for this PR?

I left some little comments, nice to resolve but not blocking.

My final concern (again, not blocking) has to do with deprecation. I think it's okay to immediately start producing keys using the new encoding from this library, but for repos that already exist, old clients will be around for a *long* time. Do we have a migration story there? I guess just wait until hopefully all the old clients are gone and then cut over?

asraa:
> but for repos that already exist, old clients will be around for a long time. Do we have a migration story there? I guess just wait until hopefully all the old clients are gone and then cut over?

Right! So they can still use `pkg/deprecated` for verification, and can still construct . To migrate, they need to re-sign using the new keys, like we are doing in Sigstore. I could add the deprecated signer to `pkg/deprecated`, which would allow them to still `Sign` with the keys in their key store with the deprecated format. I think this is worth it so I'll add.

repo.go
asraa: I think you could just use ` data.DefaultExpires(role)`? We may want to expose a `data.DefaultScheme` but I don't think it's necessary given the function sigs.

pkg/keys/ecdsa.go
asraa: Since Go doesn't have assertions, I can go nuclear and use `panic(err)`: I think we should do that until the issue I filed adding graceful error handling here is resolved.

repo.go
znewman01: Yeah, I'm wondering if maybe something like this (from Rob Pike) would be appropriate: https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

Or maybe an options struct (number 2 [here](https://golang.cafe/blog/golang-functional-options-pattern.html)

(Totally fine to punt on this for now)

pkg/keys/ecdsa.go
trishankatdatadog: Where is the 🔥 reaction emoji when you need it...

znewman01:
> To migrate, they need to re-sign using the new keys, like we are doing in Sigstore.

Won't this immediately break every old client?

>  I could add the deprecated signer to pkg/deprecated[...]I think this is worth it so I'll add.

+1

> EDIT: I just realized that adding the deprecated signer AND verifier into the same package for compatibility is problematic for users who need to make the transition. E.g. I don't want the signer to ONLY be the deprecated version. So: the deprecated package should actually handle EITHER OR format during the transition. WDYT?

I'm not sure I'm parsing. I think there are the following possible behaviors:

For parsing:
- (default) only accept the new format
- (configurable) accept both old and new formats
- (could be an option, but I don't think we should add it) accept only the old format

This can probably be globally configurable, via the `init()` trick.

For writing new keys:
- (default) write keys in the new format
- (configurable) write keys in the old format

IMO we'd want to configure this on a key-by-key basis, rather than globally.

It'd be nice if we could do *both*, but I don't think that's possible.

asraa:
> Won't this immediately break every old client?

Those old clients would need to import `pkg/deprecated` into their binaries. Originally I had deprecated support as a default, but then we opted to move it to an explicit package.

> +1

ECDSA Signing was never supported FYI. I just realized this as I was implementing. So I don't think it breaks old clients. They would have already implemented their own signing logic (as we did).

> (configurable) write keys in the old format

Don't think we need to opt to do this anymore, since it was never supported in the first place. go-tuf only had verification support for keys, which remains on import of `pkg/deprecated`

D: sorry this is confusing!

pkg/deprecated/ecdsa.go
```diff
@@ -0,0 +1,109 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+)
+
+func init() {
```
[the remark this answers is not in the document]
asraa: Having access to the functions without the type maps is very little use: most people want to be importing this to use in the Repo or Client logic, where all keys are pulled under the hood through these maps.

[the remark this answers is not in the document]
asraa: Ack for testing. That makes sense I think, but I'm again not sure if it'll be used outside the tests in this package already

trishankatdatadog:
> Those old clients would need to import `pkg/deprecated` into their binaries. Originally I had deprecated support as a default, but then we opted to move it to an explicit package.

Hold on: old clients that do not import `pkg/deprecated` will not break so long as the server keeps producing these deprecated-style public keys.

asraa:
> Hold on: old clients that do not import pkg/deprecated will not break so long as the server keeps producing these deprecated-style public keys.

Maybe we're confused! Old clients that use old go-tuf certainly won't break because they're using an old go-tuf version. Clients that update their go-tuf dependency will break, because now ECDSA verifier expects TUF-compliant keys. Unless they import that deprecated package that understands both. RIght?

trishankatdatadog:
> Maybe we're confused! Old clients that use old go-tuf certainly won't break because they're using an old go-tuf version. Clients that update their go-tuf dependency will break, because now ECDSA verifier expects TUF-compliant keys. Unless they import that deprecated package that understands both. RIght?

Oui. Sorry, just need to be super-clear here about what we meant about "old clients" here and so on. Thanks for the clarification!

znewman01:
IIUC, serializing keys was never done *directly* by go-tuf. If that's true, then I am okay with 

(I'm a little confused by the discussion about signing and verifying—the signature format was always okay, right? We're just talking about signers/verifiers because that's the part of the codebase where key serialization happened?)

> > > To migrate, they need to re-sign using the new keys, like we are doing in Sigstore.
> >
> > Won't this immediately break every old client?
> 
> Those old clients would need to import `pkg/deprecated` into their binaries. Originally I had deprecated support as a default, but then we opted to move it to an explicit package.

Let me clarify: won't that immediately break every *deployed* client that's still using the old version of go-tuf?

I'm still fuzzy on the migration story. "Wait a long time before cutting over" isn't a great solution IMO; you typically don't want to break clients no matter how old.

I guess an alternative would be to replace the keys with some other key type (say, RSA for the sake of example) whose format was never affected. Then, we can distribute a new `5.root.json` with only RSA signatures (which cuts off the historical ECDSA-hex signatures, and *then* we could actually turn off the `pkg/deprecated` flag for parsing new clients. That brings the repo into compliance with the TUF spec (err, as long as you ignore everything before `5.root.json`).

pkg/deprecated/ecdsa.go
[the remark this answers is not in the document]
znewman01: IMO it's worth it for testing alone. This would make it so that the tests didn't have to implement fake verifiers; I think that gives us more confidence in the cross-compatibility of the old/new versions.

Further, I think "importing this thing changes the behavior everywhere" is pretty surprising, even if it's documented, and so naming it something that makes it clearer that it's a destructive operation is a good idea.

pkg/deprecated/ecdsa.go
```diff
@@ -0,0 +1,109 @@
...
+	keys.VerifierMap.Delete(data.KeyTypeECDSA_SHA2_P256)
+	keys.VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, NewDeprecatedEcdsaVerifier)
```
znewman01, line 20: IIUC, this isn't race-y because this package depends on the package that happens to add the new verifier to the verifier map.

If that's so, then the `import` statement above is load-bearing; might be worth a comment of warning.

Alternatively (probably better) you could do a check in the other verifier map that stores only if there is no existing entry (I think that's what `LoadOrStore` does).

Either way, a comment about why this is okay would be warranted.

pkg/deprecated/ecdsa.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/pkg/keys"
+)
+
+func init() {
+	keys.VerifierMap.Delete(data.KeyTypeECDSA_SHA2_P256)
```
znewman01, line 19: Is this necessary? IIUC `sync.Map.Store` will clobber an existing entry (which is what we want).

asraa:
> (I'm a little confused by the discussion about signing and verifying—the signature format was always okay, right? We're just talking about signers/verifiers because that's the part of the codebase where key serialization happened?)

Correct, signature format was always correct, it's just the way that the public key was serialized in the root metadata or how the private key was stored in the repo keystore.

> Let me clarify: won't that immediately break every deployed client that's still using the old version of go-tuf?

If they don't update their go-tuf version, they are unaffected: hex ecdsa keys will be the implementation they have in that go-tuf version. Maybe you're asking, if someone is using an old version old client, and the repository owner adds new ECDSA keys their metadata, will the old deployed client break? Yes, but that's akin to using a new feature, right? I'm not sure there's a way to get compatibility here.

> you typically don't want to break clients no matter how old.

I'd say specification deviation and lack of cross-language compatibility is one potential reason :) My feeling though is that this ONLY affected go-tuf clients who were implementing their own ECDSA signers that were compatible with the bad ECDSA verifiers in go-tuf, so they are already aware of managing this implementation.

>  Then, we can distribute a new 5.root.json with only RSA signatures (which cuts off the historical ECDSA-hex signatures, and then we could actually turn off the pkg/deprecated flag for parsing new clients. 

I see. The current plan would have already supported this, since the 5.root.json would have 5 old ECDSA signatures and 5 new ECDSA signatures. Old clients would think that the new keys are bogus, but the old keys are still there as is. There is a danger of failure on parsing though, rather than ignoring.

Because I need this PR this week, I will explore the RSA solution. I just checking in whether Go PIV tools can support RSA and some keys may

pkg/deprecated/ecdsa.go
asraa: You're right, I don't need this, noted the import path from your other comment!

znewman01:
Note: use `feat!:` for a breaking change

pkg/deprecated/ecdsa.go
asraa: Testing suites mess this up, so we can't check that PEM-encoded keys are loaded at first (because each test suite will run init, and the value will be present already).

I added a comment and a check in the deprecation pkg that verifies that we DO override the only PEM verifiers in pkg/keys

pkg/deprecated/ecdsa.go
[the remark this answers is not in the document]
asraa: Separated out the init and the package code itself: users can implement tests with a deprecated ecdsa verifier, without importing the key into the map.

asraa:
When I have approval, I'll squash mine and Toby's commits to one and fix the commit to have `feat!`

asraa:
Squashed, and commit message fixed, added DCO to fix Toby's old DCO issue

trishankatdatadog (dismissed):

pkg/deprecated/ecdsa.go
znewman01: Nice! I think I'd still *also* like to modify the `pkgs/keys/ecdsa.go:init()` to use `LoadOrStore` instead of `Store`, so we never accidentally replace it.

znewman01 (dismissed):

znewman01 (approved):

trishankatdatadog (approved):

PR #369, opened 2022-09-02
fix(verify):  Fix a vulnerability in the verification of threshold signatures (due to handling of keys with multiple IDs)

cedricvanrompay-datadog:
Release Notes: Fixes a vulnerability in the verification of threshold signatures (due to handling of keys with multiple IDs)

**Types of changes**:

Go-TUF has a security vulnerability in the way it validates threshold signatures: it relies on the key IDs declared in the metadata file it is currently validating (varibale `sig.KeyID`) to tell if this new signature should count towards the threshold, assuming that these IDs will be the same as the one it generates (`verifier.MarshalPublicKey().IDs()`). This assumption can be made false and this can be leveraged to forge valid threshold signatures using a number of keys which, in theory, should be insufficient.

This PR adds a test demonstrating the vulnerability, then it fixes the vulnerability.

My opinion however is that Go-TUF should stop trying to be compatible with [old versions of TUF implementations that were assigning several IDs to a same key](https://github.com/theupdateframework/go-tuf/blob/8e84384bebe3b215a8f28f734f8227f71f737c6c/data/types.go#L50-L70), as this makes signature validation way too error-prone. Note that all versions of the TUF specification since version 1 are very clear on there being a single correct way to compute the key ID: 

> KEYID: The identifier of the key signing the ROLE object, which is a hexdigest of the SHA-256 hash of the canonical form of the key. The keyid MUST be unique in the "signatures" array: multiple signatures with the same keyid are not allowed.
>
> source: https://theupdateframework.github.io/specification/v1.0.29/index.html#role-keyid

> KEYID: A KEYID, which MUST be correct for the specified KEY. Clients MUST calculate each KEYID to verify this is correct for the associated key. Clients MUST ensure that for any KEYID represented in this key list and in other files, only one unique key has that KEYID.
>
> source: https://theupdateframework.github.io/specification/v1.0.29/index.html#root-keyid

Also, I am aware that [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md) (which does not apply to TUF 1.0.x anyway) introduces key ID flexibility. A better solution than the one in this PR would be to apply the recommendation in TAP 12 to _"use a standardized representation for keys, like the modulus and exponent for RSA or the point and curve for ECC"_. That's probably something we should do in Go-TUF without waiting for TUF 1.1.

### DIFF
--- a/verify/verify.go
+++ b/verify/verify.go
@@ -92,8 +92,8 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 	// Verify that a threshold of keys signed the data. Since keys can have
 	// multiple key ids, we need to protect against multiple attached
 	// signatures that just differ on the key id.
-	seenKeys := make(map[string]struct{})
-	valid := 0
+	verifiedKeyIDs := make(map[string]struct{})
+	numVerifiedKeys := 0
 	for _, sig := range s.Signatures {
 		if !roleData.ValidKey(sig.KeyID) {
 			continue
@@ -104,6 +104,7 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 		}
 
 		if err := verifier.Verify(msg, sig.Signature); err != nil {
+			// FIXME: don't err out on the 1st bad signature.
 			return ErrInvalid
 		}
 
@@ -115,20 +116,20 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 		keyIDs := verifier.MarshalPublicKey().IDs()
 		wasKeySeen := false
 		for _, keyID := range keyIDs {
-			if _, present := seenKeys[keyID]; present {
+			if _, present := verifiedKeyIDs[keyID]; present {
 				wasKeySeen = true
 			}
 		}
 		if !wasKeySeen {
 			for _, id := range keyIDs {
-				seenKeys[id] = struct{}{}
+				verifiedKeyIDs[id] = struct{}{}
 			}
 
-			valid++
+			numVerifiedKeys++
 		}
 	}
-	if valid < roleData.Threshold {
-		return ErrRoleThreshold{roleData.Threshold, valid}
+	if numVerifiedKeys < roleData.Threshold {
+		return ErrRoleThreshold{roleData.Threshold, numVerifiedKeys}
 	}
 
 	return nil
--- a/verify/verify.go
+++ b/verify/verify.go
@@ -92,7 +92,7 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 	// Verify that a threshold of keys signed the data. Since keys can have
 	// multiple key ids, we need to protect against multiple attached
 	// signatures that just differ on the key id.
-	seen := make(map[string]struct{})
+	seenKeys := make(map[string]struct{})
 	valid := 0
 	for _, sig := range s.Signatures {
 		if !roleData.ValidKey(sig.KeyID) {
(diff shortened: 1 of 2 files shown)


cedricvanrompay-datadog:
I realise I should probably have shared this privately with the maintainers instead of sending a public PR, apologies.

trishankatdatadog:
Thanks for reporting this. I had found a [similar issue](https://github.com/theupdateframework/python-tuf/security/advisories/GHSA-r7vq-6425-j94w) for python-tuf before, but unfortunately we never got around to publishing it, which may explain why no other implementations had noticed.

It looks good on a cursory glance, but let us review and merge ASAP.

verify/verify.go
```diff
@@ -92,7 +92,7 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 	// Verify that a threshold of keys signed the data. Since keys can have
```
ethan-lowman-dd, line 92: While this section is getting some attention: I just found that this early return will cause the threshold signature check to fail if any single signature is invalid, even if there is a threshold of valid signatures. To fix, we should replace `return ErrInvalid` with `continue`.

```go
if err := verifier.Verify(msg, sig.Signature); err != nil {
	return ErrInvalid
}
```

trishankatdatadog: Yes, please refer to the [spec](https://theupdateframework.github.io/specification/latest/index.html) ("a threshold of keys" to prevent arbitrary software attacks) and [python-tuf](https://github.com/theupdateframework/python-tuf/blob/7a760691c69bb4d58d37245d9b5132ee91f20487/tuf/api/metadata.py#L438-L452).

trishankatdatadog: And let's add unit tests to make sure this mistake gets caught.

verify/verify.go
```diff
@@ -109,9 +109,19 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
...
 		// Only consider this key valid if we haven't seen any of it's
 		// key ids before.
-		if _, ok := seen[sig.KeyID]; !ok {
-			for _, id := range verifier.MarshalPublicKey().IDs() {
-				seen[id] = struct{}{}
+		// Careful: we must not rely on the key IDs _declared in the file_,
+		// instead we get to decide what key IDs this key correspond to.
+		// XXX dangerous; better stop supporting multiple key IDs altogether.
+		keyIDs := verifier.MarshalPublicKey().IDs()
+		wasKeySeen := false
```
trishankatdatadog, line 116: Optimisation nit: we should perform this check _before_ we even verify the signature.

verify/verify.go
```diff
@@ -92,7 +92,7 @@ func (db *DB) VerifySignatures(s *data.Signed, role string) error {
 	// Verify that a threshold of keys signed the data. Since keys can have
 	// multiple key ids, we need to protect against multiple attached
 	// signatures that just differ on the key id.
-	seen := make(map[string]struct{})
+	seenKeys := make(map[string]struct{})
 	valid := 0
 	for _, sig := range s.Signatures {
```
trishankatdatadog, line 97: While we are fixing this: let's not iterate over the signatures first, but instead over the expected keyids (SHA2-256 is fine for now), then we extract the signature for the keyid and check it.

cedricvanrompay-datadog: I think that the behavior would be strictly equivalent since the first thing we do in `for _, sig := range s.Signatures {` is `if !roleData.ValidKey(sig.KeyID) { continue }`. However if we were iterating over `roleData.KeyIDs` we won't be able to do lookup in `s.Signatures` which is a slice and note a map. We could create a map over `s.Signatures` but that's more code and more opportunities for bugs.

mnm678:
I'll do a more detailed review, but it looks like this needs DCO and [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

> I realise I should probably have shared this privately with the maintainers instead of sending a public PR, apologies.

Yes, for future reference, but thanks for sharing this. We'll work to improve our security reporting docs to make the private reporting process more clear in the future.

verify/verify.go
mnm678: What is the expected behavior if there are two signatures from the same keyid, but one is invalid? I think leaving this here ensures that the valid signature would be used, whether or not it is the first one in the list

mnm678:
LGTM. I'll approve pending the failing checks

verify/verify.go
trishankatdatadog: Decided not to fix right now because it will break: (1) tests that expect the broken behaviour, and more importantly, (2) clients that expect this same behaviour. Will create another issue and track it there.

verify/verify.go
trishankatdatadog: You are right: leaving this alone.

verify/verify.go
trishankatdatadog: I like the python-tuf approach of iterating over expected keyIDs the best, but let's track it another issue and fix it going fwd.

ethan-lowman-dd (approved):

trishankatdatadog (approved):
LGTM, @cedricvanrompay-datadog, @ethan-lowman-dd, and I discussed this over Zoom, thanks!

mnm678 (approved):

PR #373, opened 2022-09-06
add back SnapshotFileMeta.Custom

arbll:
Adds back a `Custom` field to `SnapshotFileMeta` following its removal in https://github.com/theupdateframework/go-tuf/pull/345.

Release Notes: Add back a `Custom` field to `SnapshotFileMeta`

**Types of changes**:

We rely on this `Custom` field to store some custom data related to delegated `targets`. The removal of this field would force us to store this outside of the signed TUF file or in a hacky map in `Snapshot.Custom`.

### DIFF
--- a/data/types.go
+++ b/data/types.go
@@ -161,9 +161,10 @@ func (f Hashes) HashAlgorithms() []string {
 }
 
 type metapathFileMeta struct {
-	Length  int64  `json:"length,omitempty"`
-	Hashes  Hashes `json:"hashes,omitempty"`
-	Version int64  `json:"version"`
+	Length  int64            `json:"length,omitempty"`
+	Hashes  Hashes           `json:"hashes,omitempty"`
+	Version int64            `json:"version"`
+	Custom  *json.RawMessage `json:"custom,omitempty"`
 }
 
 type SnapshotFileMeta metapathFileMeta


trishankatdatadog:
I'm not strictly against this PR, but a few things:
1. Why just Snapshot? Why not the base Metadata itself, while we are it?
2. I don't think we should keep indefinitely supporting this custom (pun intended) use case. If we add this back, we should add a deprecation notice to observe that we can remove this later.

asraa:
Thanks @arbll! 

As with @trishankatdatadog, this field is not in the TUF specification: https://theupdateframework.github.io/specification/latest/#file-formats-snapshot

If you want to keep information about a delegated target, then ideally the information should live in/signed by the parent role, rather than the snapshot role, correct?

It seems like you may want CUSTOM metadata inside a `DELEGATIONS` object long-term, right? https://theupdateframework.github.io/specification/latest/#delegations

I'm on the side of adding this specifically to `SnapshotFileMeta`, and not to the base `FileMeta` that TimestampFileMeta uses, I'd rather scope this usage minimally.

trishankatdatadog:
> I'm on the side of adding this specifically to `SnapshotFileMeta`, and not to the base `FileMeta` that TimestampFileMeta uses, I'd rather scope this usage minimally.

Agreed. Is there a good way to make this optional and not built or at least included in the metadata (even as an empty field) by default?

trishankatdatadog:
> I'm on the side of adding this specifically to `SnapshotFileMeta`, and not to the base `FileMeta` that TimestampFileMeta uses, I'd rather scope this usage minimally.

I also want to make sure that we only need this kind of surgical addition, and not have additional PRs later adding this in other places.

asraa (approved):
Approval, but to accomodate, let's add a commend above `type SnapshotFileMeta` that it includes `Custom`.

Don't think end-users will even see the field if it isn't populated, but worth a comment

trishankatdatadog:
> Approval, but to accomodate, let's add a commend above `type SnapshotFileMeta` that it includes `Custom`.
> 
> Don't think end-users will even see the field if it isn't populated, but worth a comment

@arbll pls make the change? thx!

trishankatdatadog:
Thanks @arbll. Would you please:
1. Sign off DCO and rebase + force-push
2. Confirm that custom is not populated by default unless explicitly set?

arbll:
@trishankatdatadog Done

trishankatdatadog (approved):
I'm not crazy about this PR (we should add a deprecation notice and figure out alternative), but happy to unblock for now.

Issue #380, opened 2022-09-09
[bug] RemoveTargetFromExpires attempts to remove a target from versioned top-level roles


asraa:
In https://github.com/theupdateframework/go-tuf/blob/61872a3ac6e6a475771c23bf1592a00c1773b3e7/repo.go#L1195

we will remove a target from all delegated roles. However, the check `roles.IsDelegatedTargetsManifest(metaName)` will return TRUE on metadata like `x.root.json` or `x.snapshot.json`, because it's implementation only checks for exact match on the top-level roles like `root.json` or `.snapshot.json`.

I can put out a quick PR for this. It's just annoying because the log outputs that messages when a target was not found in the metadata: `[2.root.json] The following target is not present: rekor.0.pub`. 

https://github.com/theupdateframework/go-tuf/blob/61872a3ac6e6a475771c23bf1592a00c1773b3e7/repo.go#L1226

PR #381, opened 2022-09-09
fix: fix is top level calculation for versioned manifests

asraa:

Fixes https://github.com/theupdateframework/go-tuf/issues/380
Release Notes: 
* fix: fixes bug in calculation of top-level manifest for versioned manifests

**Types of changes**:

Fixes the calculation of whether a manifest is a top-level one when the manifest is versioned. Previously, a manifest like 1.root.json would be detected as a delegated manifest because it was 1.root was not in the set of top-level manifests. This now accounts for versioned manifests.

- [ x ] Tests have been added for the bug fix or new feature

### DIFF
--- a/internal/roles/roles.go
+++ b/internal/roles/roles.go
@@ -22,6 +22,9 @@ func IsDelegatedTargetsRole(name string) bool {
 }
 
 func IsTopLevelManifest(name string) bool {
+	if IsVersionedManifest(name) {
+		name = strings.Join(strings.Split(name, ".")[1:], ".")
+	}
 	return IsTopLevelRole(strings.TrimSuffix(name, ".json"))
 }
 
--- a/internal/roles/roles_test.go
+++ b/internal/roles/roles_test.go
@@ -24,18 +24,24 @@ func TestIsDelegatedTargetsRole(t *testing.T) {
 
 func TestIsTopLevelManifest(t *testing.T) {
 	assert.True(t, IsTopLevelManifest("root.json"))
+	assert.True(t, IsTopLevelManifest("1.root.json"))
 	assert.True(t, IsTopLevelManifest("targets.json"))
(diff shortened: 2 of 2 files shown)


internal/roles/roles.go
```diff
@@ -22,6 +22,9 @@ func IsDelegatedTargetsRole(name string) bool {
 }
 
 func IsTopLevelManifest(name string) bool {
+	if IsVersionedManifest(name) {
+		name = strings.Join(strings.Split(name, ".")[1:], ".")
```
trishankatdatadog, line 26: Maybe slightly safer when splitting by '.' to get only the last 2 tokens, but the function below will catch shenanigans anyway.

znewman01: [all nitpicks]

Note to self: it should never be possible that `name` has no `.` in it, because `IsVersionedManifest` checks that. Maybe worth an assertion (via `panic`)? (There's currently an implicit assertion when we index into the result of `Split` 😄 )

IMO this is nicer as `name := strings.SplitN(name, ".", 2)[1]` or `_, name, _ := strings.Cut(name, ".")`—WDYT?

znewman01 (approved):

internal/roles/roles.go
asraa: Nice! I forgot about `SplitN` and `Cut`.

The `Cut` solution is really nice because I can assert that it was found

znewman01 (dismissed):
LGTM if you can make lint happy

Also  do you need to declare `var found bool` on a separate line?

asraa:
> Also do you need to declare var found bool on a separate line?

Yes, only because I re-use the `name` field. I either have to declare another name var on another separate line or that one

znewman01 (approved):

mnm678 (approved):

Issue #376, opened 2022-09-08
go-tuf logs to stdout/stderr when used as a library


tamird:
Internal operations log directly to stdout/stderr starting with #160.

It would be preferable not to interact with stdout/stderr except from within cmd/.

asraa:
I agree! We get logging from libraries. Repo commands can be logged directly from the cmd/.

We can make sure there is no logging done in `repo.go` or `pkg/`

What about our deprecated code? I think it's ok to remove the logging that you are using a deprecated key, since libraries will obviously import them proactively.

joshuagl:
Thanks for filing this, and for the suggestion in https://github.com/theupdateframework/go-tuf/pull/160#issuecomment-1241063478

> Would it be possible to thread an `io.Writer` through rather than calling `fmt.Print{f,ln}` directly?

That's a good suggestion. I think we want to:
* switch from `fmt.Print{f,ln}` to `log.Print{f,ln}`
* take an optional `io.Writer` in library code (notably repo.go) so the caller can choose where the logging messages go

asraa:
I'm going to prioritize this and take this on: `cosign` users are complaining :)

asraa:
So currently I'm piping an optional logger through the client.

I would like to only log the tuf ecdsa deprecated message once per client. that would maybe require a singleton in the client for that message? that feels sorta messy. anyone have any other nice go suggestions?

trishankatdatadog:
> anyone have any other nice go suggestions?

@ethan-lowman-dd wdyt?

asraa:
It could be that we don't care at all for repeated lines now that we can divert the logs to a file!

PR #385, opened 2022-09-16
feat: pass logger into repo and client

asraa:

Fixes https://github.com/theupdateframework/go-tuf/issues/376

Release Notes: 
```
* feat: Allows passing an optional logger into `repo`. Default uses `io.Discard` to swallow logs.
* feat: Exposes a `repo.NewRepoWithOpts` that can be used to instantiate a repo with repo options like prefix, indent, hash algorithms, and logger.
```

**Types of changes**:

- [ x ] New feature (non-breaking change which adds functionality)

Client and Repo can now pass in an optional logger. This will not break any existing clients, the argument is optional.

- [ x ] Tests have been added for the bug fix or new feature

**This is pretty preliminary, please let me know if you agree/disagree with my approach. 

I would still need to fix the CLI as a follow-up, this PR as it stands totally removes logging since none is explicitly passed.
**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,3 +13,4 @@ linters:
     - gosimple
     - unused
     - typecheck
+    - forbidigo
--- a/cmd/tuf-client/main.go
+++ b/cmd/tuf-client/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"log"
+	"os"
 
 	docopt "github.com/flynn/go-docopt"
 	tuf "github.com/theupdateframework/go-tuf/client"
@@ -32,7 +33,7 @@ See "tuf-client help <command>" for more information on a specific command.
 
 	if cmd == "help" {
 		if len(cmdArgs) == 0 { // `tuf-client help`
-			fmt.Println(usage)
+			fmt.Fprint(os.Stderr, usage)
 			return
 		} else { // `tuf-client help <command>`
 			cmd = cmdArgs[0]
--- a/cmd/tuf/main.go
+++ b/cmd/tuf/main.go
@@ -58,7 +58,7 @@ See "tuf help <command>" for more information on a specific command
 
 	if cmd == "help" {
 		if len(cmdArgs) == 0 { // `tuf help`
-			fmt.Println(usage)
+			fmt.Fprint(os.Stderr, usage)
 			return
 		} else { // `tuf help <command>`
 			cmd = cmdArgs[0]
@@ -115,7 +115,11 @@ func runCommand(name string, args []string, dir string, insecure bool) error {
 	if !insecure {
 		p = getPassphrase
 	}
-	repo, err := tuf.NewRepo(tuf.FileSystemStore(dir, p))
+	logger := log.New(os.Stdout, "", 0)
+	storeOpts := tuf.StoreOpts{Logger: logger, PassFunc: p}
+
+	repo, err := tuf.NewRepoWithOpts(tuf.FileSystemStoreWithOpts(dir, storeOpts),
+		tuf.WithLogger(logger))
 	if err != nil {
 		return err
 	}
--- a/pkg/keys/deprecated_ecdsa.go
+++ b/pkg/keys/deprecated_ecdsa.go
@@ -9,13 +9,25 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
+	"log"
 
 	"github.com/theupdateframework/go-tuf/data"
 )
 
-func NewDeprecatedEcdsaVerifier() Verifier {
-	return &ecdsaVerifierWithDeprecatedSupport{}
+var (
+	WarnDeprecatedEcdsaKey = "tuf: warning using deprecated ecdsa hex-encoded keys"
+)
+
+func NewDeprecatedEcdsaVerifier(opts ...VerifierOpts) Verifier {
+	verifier := &ecdsaVerifierWithDeprecatedSupport{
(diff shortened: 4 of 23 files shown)


pkg/keys/deprecated_ecdsa.go
```diff
@@ -98,6 +116,6 @@ func (p *deprecatedP256Verifier) UnmarshalPublicKey(key *data.PublicKey) error {
 	}
 
 	p.key = key
-	fmt.Fprintln(os.Stderr, "tuf: warning using deprecated ecdsa hex-encoded keys")
+	p.logger.Print(WarnDeprecatedEcdsaKey)
```
znewman01, line 119: Maybe gate this behind a https://pkg.go.dev/sync#Once in the struct?

Is "once per verifier" okay? Or is that still too much?

znewman01: TBH I'm not convinced we need to have this at all, if clients opt into it, they probably intend for this to be the case.

znewman01 (approved):
BTW nice refactor with the "Opts" pattern :+1:

pkg/keys/deprecated_ecdsa.go
asraa: RIGHT! I went through the same thought. 

I thought: once per client makes sense, but then I thought: this really shouldn't be happening, and if you feel logs are spammy, go log to a file you don't care for

znewman01: Oh, I meant the opposite—do we need to log at all? It shouldn't be a surprise to anybody that this is happening.

asraa: Oh. I think it's hidden behind 3 layers of deps (cosign -> sigstore -> go-tuf) if not more, so I thought so.

It would be nice for my sigstore layer testing in root-signing to have the affirmation that no longer need that codepath and we can deprecate it. But I'm up for just removing it entirely.

znewman01: I don't feel strongly. But I do think it's kinda gross to expose this to *clients* (e.g. cosign users). I guess Cosign can just swallow it which helps.

Up to you!

asraa: Yes, the point is that by default it does not log any messages at all, but we can flip the log in, let's say, the root-signing presubmits.

Will think! I have to sign off :/

asraa: @theupdateframework/go-tuf-maintainers OK: I think I'm going to remove the deprecated ECDSA log message, and leave logging to the repository top-level options.

Are there any client operations that would benefit from info logging? logging updated metadata versions? I went through, and nothing seems like an exceptional case that requires logging (to me). If not, I can just clear out all client logging.

znewman01: IMO logging at the debug level would be really useful for complicated things like delegations. But that can happen later; I'm happy to have it all ripped out now.

asraa: True! Searching delegations for a method like `GetTarget` would benefit from debug logging which target delegation was required. 

Removing :) someone can take a look at this PR if they want to grab the client logging code then.

znewman01 (approved):

.golangci.yml
```diff
@@ -13,3 +13,4 @@ linters:
     - gosimple
     - unused
     - typecheck
+    - forbidigo
```
trishankatdatadog, line 16: What does this do?

znewman01: https://github.com/ashanbrown/forbidigo

It's a golanglint check that looks for `fmt.Print` and friends! (It can be configured to look for other regexes too)

trishankatdatadog (approved):

cmd/tuf-client/main.go
```diff
@@ -32,7 +33,7 @@ See "tuf-client help <command>" for more information on a specific command.
 
 	if cmd == "help" {
 		if len(cmdArgs) == 0 { // `tuf-client help`
-			fmt.Println(usage)
+			fmt.Fprint(os.Stderr, usage)
```
tamird, line 36: why is usage going to stderr?

cmd/tuf/main.go
```diff
@@ -58,7 +58,7 @@ See "tuf help <command>" for more information on a specific command
 
 	if cmd == "help" {
 		if len(cmdArgs) == 0 { // `tuf help`
-			fmt.Println(usage)
+			fmt.Fprint(os.Stderr, usage)
```
tamird, line 61: Ditto.

cmd/tuf-client/main.go
znewman01: 🤷  I've seen it both ways. There are some tools that define an explicit, **strict** "everything that gets logged to stdout must be computer-readable" philosophy and apply that to help text too. Many tools do, of course, print usage on STDOUT.

See for instance [this request for usage on STDERR](https://github.com/cbeust/jcommander/issues/339). I particularly like the recommendation to follow [picocli](https://picocli.info/) (just one example):

> when the user specified invalid input, the error message and usage help should be printed to stderr, but when the user requests help with <cmd> --help, the usage help should be printed to stdout so users can pipe the output to tools like grep and less.

asraa: Ah, thank you for the comment: Yeah, I think I originally assumed that help should be part of STDERR, but here, since the user explicitly requested for help, I think I concur, it should be STDOUT. I'll put out a fix PR.

PR #396, opened 2022-09-23
feat: Tap 4 Implementation (TUFClient interface)

### LINKED ISSUE #348: Implement TAP 4: Multi-repository mappings
This tracking issue is to implement [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md): Multiple repository consensus on entrusted targets.

To implement this, I propose creating a NEW type `MultiRepositoryClient` which takes a `map.json` file. It would expose `NewClient`, `Init`, `Update`, `Targets`, and  `Download` similar to the existing client.

Internally, it would contain a map from repository name to existing TUF client updaters. The logic for priority and consensus would exist in the `MultiRepositoryClient`.

The purpose of this is to use in Sigstore's TUF client to support multiple repositories, augmenting the public root with custom, hosted, or test repositories with an AND relationship.

cc @kommendorkapten @haydentherapper

dibrinsofor:
Fixes #348
Release Notes: 

**Types of changes**:

**Description of the changes being introduced by the pull request**: Began abstracting Client into interface in `interface.go` for future use towards multi-client mapping. currently setup with `Update`, `Download`, `VerifyDigest`, `Target` and `Targets` functionality. with @znewman01 and @vaikas

### DIFF
--- a/docs/MAINTAINERS
+++ b/docs/MAINTAINERS
@@ -3,5 +3,4 @@ Trishank Karthik Kuppusamy <trishank.kuppusamy@datadoghq.com> (github: trishanka
 Joshua Lock <jlock@vmware.com> (github: joshuagl)
 Marina Moore <mm9693@nyu.edu> (github: mnm678)
 Zack Newman <zjn@chainguard.dev> (github: znewman01)
-Hossein Siadati <hossein.siadati@datadoghq.com> (github: hosseinsia)
 Radoslav Dimitrov <dimitrovr@vmware.com> (github: rdimitrov)


asraa:
Thanks for starting this! Do you need any help? Happy to set up a chat or a sync

znewman01:
I owe an update here sorry! Probably will help make progress early next week

dibrinsofor:
Sorry I've been busy, I think I have an idea on how to approach this. I'll spend the next couple of days trying to get it setup and reach out if I have any challenges. @znewman01 @asraa

dibrinsofor:
I was thinking of extending the new client interface to include the `Init`, will `Init` work the way it does with Multi Clients?

Do you have any ideas beyond this? or can you offer some suggestions? @znewman01

znewman01:
Hey Dibri!

Sorry for the delay here. This is my advice for how to implement this feature.

The overall plan:

1.  Get the current PR merged. I think it&rsquo;s useful as-is! I can give some feedback soon.
2.  Make a new file, `pkg/client/multirepository.go`.
3.  Write a simple `Map` struct, based on the example JSON file in TAP-4. Note that this JSON format is not actually part of the TAP! But it&rsquo;s a good inspiration.
4.  Write a multirepository client struct.
    -   This basically holds a bunch of `TUFClient`s along with the &ldquo;mappings&rdquo; from the `Map`
    -   I recommend that you start with a commit that includes just the struct and implements the interface, but all the interface methods just `panic()` when you call them.
5.  Initialize the multirepository client from a `Map`
6.  Uses the TAP-4 algorithms to implement the `TUFClient` interface.
7.  Write tests.

It'd be great to do these as individual steps—much easier for you to merge many small PRs than one big one.

I also recommend you take a look at the [reference implementation, in Python](https://github.com/theupdateframework/python-tuf/pull/504/files#diff-9ea8a1bf8ff58c5148fbd8bafc0792dde8ef97ec8d86d998119e0f2537026046) for inspiration.

More detail on the above steps below; all of my code examples are untested and might be totally broken, so feel free to ignore any of my suggestions!


### Map struct

Based on TAP 4 `map.json` file. Here&rsquo;s what it could look like:

```go
type struct Mapping {
    paths []string
    repositories []names
    // threshold of repos to sign for a specific target
    threshold int
    terminating bool
}

type struct Map {
    // List of repository names
    repositories: []string
    mappings: []Mapping
}
```

You can implement useful things like path matching on these:

```go
func (m *Mapping) MatchesPath(path string) bool {
    for pathPattern := range paths {
        // https://pkg.go.dev/path#Match
        // TODO: double check that `path.Match` is the same as PATHPATTERN:
        // https://theupdateframework.github.io/specification/latest/#pathpattern
        matches, err := path.Match(pathPattern, path)
        if err != nil {
            // swallow err; some other pathPattern might match
        }
        if matches {
            return true
        }
    }
    return false
}
```

You're going to initialize the `Map` from JSON using `json.Unmarshal`. That means that the data in `Map` could be really weird and invalid!

I'm going to recommend that you introduce the following pattern:

```go
type struct RawMap {
	repositories: map[string]string
	mappings: []Mapping
}

func (r RawMap) ToMap() Map {
	// TODO: validate map!
	// - each mapping must refer only to repositories in `repositories`
	Map {
		repositories: repositories,
		mappings: mappings
	}
}
```

You can make `Map` private (call it `map`) and expose a `Map` interface:

```go
interface Map {
	Repositories() map[string]string
	Mappings() []Mapping
}
```

This way, nobody can ever make a `map`! They have to go through `RawMap` which will do validation.

**Proposed minor change:** instead of a map from repository name to URL, just have a list of repository names.


### MultiRepositoryClient

Then, you can make a `MultiRepositoryClient` class!

```go
type struct MultiRepositoryClient {
	clients: map[string]TUFClient
	mappings: []mappings
}
```

I'd actually recommend **against** adding an `Init()` for the `MultiRepositoryClient`. When you make a `MultiRepositoryClient`, you should require that the clients you pass in are already `Init()`'d.

```go
// `clients` is a map from repository URL to TUFClient.
func NewMultiRepositoryClient(clients: map[string]*TUFClient, map Map) (MultiRepositoryClient, error) {
    mappings := map.Mappings()

    // Find all unique repository names in the `Map`.
    allRepoNames := make([]string)
    for mapping := range mappings {
        for repoName := range mapping.repositories {
            contained := false
            for prevFoundRepoName := range allRepoNames {
                if prevFoundRepoName == repoName {
                    contained = true
                    break
                }
            }
            if !contained {
                allRepoNames = append(allRepoNames, repoName)
            }
        }
    }

    // Make sure we have a client for each repository name.
    clientsByRepoName := make(map[string]Client)
    repositories := map.Repositories()
    for repoName := range allRepoNames {
        repoURLs, ok := repositories
        if !ok {
            return errors.New("missing repository URLs for repo name")
        }
        for repoURL := range repoURLs {
            if client, ok := clients[repoURL]; ok {
                clientsByRepoName[repoName] = client
            }
        }
        if _, ok = clientsByRepoName[repoName]; !ok {
            return errors.New("no client for repo URL")
        }
    }
    client := MultiRepositoryClient {
        clients: reposByName,
        mappings: mappings

    }
    return client, nil
}
```


### TAP-4 Algorithms

`Update()` is pretty easy:

```go
func (mrc *MultiRepositoryClient) Update() (*data.TargetFiles, error) {
    t := make([]*data.TargetFiles, len(mrc.clients))
    for _, client := range mrc.clients {
        t = append(t, client.Update())
    }
    return t // TODO: combine into one big *data.TargetFiles
}
```

Introduce the following helper method; we'll need it later.

```go
// Get all clients matching the given repository names.
func (mrc *MultiRepositoryClient) getClientsByRepositoryNames(repoNames []string) []*Client {
    clients := make([]*Client, len(repoNames))
    for repoName := range repoNames {
        client, ok := mrc.repositories[repoName]
        if !ok {
            panic("MRC in invalid state")
        }
        clients = append(clients, client)
    }
}
```

This is the fun part, where the [TAP-4 logic](https://github.com/theupdateframework/taps/blob/master/tap4.md#searching-for-files-on-multiple-repositories) comes in! I've sketched out what it might look like.

```go
// counter counts how many times we've seen a given value.
// TODO: I might be doing generics wrong.
type struct counter[C any] {
	counts map[C]int
}

func (c counter[C]) add(value C) int {
	count, ok := m.counts[value]
	if ok {
		count += 1
	} else {
		count = 1
	}
	m.counts[value] = count
	return count
}

func (mrc *MultiRepositoryClient) Target(name string) (data.TargetFileMeta, error) {
    // Check each mapping, in the listed order, and identify the first mapping
    // that matches the requested file. In figure 1, the client should choose the
    // second mapping because the glob pattern (foo*.tgz) matches the file.
	// -- TAP 4
	for mapping := range mrc.mappings {
		// Once a mapping is identified for the requested file, TUF metadata is
		// downloaded and verified from the assigned repositories it lists.
		// -- TAP 4
		if mapping.MatchesPath(name) {
			// Verification occurs if the length and hashes about the
			// target match...Custom targets metadata are exempt from
			// this requirement. -- TAP 4
			numMatching := make(counter[FileMeta])
			for client := range mrc.getClientsByRepoNames(mapping.repositories) {
				if target, err := client.Target(name); err != nil {
					continue
				}
				// In TAP 4 we only care about length/hashes; that's what's in `FileMeta`.
				if count := numMatching.add(target.FileMeta); count >= mapping.threshold {
					return metadata
                }
			}
			// If the terminating flag (per element C) is set to true, report
			// that either the repositories do not agree on the target
			// none of them have signed for the target. -- TAP 4
			if mapping.terminating {
				return errors.New("encountered terminating mapping; no match :(")
			}
			// Otherwise, go back to step 1 and process the next mapping that
			// matches the requested file. -- TAP 4
		}
	}
	return errors.New("no match")  // TODO: better errors
}
```

Then, other stuff is pretty easy to implement in terms of `Target()`:

```go
func (mrc *MultiRepositoryClient) Download(name string, dest Destination) (err error) {
    // need to get the *metadata* for `name` and see if it matches before downloading target
    metadata, error := mrc.Target(name)
    if err != nil {
        return err
    }
    // TODO: download target for metadata
    // This might be easier if you make a separate, private method called
    // `MultiRepositoryClient.targetAndClient` that does what `Target()` does
    // now, but *also* returns the repository client that makes the threshold match.
    // Then, `Target()` will just call that and ignore the repository client.
    // `Download` can just call `client.Download(name)`.
}

func (mrc *MultiRepositoryClient) VerifyDigest(digest string, digestAlg string, length int64, path string) error {
    // TODO: you can also copy `client.VerifyDigest`; it should look almost identical.
    target, err := mrc.Target(path)
    if err != nil {
        return err
    }

    if target.FileMeta.length != length {
        return errors.New("invalid length")
    }

    // Get the right has format
    hash, ok := target.FileMeta.Hashes[digestAlg];
    if !ok {
        return errors.New("no hash matching given algorithm")
    }
    if digest != hash {
        return errors.New("bad digest")
    }
    return nil
}

func (mrc *MultiRepositoryClient) Targets() (data.TargetFiles, error) {
    allTargets := make([]data.TargetFiles)
    for repoName, client := range mrc.clients {
        repoTargets, err := client.Targets()
        for target := range repoTargets {
            allTargets = append(allTargets, target)
        }
    }

    targets := make([]data.TargetFiles)
    // One way to do this; not very efficient at all :(
    for target := range targets {
        // The "right" thing do would be to keep track of which targets came from
        // which repos. You'd then have to go through the full TAP-4 "download"
        // procedure based on that.
        _, err := mrc.Target(target.name)
        if err != nil {
            targets = append(targets, target)
        }
    }
    return targets
}
```

### Tests

You want the usual test suite: if you wrap a `Client` in a
`MultiRepositoryClient` based on a `Map` with one repo and one catchall
`Mapping` that accepts all paths (`**/*`) it should behave exactly the same in
the other tests.

You also want specific multi repo tests. See [Python for inspiration](https://github.com/vladimir-v-diaz/tuf/blob/14ffb0d45c7e36b16ce2c0d4198c2eb59a99f1ce/tests/test_updater.py#L1640). I think most test cases will involve 2 repositories, A and B, and mostly test `Target()` with threshold=1:

| `Map`                            | A has target? | B has target? | Result    |
|--------------------------------- | ------------- | ------------- | ----------| 
| one mapping, to either repo      | no            | no            | not found |
|                                  | yes           | no            | found     |
|                                  | no            | yes           | found     |
| one mapping, to repo A           | no            | no            | not found |
|                                  | yes           | no            | found     |
|                                  | no            | yes           | not found |
| two mappings, first terminal     | yes           | yes           | found     |
|                                  | no            | yes           | not found |
| two mappings, first not terminal | yes           | yes           | found     |
|                                  | no            | yes           | found     |

There will also be a bunch of threshold tests. For that, I&rsquo;d use a `Map` with one mapping to three repos, A, B, and C:

| Test case name                         | A has target? | B has target? | C has target? | Threshold | Result        |
| -------------------------------------- | ------------- | ------------- | ------------- | --------- | --------------|
| no repo has target                     | no            | no            | no            | 1,2,3     | not found     |
| one repo has target -- threshold=1     | yes           | no            | no            | 1         | found!        |
| one repo has target --  threshold>1    | yes           | no            | no            | 2,3       | not found     |
| two repos have target -- theshold <= 2 | yes           | yes           | no            | 1,2       | found!        |
| two repos have target -- theshold > 2  | yes           | yes           | no            | 3         | not found!    |
| all repos have target -- any trheshold | yes           | yes           | yes           | 1,2,3     | found         |
| threshold=0                            | ?             | ?             | ?             | 0         | invalid `Map` |
| threshold > len repos                  | ?             | ?             | ?             | 4         | invalid `Map` |

Two more:

-   repos A/B/C/D, A and C have target t1, B and D have target t2. Should find t1!
-   repos A/B/C/D, A has target t1, C has nothing, B and D have target t2. Should find t2!

I think the tests would be a lot easier if you made a fake `TUFClient`:

```go
type struct fakeTUFClient {
	targets TargetFiles
}

func (c *fakeTUFClient) Target(name string) (data.TargetFileMeta, error) {
	targetMeta, ok := c.targets[name]
	if !ok {
		return nil, errors.New("missing")
	}
	return targetMeta, nil
}

func (c *fakeTUFClient) Update() (*data.TargetFiles, error) {
	return c.targets, nil
}

func (c *fakeTUFClient) Download(name string, dest Destination) (err error) {
	panic("not needed for most tests") // if you want, you could put fake content in ~fakeTUFClient~
}

func (c *fakeTUFClient) VerifyDigest(digest string, digestAlg string, length int64, path string) error {
	// this is going to look the exact same as the other ~VerifyDigest~ implementations
}

func (c *fakeTUFClient) Targets() (data.TargetFiles, error) {
	return c.targets, nil
}
```

dibrinsofor:
Thank you for this, @znewman01! It looks very clear. I'll get started right away and reach out via email if I run into any issues.

@asraa can we get this merged as is? or do I have to make any other changes

znewman01 (approved):

znewman01:
@dibrinsofor I think you can take this off of "draft" status

You'll also need to add the "[DCO](https://developercertificate.org/)" with `git commit --amend --sign-off`

Issue #370, opened 2022-09-02
Threshold signature check fails if any single signature is invalid


trishankatdatadog:
While this section is getting some attention: I just found that this early return will cause the threshold signature check to fail if any single signature is invalid, even if there is a threshold of valid signatures. To fix, we should replace `return ErrInvalid` with `continue`.

```go
if err := verifier.Verify(msg, sig.Signature); err != nil {
	return ErrInvalid
}
```

_Originally posted by @ethan-lowman-dd in https://github.com/theupdateframework/go-tuf/pull/369#discussion_r961943318_

joshuagl:
Per https://github.com/theupdateframework/go-tuf/pull/369#discussion_r961945970 and https://github.com/theupdateframework/go-tuf/pull/369#discussion_r962008245 we should be careful with this fix.

We must add tests, fix broken tests, and above all else communicate the impact of this change well.

znewman01:
@asraa to reroute

asraa:
cc @jku

Issue #417, opened 2022-10-14
[bug] client/testadata/Published2Times_keyrotated_invalidOldRootSignature is incorrect


asraa:
Blocks https://github.com/theupdateframework/go-tuf/issues/370

The client test: https://github.com/theupdateframework/go-tuf/blob/3890c1e7ace43d67622428187a85ba486c2528e5/client/client_test.go#L520

intended on validating that the client will fail with the old root key did not sign the new root. It currently passes on a correct error. The error is due to an invalid signature on the new root:
```
$ cat client/testdata/Published2Times_keyrotated_invalidOldRootSignature/server/metadata/2.root.json | jq
{
  "signatures": [
    {
      "keyid": "05e17c1501d627b2597322f80d33aacec6f30a507552d3326a88913422b0e30b",
      "sig": "000000"
    },
```

BUT! Critically: this was the NEW added key. The old key *did* sign the new root. And because it's also part of the new root's trusted root keys, that single signature suffices the update.

In reality, this was a VALID update, because the threshold is 1, and the old key correctly signed both the old and new root.

In order to fix, note that
1. the testcase `client/testdata/Published2Times_keyrotated_invalidNewRootSignature` is actually the one testing tampering of the old root signature. 
2. this testcase should have revoked the old key.

trishankatdatadog:
Yeap, I think @ethan-lowman-dd and I ran into this while trying to fix it during #369

PR #418, opened 2022-10-14
fix: fix verification to continue on invalid sigs

asraa:

Fixes https://github.com/theupdateframework/go-tuf/issues/370

Blocked on https://github.com/theupdateframework/go-tuf/issues/417

As described in the issue, client will currently fail on any single invalid signature rather than counting the number of valid signatures. The fix is that it continues on invalid signatures, not counting them to the threshold of valid signatures.

Fixes #<Issue>

Release Notes: 
* fix!: client will allow invalid signatures on metadata, so long as the number of valid ones reaches the threshold. Previously, the client rejected metadata signed with any invalid signatures.

**Types of changes**:


repo.go
```diff
@@ -862,26 +863,34 @@ func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signatur
...
+	// the db verification to allow a single signature to suffice.
+	// This way, we test if this new signature is valid.
+	oldSigs := s.Signatures
+	s.Signatures = []data.Signature{
+		signature,
+	}
+	for _, db := range dbs {
+		dbRole := db.GetRole(role)
+		// NOTE: We modify the database used to verify to have a threshold of 1.
+		// This way, we are testing whether this singly added new signature is valid.
```
joshuagl, line 877: This is a neat hack, but also provides a good example of they the database model is often more trouble than it's worth. I do see the value in providing a high-level API to interact with the evaluated state of the system, but we must also have the ability to operate on and evaluate individual metadata to improve clarity of the code.

This isn't a criticism of the change here, just continuing to collect evidence for a simplification effort.

joshuagl (approved):
This change LGTM, thanks Asra!

repo.go
asraa: In all honesty, I reflected on what I'd done on Friday and realized I was making it way too complicated for anyone else to read that code.

The better thing would be if I just factored out the verification logic from the DBs (the reason why I couldn't just use a verifier and go is because I'd have to dupe the canonicalization logic). So I will do that!

joshuagl: That sounds great, thanks! 
(I've moved the PR to draft for now while you work on that)

asraa: https://github.com/theupdateframework/go-tuf/pull/418/commits/8af86a0c71cb8630d40f2728398f5853556e1145

This commit implements the change: the only pitfall is that in the DB model, we need to recanonicalize the metadata at each interval.

As a second follow-up, I can consolidate all the places where we `EncodeCanonical` payload (in repo.go we do this to retrieve signed metadata files for offline signing, it sign.go we do this to sign, and in verify.go we do this to verify). This would remove re-canonicalization as part of the sign/verify flow, and put canonicalization logic closer to the `data` API. WDYT?

rdimitrov (approved):

repo.go
```diff
@@ -843,23 +843,40 @@ func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signatur
...
+
+			verifier, err := db.GetVerifier(signature.KeyID)
+			if err != nil {
+				continue
+			}
+
+			// Now check if this validly signed the metadata.
+			if err := verify.VerifySignature(s.Signed, signature.Signature,
+				verifier); err == nil {
+				validSig = true
```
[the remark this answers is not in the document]
asraa: Yes we can!

repo.go
```diff
@@ -872,16 +889,6 @@ func (r *Repo) AddOrUpdateSignature(roleFilename string, signature data.Signatur
 	signatures = append(signatures, signature)
 	s.Signatures = signatures
 
-	// Check signature on signed meta. Ignore threshold errors as this may not be fully
-	// signed.
-	for _, db := range dbs {
-		if err := db.VerifySignatures(s, role); err != nil {
-			if _, ok := err.(verify.ErrRoleThreshold); !ok {
```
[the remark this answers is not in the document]
asraa: I'm not sure if you mean why did I delete this code, or why wasn't this code originally checking for `not verify.ErrRoleThreshold`.

For the first: I moved the signature check to above (`validSig`). Critically, instead of using this code that uses the key DB model, I use a single key verifier rather than the whole db. Otherwise, we can't detect whether the sig was valid, because now we don't error out on ONE invalid sig in the db verification function.

For the second: we were checking for errors that are _not_ `verify.ErrRoleThreshold`: see `!ok`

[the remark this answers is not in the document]
trishankatdatadog: Sorry, I meant the second meaning as you defined. Yes, agreed: doesn't make sense for checking non-threshold-related errors over the whole signatures collection when we are only adding/updating one sig here.

trishankatdatadog (approved):

rdimitrov (approved):

PR #425, opened 2022-10-28
fix(data): initialise new root/targets metadata at v1 instead of v0

### LINKED ISSUE #272: cleanup: Initialize new root and targets at version 1
Currently, the API for `NewRoot()` and `NewTargets()` initialized version at 0, whereas they should generally be started at 1. 

Clean this up, and existing tests.
https://github.com/theupdateframework/go-tuf/pull/175/files#r850775514

abs007:
Fixes #272

Release Notes: @trishankatdatadog and I worked on this issue to have a new field for `NewRoot()` and `NewTargets()`. This new field stores a boolean value that gets set to True if either of the functions mentioned above are used. This makes sure that the newly created target/ root file doesn't have it's `Version` incremented as it's not staged.

**Types of changes**:

### DIFF
--- a/data/types.go
+++ b/data/types.go
@@ -124,6 +124,7 @@ func NewRoot() *Root {
 		Expires:            DefaultExpires("root"),
 		Keys:               make(map[string]*PublicKey),
 		Roles:              make(map[string]*Role),
+		Version:            1,
 		ConsistentSnapshot: true,
 	}
 }
--- a/repo.go
+++ b/repo.go
@@ -30,6 +30,8 @@ const (
 	defaultMaxDelegations = 32
 )
 
+var errUninitializedRoot = errors.New("a root must be initialized")
+
 // topLevelMetadata determines the order signatures are verified when committing.
 var topLevelMetadata = []string{
 	"root.json",
@@ -97,6 +99,13 @@ func NewRepoIndent(local LocalStore, prefix string, indent string,
 	if err != nil {
 		return nil, err
 	}
+	// Stage a new root if one isn't present
+	if _, err := r.root(); err != nil && errors.Is(err, errUninitializedRoot) {
+		if _, err = r.NewRoot(); err != nil {
+			return nil, err
+		}
+	}
+
 	return r, nil
 }
 
@@ -119,7 +128,10 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	if len(t.Targets) > 0 {
 		return ErrInitNotAllowed
 	}
-	root := data.NewRoot()
+	root, err := r.root()
+	if err != nil {
+		return err
+	}
 	root.ConsistentSnapshot = consistentSnapshot
 	// Set root version to 1 for a new root.
 	root.Version = 1
@@ -136,10 +148,19 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	return nil
 }
 
+func (r *Repo) NewRoot() (*data.Root, error) {
+	root := data.NewRoot()
+	// This will stage the metadata, which means that there won't be any further increments.
+	return root, r.setMeta("root.json", root)
+}
+
 func (r *Repo) topLevelKeysDB() (*verify.DB, error) {
 	db := verify.NewDB()
 	root, err := r.root()
 	if err != nil {
+		if errors.Is(err, errUninitializedRoot) {
+			return db, nil
+		}
 		return nil, err
 	}
 	for id, k := range root.Keys {
@@ -158,7 +179,7 @@ func (r *Repo) topLevelKeysDB() (*verify.DB, error) {
 func (r *Repo) root() (*data.Root, error) {
 	rootJSON, ok := r.meta["root.json"]
 	if !ok {
-		return data.NewRoot(), nil
+		return nil, errUninitializedRoot
 	}
 	s := &data.Signed{}
 	if err := json.Unmarshal(rootJSON, s); err != nil {
--- a/repo_test.go
+++ b/repo_test.go
@@ -719,8 +719,8 @@ func (rs *RepoSuite) TestCommit(c *C) {
 	r, err := NewRepo(local)
 	c.Assert(err, IsNil)
 
-	// commit without root.json
-	c.Assert(r.Commit(), DeepEquals, ErrMissingMetadata{"root.json"})
+	// commit without targets.json
+	c.Assert(r.Commit(), DeepEquals, ErrMissingMetadata{"targets.json"})
 
 	// Init should create targets.json, but not signed yet
 	r.Init(false)


trishankatdatadog (commented):
Looks great, thanks! A few comments:

Do we have unit tests for testing that this PR is doing the right thing for:
1.1. When NewRoot/NewTarget (v1) is the case?
1.2. When it's v1 but loaded off RAM/file?
1.3. When it's > v1?
2. That we do not accidentally write NewRoot/NewTarget to JSON.

data/types.go
```diff
@@ -115,6 +115,7 @@ type Root struct {
 	Custom      *json.RawMessage      `json:"custom,omitempty"`
 
 	ConsistentSnapshot bool `json:"consistent_snapshot"`
+	NewRoot            bool `json:"-"`
```
trishankatdatadog, line 118: Could you add a comment for why we have this variable, and why we do not serialise it to JSON?

data/types.go
```diff
@@ -226,6 +229,7 @@ type Targets struct {
 	Targets     TargetFiles      `json:"targets"`
 	Delegations *Delegations     `json:"delegations,omitempty"`
 	Custom      *json.RawMessage `json:"custom,omitempty"`
+	NewTarget   bool             `json:"-"`
```
trishankatdatadog, line 232: Maybe `NewTargets` is a more consistent name.

data/types.go
abs007: Should I add another comment for `NewTargets` too?

abs007:
From what I could understand, 

1.1: Have tests to check if NewRoot/NewTarget returns a struct with`Version`: 1
1.2: To check if the new`NewRoot` bool field doesn't cause clashes for when files are loaded in?

data/types.go
asraa: It seems like this is just meant to check if the root is at the initial. 

Maybe it's best to stage the root and targets file in the same place that `NewTargets` is called? @ethan-lowman-dd suggested this here: https://github.com/theupdateframework/go-tuf/issues/272#issuecomment-1271733490

Consider replacing all the `data.NewRoot()` function calls in `repo.go` with a new function
```
func (r *Repo) NewRoot()  error {
	root := data.NewRoot()
        // This will stage the metadata, which means that there won't be any further increments.
        return r.setMeta("root.json", root)
}
```

trishankatdatadog: You know, that's not a bad idea. Let me talk to @abs007 about this.

trishankatdatadog:
> Could you confirm if this is what you mean:
> 
> 1.1: Have tests to check if NewRoot/NewTarget returns a struct with`Version`: 1 1.2: To check if the new`NewRoot` bool field doesn't cause clashes for when files are loaded in?

Let's sync over Zoom this week?

data/types.go
abs007: > > Could you confirm if this is what you mean:
> > 1.1: Have tests to check if NewRoot/NewTarget returns a struct with`Version`: 1 1.2: To check if the new`NewRoot` bool field doesn't cause clashes for when files are loaded in?
> 
> Let's sync over Zoom this week?

Let's do that 👍

abs007:
> > Could you confirm if this is what you mean:
> > 1.1: Have tests to check if NewRoot/NewTarget returns a struct with`Version`: 1 1.2: To check if the new`NewRoot` bool field doesn't cause clashes for when files are loaded in?
> 
> Let's sync over Zoom this week?

Let's do that 👍

data/types.go
trishankatdatadog: We tried this approach just now, and unfortunately, due to a `repo.setMeta()->repo.root()` recursion, we are unable to use this trick, at least in the `repo.root()` function. I don't see a better way around this than to rework the API. WDYT?

abs007: @asraa could we get your feedback? Thanks.

asraa: Got it. The problem with the recursion is because `setMeta()` is trying to retrieve the `root()` to gather the signers. If there's no root yet, then you shouldn't have anything to sign. I actually think it SHOULDN'T fail in that case (if there's no signers). 

If you fix the recursion, the tests still fail because the tests try to do repo operations like create keys when `repo.Init` was never called.

Should `Init` always be called before trying to add root keys and targets? I think it should...

trishankatdatadog: I think so. My concern is that if the API cannot handle such a simple change (e.g., going from v0 to v1), then I'm worried about refactoring it just to fix the issue in hand now, only to find unforeseen consequences later 🙁 Personally, I would much rather redesign and rewrite the API instead of trying to fix it now, but this might be the wrong thing to do if no one has the time to work on the better API (not to mention that the kludge in this PR is hacky). What do others think? @znewman01 @mnm678 

Also, more importantly, how important is this issue to fix right now? 🙂

asraa: Yeah: agree, it's not so important to handle right now because there are ways the user won't encounter the issue.

I managed to get a patch & tests working that removed the recursion so long as a root is staged whenever a repo is initialized... I can share it with you, but yes: it involves that `NewRepo` or `Init` must stage a new root file (I think that's a safe assumption, but right now clients are probably unsure when Init is requried. My understanding is that it's required for a new repo, where `NewRepo` can be created off an existing store).

trishankatdatadog: That's great! Would you like to handover the patch here so that @abs007 and I could help push it thru?

asraa: https://gist.github.com/asraa/47a820c3c8e593416d2d67680ea48107

here it is! you have to regen the client interop code though too, but I tested it to work

Right now the least changes were done staging a new root when you call `NewRepo`, but ideally I think that belongs in `Init`. But tests usually don't call `Init`, aha

trishankatdatadog: Thanks very much! @abs007 Do you think you have all the pieces? Happy to meet this weekend on the same day/time to help.

abs007: So, I tried exploring this call stack: ` setMeta() -> signersforRole() -> getSignersInDB() -> GetSigners()` and noticed this [line](https://github.com/theupdateframework/go-tuf/blob/master/local_store.go#L150) in `local_store.go`. Aren't we supposed to be returning `nil, err` because we do expect an err in [here](https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L909) it seems.

abs007: Also, I tried exploring this call stack: ` setMeta() -> signersforRole() -> getSignersInDB() -> GetSigners()` and noticed this [line](https://github.com/theupdateframework/go-tuf/blob/master/local_store.go#L150) in `local_store.go`. Aren't we supposed to be returning `nil, err` because we do expect an err in [here](https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L909) it seems.

It seems that this might be allowing the empty db obtained from [here](https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L747)

abs007: @asraa  wdyt?

asraa: > Aren't we returning an empty db here? https://gist.github.com/asraa/47a820c3c8e593416d2d67680ea48107#file-gistfile1-txt-L67

Yes, but we're only returning the empty DB when the root is Unitialized (as in, just this first instance)

asraa: > Also, I tried exploring this call stack: setMeta() -> signersforRole() -> getSignersInDB() -> GetSigners() and noticed this [line](https://github.com/theupdateframework/go-tuf/blob/master/local_store.go#L150) in local_store.go. Aren't we supposed to be returning nil, err because we do expect an err in [here](https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L909) it seems.

I think this is returning the empty list of signers though, which is OK because we don't require any signers. See how these return nil as well https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L923 -- as in, it's not an error that we don't have any signers available.

abs007: Yeah, makes sense. That being said, why does [this](https://github.com/theupdateframework/go-tuf/blob/master/repo.go#L909) expect an error though if we'll never return one?

abs007:
@asraa could you review the changes, thanks

repo.go
```diff
@@ -119,7 +128,10 @@ func (r *Repo) Init(consistentSnapshot bool) error {
 	if len(t.Targets) > 0 {
 		return ErrInitNotAllowed
 	}
-	root := data.NewRoot()
+	root, err := r.root()
```
abs007, line 131: Was this supposed to be `r.NewRoot()`?

repo.go
```diff
@@ -97,6 +99,13 @@ func NewRepoIndent(local LocalStore, prefix string, indent string,
 	if err != nil {
 		return nil, err
 	}
+	// Stage a new root if one isn't present
+	if _, err := r.root(); err != nil && errors.Is(err, errUninitializedRoot) {
+		if _, err = r.NewRoot(); err != nil {
+			return nil, err
+		}
+	}
+
```
abs007, line 108: @asraa Could you please explain what this part of the code does and does it have anything to do with why `r.RootVersion()` (and other methods calling `r.root()`) doesn't get an error when it calls `r.root()`?

abs007:
Thank you @trishankatdatadog. Appreciate it :)

rdimitrov:
> I think, given that @rdimitrov is working on a better go-tuf, we should go ahead and close this. What does everyone think?
> 
> Even if we do so, I want to thank our @abs007 very much for working hard on this PR. Your efforts have not been wasted, and you have been very helpful to us.

Hey, @trishankatdatadog, I think that makes sense 👍 It's unfortunate that something that seemed quite straightforward to do ended up being quite the opposite. In any case, the work did by @abs007 in this PR is a valuable example for some of the issues we face with go-tuf and thus the reasoning behind [rdimitrov/go-tuf-metadata](https://github.com/rdimitrov/go-tuf-metadata). Thank you, @abs007! 💯

trishankatdatadog:
I'll wait for @asraa to also confirm/deny whether we need this fix for v0.7.

asraa:
> In any case, the work did by @abs007 in this PR is a valuable example for some of the issues we face with go-tuf and thus the reasoning behind [rdimitrov/go-tuf-metadata](https://github.com/rdimitrov/go-tuf-metadata). Thank you, @abs007! 100

I agree! @abs007 with @rdimitrov new API, it will be easier to make changes and I hope we create some new issues that we can help tackle :)

> I'll wait for @asraa to also confirm/deny whether we need this fix for v0.7.

It was largely an API cleanup but doesn't affect any external behavior, so I think it's OK to pass

trishankatdatadog:
Great, thanks everyone, especially @abs007: we hope to see you again in the near future.

abs007:
Thank you everyone and I do hope to make future changes :) .

trishankatdatadog:
> Thank you everyone and I do hope to make future changes :) .

You are a very bright, young engineer. As I have mentioned personally before, let us know if there is anything we can do to help with your career.

abs007:
Thank you @trishankatdatadog. I will keep that in mind :)

PR #431, opened 2022-11-08
feat: Tap 4 Implementation (TUFClient interface)

### LINKED ISSUE #348: Implement TAP 4: Multi-repository mappings

dibrinsofor:
See #348 

Release Notes: 

**Types of changes**:

- [X] New feature (non-breaking change which adds functionality)

**Description of the changes being introduced by the pull request**:  Began abstracting Client into interface in interface.go for future use towards multi-client mapping. currently setup with Update, Download, VerifyDigest, Target and Targets functionality. with @znewman01 and @vaikas

- [X] Tests have been added for the bug fix or new feature


dibrinsofor:
I had some issues rebasing with just these commits as I had already started to implement some of your suggestions locally, hence the new PR.

client/interface.go
```diff
@@ -0,0 +1,30 @@
...
+	// dest will be deleted and an error returned in the following situations:
+	//
+	//   - The target does not exist in the local targets.json
+	//   - Failed to fetch the chain of delegations accessible from local snapshot.json
+	//   - The target does not exist in any targets
+	//   - Metadata cannot be generated for the downloaded data
+	//   - Generated metadata does not match local metadata for the given file
+	Download(name string, dest Destination) (err error)
+
+	VerifyDigest(digest string, digestAlg string, length int64, path string) error
```
asraa, line 21: why does this need to be exposed? Is this for verifying a target against some local metadata?

asraa (approved):
other than the usage of verifydigest, lgtm

client/interface.go
znewman01: > why does this need to be exposed? 

The unhelpful answer is that it's part of the effective public interface that a Client exposes and it's covered in `client_test.go` so if we want the whole test suite to pass we need to expose it 😛 

I think specifically it's for OCI and other, similar contexts: https://github.com/theupdateframework/go-tuf/pull/170

It feels like a useful part of the public interface generally IMO. It's already part of the `Client` struct so we're somewhat committed to it. I think if we have this method in `Client`, it should be part of the `TUFClient` interface. I

You can implement this by calling `Target(path)` and then checking the hash etc. manually; that seems slightly error-prone 🤷 I'd consider removing it from the struct and the interface, can't find anybody that uses it. But I think let's make that a separate discussion.

>  Is this for verifying a target against some local metadata?

Roughly, yes. @dibrinsofor could we add a comment like the following?

> VerifyDigest verifies a target at the given path by its digest.
>
> This is useful in contexts where the targets are distributed via some channel besides TUF (e.g., OCI registries).

asraa: Got it, the linked PR was useful for context!

just for future reference testing in the same package doesn't require only using exported methods, you can test internal methods in the same package.

I do think it's useful, but I almost think my gripe was that the interface was kind of confusing. I'm supplying a hash, then alg, then size, and then path. Presumably of my target, and expect that this method will get the target the path and compare the metadata. 

I think VerifyTargetWithDigest(targetPath, digestAlg, digest, length) woulda made more sense, but we can work on it separately from this interface discusison

asraa (approved):

znewman01 (approved):

znewman01:
This looks good expect for two things:

1. You need to add the [DCO](https://developercertificate.org/) to your commits: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md
2. Can you modify the PR to remove "Fixes #348"? It should just say "See #348" or something

znewman01 (dismissed):

znewman01:
Sorry about this @dibrinsofor but can you rebase? We can do a quick re-review, should be an easy change then.

dibrinsofor:
@znewman01 all done!

znewman01 (approved):
CC @asraa for an extra maintainer approval

Issue #286, opened 2022-05-09
tests: create unit tests for verifying the protection against rollback attacks


rdimitrov:
This is a follow-up issue raised during the discussion of [No protection against rollback attacks for roles other than root](https://github.com/theupdateframework/go-tuf/security/advisories/GHSA-66x3-6cw3-v5gj). 

The goal is to update or create new test cases which verify the rollback attack protection against the different roles in TUF - root, timestamp, snapshot, targets and delegated targets.

References:
* Link to the advisory describing the reason for the implementation - https://github.com/theupdateframework/go-tuf/security/advisories/GHSA-66x3-6cw3-v5gj
* Link to the implementation - https://github.com/theupdateframework/go-tuf/commit/ed6788e710fc3093a7ecc2d078bf734c0f200d8d
* Link to the documentation describing the expected behavior - [TUF](https://theupdateframework.github.io/specification/latest/#detailed-client-workflow)

Issue #449, opened 2023-01-20
Would it be possible to cut a release 0.5.2?


vaikas:
Hey there,

Both cosign and policy-controller in Sigstore currently are using a non-cut release to pick up the airgap verification capabilities that was added in #397 

Before we cut new releases for those, it would be much better to point to a release here. So, asking if it would be possible to get a new release cut, or when the next planned release is? Last one 0.5.1 was in September.

asraa:
@trishankatdatadog @mnm678 @rdimitrov @znewman01 

Only thing that was Must Have on our project is some added testing. I think we can get a release out, honestly. The testing shouldn't take too long, I'll take a look this afternoon. https://github.com/theupdateframework/go-tuf/issues/286

trishankatdatadog:
> @trishankatdatadog @mnm678 @rdimitrov @znewman01
> 
> Only thing that was Must Have on our project is some added testing. I think we can get a release out, honestly. The testing shouldn't take too long, I'll take a look this afternoon. #286

Sounds good. Let me know, and I'm happy to help cut a release.

znewman01:
Thanks @asraa !

I'm supportive of a release. If #286 is going to take longer than a couple days I'd be happy to just push it to 0.5.3.

znewman01:
#450 is merged, I am going to cut a release now

znewman01:
Tag created: https://github.com/theupdateframework/go-tuf/releases/tag/v0.5.2

znewman01:
https://github.com/theupdateframework/go-tuf/releases/tag/v0.5.2

PR #450, opened 2023-01-23
test: add tests for rollback protection on snapshot, targets, delegations

asraa:

Fixes https://github.com/theupdateframework/go-tuf/issues/286

* client: adds testing for rollback protection on snapshot, targets, and delegated targets

Relevant: https://github.com/theupdateframework/go-tuf/issues/449

(Testing is already added for timestamp)


rdimitrov (approved):
LGTM. Testing the root role for rollback protection was already covered in previous tests, so with this we should have the full set now. Thanks, @asraa 👍

znewman01 (approved):

Issue #461, opened 2023-02-22
Please build releases for darwin_arm64


udf2457:
At present you are only building releases for darwin_amd64.

With the increasing prevalence of Apple Silicon based systems, you should really start to also build for darwin_arm64

trishankatdatadog:
Sounds like a good idea. Do you think you could help us with this, perhaps by updating our GitHub workflows? Not sure if it's possible, but needs investigation. Thanks!

udf2457:
@trishankatdatadog I'm no goreleaser expert, but I'm guessing it might be as simple as adding `arm64` to `.goreleaser/tuf.yml`  which currenltly only has `amd64` in it ?

Certainly I know on the CLI its as simple as passing the right `GOARCH` to `go build`.  So its likely just a question of where the same thing goes in a CI/CD process.

trishankatdatadog:
> @trishankatdatadog I'm no goreleaser expert, but I'm guessing it might be as simple as adding `arm64` to `.goreleaser/tuf.yml` which currenltly only has `amd64` in it ?

Great, so please help us send a PR to test it 🙂

PR #463, opened 2023-02-22
ci(build): Add arm64 to build

udf2457:
Per #461. 

Note that I'm no build process expert, but I reckon this should work.  ;-)

### DIFF
--- a/.goreleaser/tuf-client.yml
+++ b/.goreleaser/tuf-client.yml
@@ -16,6 +16,7 @@ builds:
       - windows
     goarch:
       - amd64
+      - arm64
     main: ./cmd/tuf-client/
 source:
   enabled: true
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
Seems legit, thanks! Let me try to get more reviewers...

trishankatdatadog (approved):

rdimitrov (commented):
Hey, @udf2457, thank you for your contribution 👋  

As far as cross-building and releasing the images go, I think this should be enough from the goreleaser side 👍 

It's worth mentioning though that we don't run our CI (thus our tests) on anything else other than the x86_64-based runner provided by GitHub. So until we either add a self-hosted runner or maybe emulate the aarch64 environment we won't run any tests on an aarch64 architecture.

rdimitrov (approved):
I think we can move this forward 👍

znewman01 (approved):

PR #500, opened 2023-06-02
ci: Disable daily checking for version but not security updates

trishankatdatadog:
Disable daily checking for version but not security updates.

Maybe we should also document a process for version updates when cutting a new release?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,9 @@ updates:
     commit-message:
       prefix: "chore"
       include: "scope"
-    open-pull-requests-limit: 10
+    # Disable version updates, but not security updates:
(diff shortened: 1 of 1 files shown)


znewman01 (commented):
Hm...I'm okay with this but I'm not 100% sure it actually saves time in the long run. I worry that when there's a security review, we'll have to deal with a bunch of accumulated changes.

trishankatdatadog:
I'm good with weekly. If you both agree, I can make that happen here

rdimitrov (commented):
Having it on a weekly basis sounds great for me! 👍

znewman01:
Weekly works for me too!

znewman01 (approved):

PR #509, opened 2023-06-28
fix: error handling for add. verifier and signer method signatures

### LINKED ISSUE #363: Fix: Handle errors in Key interface
The key interface does not support error handling on the following points:
`MarshalPublicKey()`, `Public()`, `PublicData()`

https://github.com/theupdateframework/go-tuf/blob/ebbc6b8d12d861335a3fc6e7fd8c69a53acaa1e6/pkg/keys/keys.go#L33

In case someone creates a key without unmarshalling first, we should handle these errors.

dibrinsofor:

Fixes #363 

Release Notes: 

**Types of changes**:

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected). **Please ensure that your PR title** is a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) breaking change (with a `!`, as in `feat!: change foo`). 

Added support for error handling for `MarshalPublicKey()`, `Public()`, `PublicData()`


.gitignore
```diff
@@ -3,3 +3,4 @@ cmd/tuf/tuf
 cmd/tuf-client/tuf-client
 .vscode
 *~
+pkg/client/
```
znewman01, line 6: Why is this here?

pkg/keys/ecdsa.go
```diff
@@ -40,15 +40,14 @@ type EcdsaVerifier struct {
 	key       *data.PublicKey
 }
 
-func (p *EcdsaVerifier) Public() string {
+func (p *EcdsaVerifier) Public() (string, error) {
```
znewman01, line 43: Leave this interface

pkg/keys/ecdsa.go
```diff
@@ -40,15 +40,14 @@ type EcdsaVerifier struct {
...
 	// This is already verified to succeed when unmarshalling a public key.
 	r, err := x509.MarshalPKIXPublicKey(p.ecdsaKey)
 	if err != nil {
-		// TODO: Gracefully handle these errors.
```
znewman01, line 47: How about:

```go
if err != nil {
    panic(err) // This is already verified to succeed when unmarshalling a public key.
}
```

pkg/keys/ed25519.go
```diff
@@ -32,8 +32,15 @@ type ed25519Verifier struct {
 	key       *data.PublicKey
 }
 
-func (e *ed25519Verifier) Public() string {
-	return string(e.PublicKey)
+func (e *ed25519Verifier) Public() (string, error) {
+	publicKey, err := json.Marshal(ed25519Verifier{
```
znewman01, line 36: I think this can be reverted.

pkg/keys/ed25519.go
```diff
@@ -43,8 +50,22 @@ func (e *ed25519Verifier) Verify(msg, sig []byte) error {
...
+	}
+
+	key := &data.PublicKey{
+		Type:       data.KeyTypeEd25519,
+		Scheme:     data.KeySchemeEd25519,
+		Algorithms: data.HashAlgorithms,
+		Value:      publicKey,
+	}
+
+	return key, nil
```
znewman01, line 68: I actually don't think there's a lot of value to re-marshalling in this method; I'd revert this change.

pkg/keys/ed25519.go
```diff
@@ -150,12 +171,18 @@ func (e *ed25519Signer) UnmarshalPrivateKey(key *data.PrivateKey) error {
 	return nil
 }
 
-func (e *ed25519Signer) PublicData() *data.PublicKey {
-	keyValBytes, _ := json.Marshal(ed25519Verifier{PublicKey: []byte(e.PrivateKey.Public().(ed25519.PublicKey))})
+func (e *ed25519Signer) PublicData() (*data.PublicKey, error) {
+	keyValBytes, err := json.Marshal(ed25519Verifier{
```
znewman01, line 175: Similarly, I'd revert this

pkg/keys/ed25519_test.go
```diff
@@ -18,6 +18,10 @@ type Ed25519Suite struct{}
 
 var _ = Suite(&Ed25519Suite{})
 
+func (Ed25519Suite) TestMarshalEd25519(c *C) {
```
znewman01, line 21: Empty test; remove

pkg/keys/rsa.go
```diff
@@ -35,15 +35,14 @@ type rsaVerifier struct {
 	key       *data.PublicKey
 }
 
-func (p *rsaVerifier) Public() string {
+func (p *rsaVerifier) Public() (string, error) {
 	// This is already verified to succeed when unmarshalling a public key.
 	r, err := x509.MarshalPKIXPublicKey(p.rsaKey)
 	if err != nil {
-		// TODO: Gracefully handle these errors.
```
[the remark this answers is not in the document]
znewman01: Maybe:

```go
if err != nil {
  panic(err) // This is already verified to succeed when unmarshalling a public key.
}
```

znewman01 (changes requested):
Thanks for this change!

Unfortunately, I'm going to ask you to put most of it back. We probably should have updated the issue (#363), sorry about that: we don't want to make breaking changes to the interface, and panics are fine in this instance. I'll add detailed instructions inline.

.gitignore
dibrinsofor: my bad. i'll take it out

pkg/keys/ecdsa.go
dibrinsofor: I'm formatting the error with `tuf: error unmarshalling key` as recommended by @asraa  in the [issue](https://github.com/theupdateframework/go-tuf/issues/363)

pkg/keys/ed25519.go
dibrinsofor: no need to remarshall or the breaking change

pkg/keys/ed25519_test.go
dibrinsofor: How could I go about testing the `MarshalPublicKey` method. I was considering something like this but I am not sure how I can create a custom verifier.

```go
func (Ed25519Suite) TestMarshalEd25519(c *C) {
	badKeyValue, err := json.Marshal(map[string]string{
		"bad_public": "mal key",
	})
	c.Assert(err, IsNil)

	badKey := &data.PublicKey{
		Type:       data.KeyTypeEd25519,
		Scheme:     data.KeySchemeEd25519,
		Algorithms: data.HashAlgorithms,
		Value:      badKeyValue,
	}

	fmt.Println(badKey)

	verifier := ed25519Verifier{
		PublicKey: data.HexBytes(badKey.Value),
		key:       badKey,
	}
	c.Assert(verifier.MarshalPublicKey(), ErrorMatches, "tuf: error unmarshalling key. *")
}
```

pkg/keys/ecdsa.go
dibrinsofor: also would it be helpful to return the faulty key alongside the error? Like, `tuf: error unmarshalling key: "hhdheu29394..." error`

znewman01: perfect! i don't think there's any need to stick the key in there, what you have is great

pkg/keys/ed25519.go
znewman01: Reopening—I think we want to keep returning just `e.PublicKey` and not the marshalled version

pkg/keys/ed25519.go
```diff
@@ -44,7 +51,14 @@ func (e *ed25519Verifier) Verify(msg, sig []byte) error {
 }
 
 func (e *ed25519Verifier) MarshalPublicKey() *data.PublicKey {
-	return e.key
```
znewman01, line 47: I think we still just want to return `e.key`

pkg/keys/ed25519.go
znewman01: Actually, nevermind—you can leave this with the new panic instead of ignoring the error.

pkg/keys/ed25519_test.go
```diff
@@ -33,6 +33,7 @@ func (Ed25519Suite) TestUnmarshalEd25519(c *C) {
 		Algorithms: data.HashAlgorithms,
 		Value:      publicKey,
 	}
+
```
znewman01, line 36: Extraneous newline, revert

pkg/keys/rsa_test.go
```diff
@@ -96,6 +99,7 @@ func (ECDSASuite) TestUnmarshalRSAPublicKey(c *C) {
 
 	signer := &rsaSigner{priv.PrivateKey}
 	goodKey := signer.PublicData()
+	c.Assert(err, IsNil)
```
znewman01, line 102: I think you can remove this check

sign/sign.go
```diff
@@ -21,7 +21,9 @@ func MakeSignatures(canonical []byte, k keys.Signer) ([]data.Signature, error) {
 		return nil, err
 	}
 
-	ids := k.PublicData().IDs()
+	publicData := k.PublicData()
```
znewman01, line 24: Leave this how it was

sign/sign.go
```diff
@@ -49,7 +51,8 @@ func Sign(s *data.Signed, k keys.Signer) error {
 	}
 	signatures := make([]data.Signature, 0, size+1)
 	for _, oldSig := range s.Signatures {
-		if !k.PublicData().ContainsID(oldSig.KeyID) {
+		publicData := k.PublicData()
```
znewman01, line 54: Leave this how it was

pkg/keys/rsa_test.go
```diff
@@ -60,7 +62,8 @@ func (RsaSuite) TestMarshalUnmarshalPublicKey(c *C) {
 	publicData := signer.PublicData()
 	pubKey, err := GetVerifier(publicData)
 	c.Assert(err, IsNil)
-	c.Assert(pubKey.MarshalPublicKey(), DeepEquals, publicData)
+	publicKey := pubKey.MarshalPublicKey()
```
znewman01, line 65: Leave this how it was

pkg/keys/rsa_test.go
```diff
@@ -49,7 +49,9 @@ func (RsaSuite) TestRSAVerifyMismatchPubKey(c *C) {
 	signerNew, err := GenerateRsaKey()
 	c.Assert(err, IsNil)
 
-	pubKey, err := GetVerifier(signerNew.PublicData())
+	publicData := signerNew.PublicData()
```
znewman01, line 52: Leave this how it was

pkg/keys/rsa.go
```diff
@@ -107,7 +112,9 @@ func (s *rsaSigner) SignMessage(message []byte) ([]byte, error) {
 }
 
 func (s *rsaSigner) ContainsID(id string) bool {
-	return s.PublicData().ContainsID(id)
+	publicData := s.PublicData()
```
znewman01, line 115: Leave this how it was

pkg/keys/ed25519_test.go
znewman01: The point of these `panic()` calls is that it shouldn't be possible to trigger this case, so it's not going to be straightforward to test :) 

I think the existing test coverage is fine

znewman01 (changes requested):
Couple minor changes but overall looking good! Thanks

pkg/keys/rsa.go
dibrinsofor: I don't remember making this change. maybe `origin/master` changed?

dibrinsofor:
> Couple minor changes but overall looking good! Thanks

@znewman01 just addressed, should be all good now.

znewman01 (approved):

mnm678:
Looks like there are a couple of formatting bugs (see the CI tests), but otherwise this looks good.

rdimitrov:
@dibrinsofor - Closing the PR since we've updated the code base of go-tuf. Feel free to reopen if you find it's useful to add something similar to the new one 👍

Issue #388, opened 2022-09-20
Command to remove a key


mnm678:
`revoke-key` revokes a key, but leaves it in the `keys/` directory. It would be nice to have a separate command that removed the key, something like `tuf remove-key role keyid`.

rabajaj0509:
I would like to work on this issue.

rdimitrov:
Thanks, @rabajaj0509. Feel free to reach out here in case you have any questions 👍

rdimitrov:
cc: @JuliannaReyes

rdimitrov:
Closing since the code base changed and the goal of the project would be to focus on being a library first and not so much on providing a CLI client.

There might be a CLI based on this, but it will live in a separate repository.

Thanks for raising this 👍

PR #524, opened 2023-07-25
feat: remove_key fix

JuliannaReyes:

Fixes #<Issue>

Release Notes: 

**Types of changes**:


rdimitrov (commented):
Perhaps it would be better to add this functionality to the existing revoke-key command via a flag instead of creating another command.

Can you share more details about your use case of needing that?

JuliannaReyes:
Under the issues section of theupdateframework github the command to remove-key said, "It would be nice to have a separate command that removed the key, something like tuf remove-key role keyed".  Based on the description I thought the goal was to create a separate command.

cmd/tuf/remove_key.go
```diff
@@ -0,0 +1,39 @@
...
+usage: tuf remove-key [--expires=<days>] <role> <id>
+
+Remove a signing key
+
+Before the key is removed the key will be first revoked
+The key will then be removed from the root metadata file and if the key is present in 
+"keys" directory it will also be removed 
+
+Options:
+  --expires=<days>   Set the root metadata file to expire <days> days from now.
```
rdimitrov, line 22: The `expires` value is not used anywhere later.

rdimitrov (changes requested):
That makes sense, thanks! I did not get it addresses https://github.com/theupdateframework/go-tuf/issues/388. 👍

cmd/tuf/remove_key.go
rdimitrov: What I meant is since you also revoke the key as part of the process it should be expected to take into account the expiration value too, if present.

JuliannaReyes:
@rdimitrov would you like to re-submit a review so that we can resolve the changes requested?

cmd/tuf/remove_key.go
JuliannaReyes: Just to make sure I understand exactly is that that code should be able to extract the --expires option value and then update the expiration value if --expires is provided right?

cmd/tuf/remove_key.go
```diff
@@ -0,0 +1,57 @@
...
+	if expires > 0 {
+		rooMetadata := repo.Root()
+		if rootMetadata != nil {
+			newExpires := time.Now().Add(time.Duration(expires) * 24 * time.Hour)
+			rootMetadata.Expires = &newExpires 
+			if err := repo.Commit(); err != nil {
+				return err
+			}
+		}
+	}
```
rdimitrov, line 48: I don't think there's need to do all of that, for example see how the revoke key command is implemented and copy that part from it - https://github.com/theupdateframework/go-tuf/blob/master/cmd/tuf/revoke_key.go 👍

JuliannaReyes: would it be better to call the revoke_key command in the remove_key command? instead of having to copy it over?

cmd/tuf/remove_key.go
rdimitrov: It should work fine, yes 👍

cmd/tuf/remove_key.go
rdimitrov: It should work fine, yes 👍

JuliannaReyes: Great! so then what i did was call the cmdRevokeKey(args, repo) in the remove_key command then you can take into consideration the expiration value if present. Hopefully that should work.

cmd/tuf/remove_key.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return err
+	}
+
+	keyPath := filepath.Join("keys", keyID)
+	if _, err := os.Stat(keyPath); err == nil {
+		if err := os.Remove(keyPath); err != nil {
+			return err
+		}
+	}
```
rdimitrov, line 40: Ah, so I realized that this part is going to be a bit more difficult than I thought. The revoke part is clear though!

Basically, keys are indeed stored in the `keys` directory, but a key is stored (encrypted or in plain text) in a <role>.json file so it's not going to be possible to remove it with the suggested code :/ 

In practice, removing a key would translate to:
- unmarshal that <role>.json file storing the key
- possibly decrypt the key
- delete the given key (note you'd have to find first which key entry corresponds to that keyid as there are no keyids there)
- on that note, it's not safe to assume that you're running the tuf cli from the same directory where those files and folders are, instead it would be better to use the dir variable storing the location of the filestore 👍 

Try running `./tuf init` and then `./tuf gen-key root' a few times (with a passphrase or just enter to skip it) so you can get a better idea of what I want to say 👍

JuliannaReyes: Okay yes, I see what you are saying and I think this will be a bit of a task for me since I believe I have to decrypt the key if it is encrypted. So it might take me some time to fully get a solution. I hope I may reach out if i have any question or I am a bit stuck

rdimitrov: Of course! Have a look at `local_store.go` where most of the encrypting/decrypting happens to get a sense of the workflow 👍

rdimitrov:
Closing this one since the code base changed and is no longer relevant. Thanks for your help @JuliannaReyes. 

The new code base should be much easier to work with so feel free to give it a go and contribute again if you find something interesting 👍

PR #533, opened 2023-07-30
fix: Set sig to Array when empty

### LINKED ISSUE #531: Set sig to Array when empty
right now the `tuf init` and a few other commands default to setting the `sig` field to `null` when it has no valid signatures. A few other commands and CLIs break when they try and read this new json as they are expecting an array, and null is unexpected. Subsequently we have to manually set the sig to `[]` when writing a script to handle TUF logic so that the next commands will parse the json correctly. Instead the CLI should always return valid JSON with an empty array, rather than a null, so that the downstream JSON parsing always passes correctly.

ChevronTango:

Fixes #531

Release Notes: missing signatures are now empty arrays rather than null

**Types of changes**:

- [ x] Bug fix (non-breaking change which fixes an issue)

The init command and others now returns an empty array rather than a null for the signatures field.


.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+# This configuration file was automatically generated by Gitpod.
+# Please adjust to your needs (see https://www.gitpod.io/docs/introduction/learn-gitpod/gitpod-yaml)
+# and commit this file to your remote git repository to share the goodness with others.
+
+# Learn more from ready-to-use templates: https://www.gitpod.io/docs/introduction/getting-started/quickstart
+
+tasks:
+  - init: python -m pip install --upgrade -r requirements-test.txt && go get && go build ./... && go test ./...
```
mnm678, line 8: Looks like this file was added by accident

ChevronTango: Not accidentaly, but I do think it's good for open source projects to support Gitpod setups to make contributions quicker and easier. I can remove it if it's not something you are interested in for this project though.

mnm678: I like anything that makes contributions easier :). Maybe move this to a separate pr so that we can discuss and make sure maintainers are aware.

ChevronTango:
@rdimitrov is there anything more I can do to help get this merged?

trishankatdatadog (approved):
LGTM

rdimitrov (approved):

PR #535, opened 2023-07-31
feat: #528 Add-Key to a role

### LINKED ISSUE #528: Support Adding External Keys
The CLI currently has a `gen-key` method and a `revoke-key` method, but not an `add-key` method. This makes it hard for users with external keys, for example KMS or otherwised managed key services, to add their keys to the `root.json`. Whilst it can be done with json manipulation, it would be much nicer if the tuf cli supported it natively to avoid mistakes or misconfiguration.

I propose the following

```
tuff add-key <role>
--keytype # eg. "rsa"
--public-value # The value of the public key
--scheme #eg "rsassa-pss-sha256"

# Returns the key-id
```

This will then add the key to the root,json in both the `keys`, and the roles fields with the correct hashed key id, and will return a key id which can then be used in subsequent signing requests.

ChevronTango:

Fixes #528 

Release Notes: Introduces the `add-key` command

**Types of changes**:

Introduces the `add-key` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,15 @@ snapshots (i.e. by passing `--consistent-snapshot=false`). If consistent
 snapshots should be generated, the repository will be implicitly
 initialized to do so when generating keys.
 
+#### `tuf add-key [--scheme=<scheme>] [--expires=<days>] [--public-key=<path>] <role>`
+
+Adds a new signing key for the given role.
+
+The root metadata file will be staged
+with the addition of the key's ID to the role's list of key IDs.
+
+The public value can be specified as a path or passed in via stdin.
+
 #### `tuf gen-key [--expires=<days>] <role>`
 
 Prompts the user for an encryption passphrase (unless the
--- a/cmd/tuf/add_key.go
+++ b/cmd/tuf/add_key.go
@@ -37,7 +37,7 @@ func cmdAddKey(args *docopt.Args, repo *tuf.Repo) error {
 		string(data.KeySchemeRSASSA_PSS_SHA256):
 		keyScheme = data.KeyScheme(t)
 	default:
-		fmt.Fprintf(os.Stderr, "tuf: key schema %s not recognised", t)
+		fmt.Fprintf(os.Stderr, "tuf: key schema %s not recognised\n", t)
 		return nil
 	}
 	f := args.String["--public-key"]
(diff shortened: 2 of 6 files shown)


cmd/tuf/add_key.go
```diff
@@ -0,0 +1,76 @@
...
+`)
+}
+
+func cmdAddKey(args *docopt.Args, repo *tuf.Repo) error {
+	role := args.String["<role>"]
+	var keyids []string
+
+	keyScheme := data.KeySchemeEd25519
+	switch t := args.String["--scheme"]; t {
+	case string(data.KeySchemeEd25519),
```
kommendorkapten, line 35: I think we should not default to ed25519 here, I think it's better UX to abort with a nice error message like `key type %s is not supported` (where the variable `t` should be used during logging).

cmd/tuf/add_key.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	} else {
+		expires = data.DefaultExpires(role)
+	}
+	keyids, err = repo.AddKeyWithSchemeAndExpires(role, expires, keyScheme, publicValue)
+	if err != nil {
+		return err
+	}
+	for _, id := range keyids {
+		fmt.Fprintf(os.Stdout, "Add key with ID %s", role, keyScheme, id)
```
kommendorkapten, line 73: You are missing the last two arguments here in the format string (can be seen by running `go vet`):
```shell
$  go vet ./...
# github.com/theupdateframework/go-tuf/cmd/tuf
cmd/tuf/add_key.go:73:3: fmt.Fprintf call needs 1 arg but has 3 args
```

README.md
```diff
@@ -49,6 +49,15 @@ snapshots (i.e. by passing `--consistent-snapshot=false`). If consistent
 snapshots should be generated, the repository will be implicitly
 initialized to do so when generating keys.
 
+#### `tuf add-key [--scheme=<scheme>] [--expires=<days>] [--public-value=<path>] <role>`
```
kommendorkapten, line 52: NIT: maybe call the parameter `--public-key` instead of `--public-value`?

kommendorkapten:
Nice work 🚀  
Left some comments that should be fixed.

ChevronTango:
> Nice work 🚀 Left some comments that should be fixed.

Brilliant! Thanks for those great spots! Just patched it now so should be ready to go! :)

kommendorkapten:
Sorry for not responding to this, been very busy. I will go over your latest changes next week 👍

kommendorkapten (approved):

rdimitrov (approved):
Thanks, @ChevronTango! 🚀

PR #538, opened 2023-08-01
feat: Add-Signature to support new formats of input

### LINKED ISSUE #532: add-signature to support more formats
When coming up with my solution for AWS KMS signing in https://github.com/theupdateframework/go-tuf/issues/525 I found the add-signature tool quite painful to use as it requires a file with a specific format. Instead it would be useful to be able to either pipe in that file as outlined in  https://github.com/theupdateframework/go-tuf/issues/529 or to instead specify the key manually.

I propose the following

```console
# normal use
tuf add-signature targets.json \
--signature <the-signature-string>
--format "base64" # to support tools like AWS that return base64 rather than hex. normally this is "hex"
--key-id <key-id>

# reading from stdin
echo <the-signature-string> | tuf add-signature targets.json \
--format "base64"
--key-id <key-id>
```

the inclusion of the format will help to support users making use of tools like AWS KMS to do their signing and will give us a bit more flexibility going forward

One thing to think about would be how to support backwards compatibility with the current setup that expects a full json file as input. For this I suggest we have `--format` default to "json" which will then cause `--key-id` to be redundant and will expect the signature to be a full valid json string as it is now. That way the existing approaches will continue to work as expected.

ChevronTango:

Fixes #532 add-signature to support alternative formats of input

Release Notes: 

**Types of changes**:

 add-signature to support alternative formats of input


mnm678 (approved):

rdimitrov (approved):
Thanks for you contribution @ChevronTango! 💯 🎉

PR #541, opened 2023-08-08
fix: fail to load deprecated ecdsa verifier

### LINKED ISSUE #527: Initializing client regression between v0.5.2 and v0.6.0
I am unsure exactly what caused this (and the issue may be in sigstore, but that doesn't seem to have changed), but calling this function:

https://github.com/sigstore/sigstore/blob/8fac32e450fa2fb263313255dc779799095f0ce5/pkg/tuf/client.go#L335-L349

Gives this error:

```
initializing tuf: unable to initialize client, local cache may be corrupt: tuf: error unmarshalling key: invalid PEM value:
```

Whereas reverting to v0.5.2 suceeds.

rdimitrov:

The following PR updates the `set_ecdsa` package to use the correct verifier for the deprecated ECDSA key type - deprecated is hex encoded, and the compliant one is PEM-formatted. 

The deprecated verifier is added for both key type formats - "ecdsa" and "ecdsa-sha2-nistp256".

Once this is merged, there will be a patch release of v0.6.1 which should include it.

Fixes #527 

Release Notes: 

**Types of changes**:


rdimitrov:
cc: @asraa @haydentherapper

znewman01 (approved):
Nice, thanks for digging into this!

haydentherapper (approved):
Thanks!

haydentherapper:
Can we get this merged and a new release cut?

pkg/deprecated/set_ecdsa/set_ecdsa.go
```diff
@@ -19,5 +19,8 @@ func init() {
 	if !ok {
 		panic(errors.New("expected to override previously loaded PEM-only ECDSA verifier"))
 	}
-	keys.VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, keys.NewDeprecatedEcdsaVerifier)
+	// store a mapping for both data.KeyTypeECDSA_SHA2_P256_OLD_FMT and data.KeyTypeECDSA_SHA2_P256
+	// in case a client is verifying using both the old non-compliant format and a newly generated root
+	keys.VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256, keys.NewDeprecatedEcdsaVerifier)         // compliant format
+	keys.VerifierMap.Store(data.KeyTypeECDSA_SHA2_P256_OLD_FMT, keys.NewDeprecatedEcdsaVerifier) // deprecated format
```
joshuagl, line 25: does/can this result in counting the key twice towards a threshold? would it be worth adding a test case?

rdimitrov: I'll let @asraa have a say on this too, but from what I see - I don't think it affects this. 

This adds an entry saying which verifier handles which key type. When adding a key, there's a 1:1 mapping between a keyid and its corresponding verifier so you cannot have more than 1 verifier per keyid (ref. [verify/db.go](https://github.com/theupdateframework/go-tuf/blob/4e4f7f3c2772bf637d0ff9c524ddda5e6dd8dc79/verify/db.go#L69)). Later when signatures are being verified there's a single take per signature when counting the threshold of verified signatures which essentially acts as another layer of verification against that (ref. [verify/verify.go](https://github.com/theupdateframework/go-tuf/blob/4e4f7f3c2772bf637d0ff9c524ddda5e6dd8dc79/verify/verify.go#L105))

asraa: Yes, I think that's correct. The only way it could be counted twice is if the root.json/delegation that describes the keys lists them twice with both `ecdsa` and `ecdsa-sha2-nistp256`: in which case, they're intending on them being counted as two distinct keys.

joshuagl: awesome, thanks both!

joshuagl (approved):
LGTM, thanks.

Issue #485, opened 2023-04-27
Decide on a deprecation strategy for migrating the code to a new Go-based TUF implementation - rdimitrov/go-tuf-metadata


rdimitrov:
The following issue is a placeholder for discussing the migration strategy of deprecating the current go-tuf code base in favor of https://github.com/rdimitrov/go-tuf-metadata.

**Reason**
The reasoning behind this is the discussions we've been having in the go-tuf community for the past 1+ years regarding the overall maintainability and usability of the current implementation. There are several less-than-optimal design decisions in go-tuf which often cause a lot of unnecessary implications both for the maintainers of the project but also for the actual users. Adding a feature or a simple bug fix (i.e. https://github.com/theupdateframework/go-tuf/issues/272) sometimes end up being surprisingly difficult to implement and review and unfortunately, that makes it much harder to attract new contributions too. 

Python-tuf was in the same situation and its community decided to do the same successfully redesigning its codebase approx. 2 years ago. The go-tuf community got inspired by this and evaluated that it would be worth to try and apply the same design principles used in python-tuf to create a new Go-based TUF implementation. 

The result of this effort is the https://github.com/rdimitrov/go-tuf-metadata project. 

Another example of this approach was developed in parallel - https://github.com/theupdateframework/tuf-js, which is a javascript-based implementation of TUF again based on the python-tuf design.

**Goal**
The end goal is to deprecate the current go-tuf codebase and replace it with the one from https://github.com/rdimitrov/go-tuf-metadata, so we continue to use the same repository location. The following are some of the things we need to not order to accomplish that:

- Decide on how to begin announcing this intention. Maybe have N releases (2 should be enough but maybe more?) where we say that users should expect this change in the future? Perhaps, once we have a rough estimation we can put a date on which the new release will happen?
- For how long should we support the old version after the migration is completed? Should it be in a separate branch?
- The go-tuf project has several things in it. On one side there's the go-tuf package, then there are client and repository-side CLI tools but also other miscellaneous things. We should assess what needs to be abstracted/kept away from the current code base (like https://github.com/theupdateframework/go-tuf/issues/476)
- rdimitrov/go-tuf-metadata should have proper test coverage and documentation
- Should we provide, and thus implement the CLI tools too using the new metadata implementation? If so, where should their code reside? Same or a separate repository?
[...]

PR #569, opened 2023-11-28
chore: fixes the CI status badge and updates the README.md file

rdimitrov:
The following PR fixes the CI status badge and updates the README.md file with a status update on the migration process.

Thanks to @HLeithner for bringing this up 💯 

Related to https://github.com/theupdateframework/go-tuf/issues/485

**Types of changes**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ a framework for securing software update systems.
 
 We recommend production users of go-tuf pin their dependency so that changes in main do not cause disruptions.
 
-For more details on the changes, and the rationale for making them, see issue https://github.com/theupdateframework/go-tuf/issues/485 and the release notes for [v0.7.0](https://github.com/theupdateframework/go-tuf/releases/tag/v0.7.0)
+For more details on the changes, and the rationale for making them, see [issue #485](https://github.com/theupdateframework/go-tuf/issues/485) and the v0.7.0 [release notes](https://github.com/theupdateframework/go-tuf/releases/tag/v0.7.0)
 
 ## Directory layout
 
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

README.md
```diff
@@ -1,10 +1,34 @@
...
+![ci](https://github.com/theupdateframework/go-tuf/actions/workflows/ci.yml/badge.svg) [![Coverage Status](https://coveralls.io/repos/github/theupdateframework/go-tuf/badge.svg)](https://coveralls.io/github/theupdateframework/go-tuf) [![PkgGoDev](https://pkg.go.dev/badge/github.com/theupdateframework/go-tuf)](https://pkg.go.dev/github.com/theupdateframework/go-tuf) [![Go Report Card](https://goreportcard.com/badge/github.com/theupdateframework/go-tuf)](https://goreportcard.com/report/github.com/theupdateframework/go-tuf)
 
 This is a Go implementation of [The Update Framework (TUF)](http://theupdateframework.com/),
 a framework for securing software update systems.
 
+## News
```
trishankatdatadog, line 8: Maybe instead of "News", we should call this "Backwards-compatibility"?

rdimitrov: hey, I thought of it more as giving users a heads up of the situation so they are not surprised by the upcoming changes. 

Would it be enough if I add an explicit note about the backwards-compatibility in the details section - i.e. the two versions will not be backward compatible so if someone wants to use the existing version, they should pin their version to v0.7.0?

joshuagl: Personally I wouldn't call a section "news" unless we plan to continuously update it.

I think it would be better to have an explicit, loud, statement that the API is in flux and folks using <= 0.7.0 must beware

```
# API instability

**Note**: go-tuf's API is in flux and backwards compatibility with version 0.7.0 and earlier will _**not**_ be maintained.

We recommend production users of go-tuf pin their dependency so that changes in main do not cause disruptions.

For more details on the changes, and the rationale for making them, see issue #foo
```

rdimitrov: Thanks for the suggestion! Updated the message 💯

README.md
```diff
@@ -1,10 +1,18 @@
...
 This is a Go implementation of [The Update Framework (TUF)](http://theupdateframework.com/),
 a framework for securing software update systems.
 
+## API instability
+
+**Note**: go-tuf's API is in flux and backwards compatibility with version 0.7.0 and earlier will _**not**_ be maintained.
+
+We recommend production users of go-tuf pin their dependency so that changes in main do not cause disruptions.
+
+For more details on the changes, and the rationale for making them, see issue https://github.com/theupdateframework/go-tuf/issues/485 and the release notes for [v0.7.0](https://github.com/theupdateframework/go-tuf/releases/tag/v0.7.0)
```
joshuagl, line 14: Nit: can we not include the full issue URL?

joshuagl (dismissed):
One minor request, but otherwise LGTM. Thanks.

README.md
trishankatdatadog: Yes, let's link to the words "issue" and "release notes" please, thanks

kommendorkapten (approved):
Thanks @rdimitrov

trishankatdatadog (approved):

PR #600, opened 2024-02-02
remove dependency on golang.org/x/exp

mikedanese:
Some is in the stdlib now, other parts are trivial to implement without the dep.


rdimitrov (approved):
@mikedanese - Awesome! 🚀

PR #601, opened 2024-02-02
move testutils under an ./internal/ directory

mikedanese:
This prevents these libraries from being depended on from outside the module.


rdimitrov:
@mikedanese - there's some things to fix so we can satisfy the CI but I agree it's better if these packages are internal 👍 Thanks for addressing this 🙏

mikedanese:
Oops, fixed.

rdimitrov (dismissed):

codysoyland:
FWIW, I was looking at using the RepositorySimulator in testutils as part of sigstore-go's test suite, but I can understand why it's probably best to keep it internal.

rdimitrov:
@codysoyland - oh, I wasn't aware of that, thanks 👍  Given this I think it's okay to leave it as a standalone package. cc: @mikedanese

codysoyland:
> **edit:** @codysoyland - hey, I just saw you mentioned in the PR that eventually you came up with another workaround for the tests. I don't think there's anyone else trying to use the testutils package so far, but I'm supportive of leaving it visible if it's going to help sigstore-go with the testing.
> 
> Of course, at the end we can always revisit and move it out of internal if someone comes up with a valid use case for it 👍

@rdimitrov Thanks for offering to keep it visible, but I decided late yesterday that it makes sense for us to have our own stripped down version of a repo simulator and not depend on the one here, since it may diverge and become incompatible, so I'm supportive of going ahead and merging this one!

codysoyland (approved):

rdimitrov:
@codysoyland - Thanks! 😃👍 

@mikedanese - There's a few conflicts you have to fix, but otherwise we should be good to go then 👍

mikedanese:
Rebased.

rdimitrov (approved):
The failing CI job is not related to the changes in this PR (tries to run go 1.22 and apparently something is not supporting it yet).

PR #603, opened 2024-02-06
Don't rename unless file is in same dir

jonnystoten:
This was originally done for windows only, but renaming can cause issues regardless of OS if the target dir is on another device.


kipz (approved):
Hard to test this one on GHA. Works for me at least.

rdimitrov (dismissed):
Thanks for addressing this! 🚀

rdimitrov:
> Hard to test this one on GHA. Works for me at least.

Thank you for testing this! 💯 

My assumption is the examples that we run now both on windows and linux should be enough to catch potential errors of this, i.e. a temp file failed to be renamed. On that note, I'll go ahead and enable the macos environment too 👍

rdimitrov:
Here's the PR enabling the runner environments - https://github.com/theupdateframework/go-tuf/pull/604. My assumption was we enabled the windows one for all, but it was just for the unit tests. Now it's going to run the examples too 👍

rdimitrov (approved):
When we enabled the windows runner it manifested some errors which is not surprising since we haven't tested go-tuf-metadata at all on Windows. I've opened an issue about it if someone is interested in addressing this - https://github.com/theupdateframework/go-tuf/issues/605.

I'm okay to accept this PR since it doesn't seem related to the issue I mention 👍

jonnystoten:
Hey thanks @rdimitrov ! I actually don't have a windows machine either but I could have a look using the CI to test, the failure I can see looks pretty straightforward.

PR #606, opened 2024-02-07
Do not run CI for all Go versions and use caching

rdimitrov:
The following PR:
* Removes the tests we were running against all available Go versions above the one we are using in the go.mod file. This sounded useful, but it caused more CI failures than it helps.
* Sets the cache: true for most go actions

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f
         with:
           go-version-file: 'go.mod'
-          cache: false
+          cache: true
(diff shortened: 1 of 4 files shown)


kommendorkapten (approved):
Looks good, is the windows tests expected to fail?

rdimitrov:
> Looks good, is the windows tests expected to fail?

@kommendorkapten - Yes, they were disabled before, but once we enabled them we saw there're a few issues when targets use the prefixed format we use in the tuf-demo repo. 

I've opened an issue, but from what I see it's probably caused by the wrong handling of back/forward slashes when we are translating a target name to a local path name when we create or look for the target file locally. 

I don't have a windows machine to properly debug and fix this, so I thought it's better to leave the tests enabled and failing until we get some help on this. Also this way we are open in sharing windows support is not working properly yet.

Issue #605, opened 2024-02-06
Revisit path handling for proper Windows support


rdimitrov:
This issue is about revisiting the way paths are handled within the project and update them accordingly so they work properly on non-Unix environments, such as Windows.

There are places where we construct paths out of one or more strings (even URLs) and these seem to cause failures on Windows because of the improper path separator.
 
Running our set of examples on `windows-latest` seems to help reproduce this issue, for reference - https://github.com/theupdateframework/go-tuf/pull/604

rdimitrov:
FYI @jonnystoten

rdimitrov:
I've reopened it until we fixed the multi repo issue too 👍

PR #612, opened 2024-02-23
Always use forward slash when splitting target names

kommendorkapten:
Even when running on Windows, the targets in the root uses forward slash.

Closes https://github.com/theupdateframework/go-tuf/issues/605


rdimitrov (approved):
Ah, nice catch! 🚀 I was expecting the issue to be in the library.

PR #623, opened 2024-03-08
feat: automatically comment on PR for govulncheck

MDr164:


.github/workflows/linting.yml
```diff
@@ -28,6 +30,14 @@ jobs:
         with:
           go-version-file: 'go.mod'
           go-package: ./...
+      - name: Comment on PR
+        if: steps.govulncheck.outcome == 'failure'
+        uses: thollander/actions-comment-pull-request@fabd468d3a1a0b97feee5f6b9e499eab0dd903f6 # v2.5.0
```
rdimitrov, line 35: Is it difficult if we use the gh api instead of another action?

MDr164: There is an approach that uses curl and the API and one that uses the script action to allow calling gh-action functions. I just briefly looked into it though. I suspect they'd also need write rights in order to post the comment.

rdimitrov: I was actually referring to the `gh` client as it's already available in the action but I haven't played around with it that much to tell if it will actually work properly for our use case (comment once -> edit afterwards)

MDr164: Looks like that might also be feasible but I actually discovered an even better way using a rather underdocumented feature: The env var `$GITHUB_STEP_SUMMARY` which you can just echo into and it displays this as a comment on the discussions tab. I'll try this out and update this draft.

MDr164: Nevermind, it displays on the summary and not discussion tab. I added a way that uses the gh client as you noted now.

MDr164:
It would still require write permissions to actually allow commenting. That's why most jobs don't run right now.

rdimitrov:
Looking good 👍 Is there anything else left to do or we can merge it?

MDr164:
> Looking good 👍 Is there anything else left to do or we can merge it?

Please don't merge it yet. Code-wise there is nothing needed to do but as of right now it will disable the majority of the workflows due to a violation of the permissions. The whole org forbids workflows right now that have write permissions and given that we try to claim write rights on that one job it disables all associated other workflows.

I think I'm not in the position to decide a change as it will affect every repo in the org...

MDr164:
I'm revisiting this PR since govulncheck 1.1.x is out but it does not seem like there is a fix for for the false positive reporting for the Go stdlib issue it will mark each run creating a lot of noise. It also doesn't seem like this behavior is configurable. Therefore I'm closing this PR given that elevating the rights for actions does not seem like a good idea and having a red-cross on every run also doesn't seem ideal.

PR #625, opened 2024-03-09
Use the correct verifier for RSA PSS scheme keys

rdimitrov:
The following PR fixes an issue with go-tuf's support for RSA key types.

**Details:**

Apparently for RSA key types `signature.LoadVerifier()` is defaulting to returning a verifier that is `RSAPKCS1v15Verifier` which corresponds to the RSA PKCS#1 scheme. 

Based on [the latest TUF spec](https://theupdateframework.github.io/specification/latest/#file-formats-keys), TUF uses RSA PSS instead of PKCS#1. This manifests itself by causing go-tuf v2 TUF clients to fail metadata verification in case of RSA key types.

The fix for this is to invoke the proper verifier for the PSS scheme. This means that when we are verifying a metadata signed by an RSA key type, we should use `LoadRSAPSSVerifier()` instead of the generic `LoadVerifier()`.

**References:**

* https://theupdateframework.github.io/specification/latest/#file-formats-keys
* https://github.com/sigstore/sigstore/blob/a153736bf87443389feee6aa787e97b81839175d/pkg/signature/verifier.go#L41

**Credits:**

Thanks to @kairoaraujo for reaching out and helping track this together! 🚀 Cheers! 🍻 

**What's left:**
- [ ] Update the tests

### DIFF
--- a/internal/testutils/signer/signer.go
+++ b/internal/testutils/signer/signer.go
@@ -1,0 +1,203 @@
+// Copyright 2024 The Update Framework Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"crypto"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/theupdateframework/go-tuf/v2/metadata"
+)
+
+type tufSigner interface {
+	Sign(s signature.Signer) (*metadata.Signature, error)
+}
+
+/*
+   Run this to sign test data, like this:
+   ~/git/go-tuf/internal/testutils $ go run \
+       signer/signer.go \
+       -k repository_data/keystore/timestamp_key \
+       -s rsassa-pss-sha256 \
+       -f repository_data/repository/metadata/timestamp.json
+*/
+
+func main() {
+	var scheme = flag.String("s", "", "set scheme to use for key")
+	var key = flag.String("k", "", "key file to load")
+	var f = flag.String("f", "", "file to sign")
+
+	flag.Parse()
+
+	if *scheme == "" {
+		fmt.Println("no scheme is set")
+		os.Exit(1)
+	}
+	if *key == "" {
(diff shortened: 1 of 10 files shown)


kairoaraujo:
I've tested it, and it solves the issue 😃

metadata/metadata.go
```diff
@@ -315,7 +316,21 @@ func (meta *Metadata[T]) VerifyDelegate(delegatedRole string, delegatedMetadata
...
+		// Note we should support RSA PSS, not RSA PKCS1v15 (which is what LoadVerifier would return)
+		// Reference: https://theupdateframework.github.io/specification/latest/#file-formats-keys
+		var verifier signature.Verifier
+		if key.Type == KeyTypeRSASSA_PSS_SHA256 {
+			// Load a verifier for rsa
+			publicKeyRSAPSS, ok := publicKey.(*rsa.PublicKey)
+			if !ok {
+				return &ErrType{Msg: "failed to convert public key to RSA PSS key"}
+			}
+			verifier, err = signature.LoadRSAPSSVerifier(publicKeyRSAPSS, hash, &rsa.PSSOptions{Hash: crypto.SHA256})
```
haydentherapper, line 329: FYI, if you're using the Verifier interface, you can also load a verifier with a PSS option now - https://github.com/sigstore/sigstore/blob/main/pkg/signature/signerverifier_test.go#L30-L38

jku:
> Based on the latest TUF spec, TUF uses RSA PSS instead of PKCS#1

Note that these are just examples essentially, definitely not an exhaustive list of the only schemes that can be supported: I would avoid removing support for the pkcs1v15 scheme if possible as removing optional features can be a real head ache for users...

jku:
Just as an idea... I would love a keytype support test in tuf-conformance: https://github.com/theupdateframework/tuf-conformance/issues/159 -- implementing that would give you keytype test coverage "for free" in go-tuf and should not be more than 20 lines of code.

kommendorkapten:
A lot of the tests are working, I'll hunt down the rest failures tomorrow I hope

kommendorkapten:
This is now ready for review, the tests were failing as they were signed using PKCS#1 v1.5 but the key scheme said `rsa-pss-sha256`, I resigned the test data (with the helper tool added to this pr), and then made sure any signing and verification in the tests uses PSS. And when comparing to/from bytes I ignore whitespaces as its brittle to care about them.

internal/testutils/signer/signer.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/theupdateframework/go-tuf/v2/metadata"
+)
+
+type tufSigner interface {
+	Sign(s signature.Signer) (*metadata.Signature, error)
+}
+
+/*
+   Run this to sign test data, like this:
```
rdimitrov, line 41: Thanks for this! 🚀

internal/testutils/signer/signer.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func loadSigner(k, s string) (signature.Signer, error) {
+	var pk any
+	var err error
+	var opts []signature.LoadOption
+	var rawKey []byte
+
+	switch s {
+	case metadata.KeySchemeRSASSA_PSS_SHA256:
```
kommendorkapten, line 147: I didn't bother to add support for more schemes right now, to do so is easy but not needed _now_.

mnm678 (approved):

rdimitrov:
Thank you all for reviewing and contributing to this! 🙌 💯

Issue #628, opened 2024-04-01
bug: hash algorithm defaults to SHA256 for most algorithms


mrjoelkamp:
I am using [tuf-on-ci](https://github.com/theupdateframework/tuf-on-ci) and running into issues when the [key scheme is ecdsa-sha2-nistp384](https://github.com/theupdateframework/tuf-on-ci/blob/a66f063e23739404499e1ab54677a7be69856be6/repo/test/test_repo2/root.json#L18) (worth noting that these tests aren't failing when using the python tuf client)

I've root caused the issue to the hash algorithm basically hardcoded to use SHA256 here:
https://github.com/theupdateframework/go-tuf/blob/48216cfc42444c3eaa974718091986a08c8c9f44/metadata/metadata.go#L312-L316

It looks like mainly the 256-bit curves are supported here but with a slight change this would also support ECDSA P384.

I've modified this code to the following locally and clear the error using `ecdsa-sha2-nistp384` as the key scheme:
```go
// use corresponding hash function for key type
hash := crypto.Hash(0)
if key.Type != KeyTypeEd25519 {
	switch key.Scheme {
	case KeySchemeECDSA_SHA2_P256:
		hash = crypto.SHA256
	case KeySchemeECDSA_SHA2_P384:
		hash = crypto.SHA384
	default:
		hash = crypto.SHA256
	}
}
```

PR #629, opened 2024-04-01
fix: use SHA384 for ECDSA P384

mrjoelkamp:
Handle hash algorithm selection for `ecdsa-sha2-nistp384`

fixes https://github.com/theupdateframework/go-tuf/issues/628


kommendorkapten (approved):
Nice, thank you!

rdimitrov (approved):
IIRC this was done this way as technically the spec did not list any other key schemes, but in any case it's good to support more 👍 

Thanks! @mrjoelkamp 🚀

Issue #631, opened 2024-04-10
bug: deep target file paths for consistent snapshots


mrjoelkamp:
## Issue
Deep target file paths, or file paths that are more than one subdirectory deep (e.g. `<dir1>/<dir2>/file`) fail to download due to the `targetRemotePath` being built wrong when using consistent snapshots.

## Root Cause
Current state of the code uses `strings.Cut()` to parse the directory name from the file name. This works for target files that are only one directory deep (`<subdir>/file.ext`) because `strings.Cut()` splits the string on the _first_ occurrence of the separator. 

https://github.com/theupdateframework/go-tuf/blob/b2e024ad4752cc0c4a4e376460b21deb79e40ded/metadata/updater/updater.go#L237-L244

For nested target file paths (`<subdir1>/<subdir2>/file.ext`) the `targetRemotePath` ends up being generated incorrectly since the first occurrence of `/` does not separate the directory from the file name.

When using consistent snapshots, this ends up making the path be `<subdir1>/<hash>.<subdir2>/file.ext` instead of `<subdir1>/<subdir2>/<hash>.file.ext`

PR #632, opened 2024-04-10
fix: deep targets file path

mrjoelkamp:
## Summary
* fixes bug in `DownloadTarget` for target files that are in subdirectories that are more than one level deep

## Fixes
fixes https://github.com/theupdateframework/go-tuf/issues/631


metadata/updater/updater.go
```diff
@@ -234,10 +234,11 @@ func (update *Updater) DownloadTarget(targetFile *metadata.TargetFiles, filePath
 			hashes = hex.EncodeToString(v)
 			break
 		}
-		dirName, baseName, ok := strings.Cut(targetFilePath, "/")
+		baseName := filepath.Base(targetFilePath)
```
mrjoelkamp, line 237: This probably won't work on windows, since `targetFilePath` always uses `/`

mrjoelkamp: never mind, tested it on windows and it works 😄

metadata/updater/updater.go
```diff
@@ -234,10 +234,11 @@ func (update *Updater) DownloadTarget(targetFile *metadata.TargetFiles, filePath
...
+		dirName, ok := strings.CutSuffix(targetFilePath, "/"+baseName)
```
MDr164, line 238: I know it's just a minor difference in this case but when it comes to file paths I always prefer things like `filepath.Join` instead of direct string concatenation. It's also more OS independent usually.

mrjoelkamp: I completely agree! Especially for windows builds. However, in this case the `targetFilePath` comes from metadata and I believe that this is standardized to use `/` only and if you use `filepath.Join()` on windows you end up generating an incorrect `targetRemotePath`. 

This was recently fixed in https://github.com/kommendorkapten/go-tuf/commit/484e95dc84e3daf7ee6a0933e159a257126670d0 to always use `/` for remote paths.

MDr164: Ah I see, then we should be good here.

MDr164 (approved):

trishankatdatadog:
Thanks @MDr164.

@mrjoelkamp Would you mind terribly much rebasing against base, please?

PR #633, opened 2024-04-17
Remove nil error from being printed in "persist metadata" error message

malancas:
When go-tuf is unable to persist the TUF metadata, I receive the error "failed to persist metadata: %!w(\<nil\>)". Because the code already checks if the error is not nil and returns it if not on [line 651](https://github.com/theupdateframework/go-tuf/blob/b2e024ad4752cc0c4a4e376460b21deb79e40ded/metadata/updater/updater.go#L651), the "failed to persist" error message will always try to print a nil error.

### DIFF
--- a/metadata/updater/updater.go
+++ b/metadata/updater/updater.go
@@ -652,7 +652,7 @@ func (update *Updater) persistMetadata(roleName string, data []byte) error {
 		return err
 	}
 	if string(read) != string(data) {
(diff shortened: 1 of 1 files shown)


MDr164 (approved):
I'm currently preparing a really big PR that takes care of error checking in general to avoid things like this from happening but good thing you spotted this one here early.

PR #635, opened 2024-05-03
feat: add missing CODEOWNERS and MAINTAINERS file

MDr164:


trishankatdatadog (approved):

MAINTAINERS
```diff
@@ -0,0 +1,11 @@
+# Maintainers should review changes promptly.
+# Github will auto-assign the below maintainers to a pull request.
+
+"**":
+    - rdimitrov
+    - kommendorkapten
+    - mnm678
+    - znewman01
```
kommendorkapten, line 8: I think @znewman01 left as a maintainer?

trishankatdatadog: Good point. Zack should say something if that is not the case, but he is still also in the @theupdateframework/go-tuf-maintainers list.

MDr164: Right, I just copied who was in the maintainers list in the org. If that info was outdated we should booth change it there and in this file.

kommendorkapten: Yeah that makes sense, we should use that list as authoritative.

PR #636, opened 2024-05-21
Update MAINTAINERS

trishankatdatadog:


MDr164 (approved):

MDr164:
Good to see the autoassign works, thanks Trishank for taking care of the update!

trishankatdatadog:
@JustinCappos can you give @theupdateframework/go-tuf-maintainers admin access to the team itself?

MDr164 (approved):

trishankatdatadog:
> Good to see the autoassign works, thanks Trishank for taking care of the update!

Sorry, last change, I swear

trishankatdatadog:
> Githubs CI also seems to be down so technically we should also wait for that to pass even though none of the checks except DCO are mandatory.

There's also that too much bold heading in the Markdown, but I'll let someone else fix that now if they like...

mnm678 (approved):

MDr164 (approved):
I'm fine with the markdown as is

PR #638, opened 2024-06-11
fix: configurable temp file directory

### LINKED ISSUE #639: temporary file creation for persisted metadata
## Problem
A go-tuf client is interrupted or errors when writing remote metadata to disk during `persistMetadata` in `updater.go` or a multi-client system shares `update.cfg.LocalMetadataDir` where metadata files are persisted and a client is writing persisted metadata while simultaneously another client reads it.

The risk of a client reading partially written files is low for single client systems, since the file writing error handler removes partially written files, but it can certainly still occur.

Incomplete, corrupted, or invalid persisted metadata can also occur via other means (disk read error, malicious code, bad actor)  

## Existing solution
Currently, the `persistMetadata` function writes to a temporary file in the current working directory (CWD). Once the file write is complete, then the temp file is either moved to the configured `LocalMetadataDir` or renamed if `LocalMetadataDir == CWD`. 

This mechanism should prevent empty or partially written metadata files from ever existing in the persisted metadata directory. Although, the currently implemented `moveFile` uses `io.Copy` that could also be interrupted leading to partially written or empty files.

### Use of `io.Copy`
`Copy` copies from src to dst until either EOF is reached on src or an error occurs. It returns the number of bytes copied and the first error encountered while copying, if any.

This could lead to partially written or empty files when copying the temp file to the destination, not providing an atomic write.

### Use of current working directory
Writing to the current working directory is problematic for containerized workloads (as described in https://github.com/theupdateframework/go-tuf/pull/638).

This should be an easy fix to either add a new configuration parameter for a temp file location or just use the `LocalMetadataDir` to store temporary files (`tuf_tmp`). But I ran into problems...

### Windows
I attempted to use `update.cfg.LocalMetadataDir` for the `tuf_tmp` file location but this [produced an error in windows tests](https://github.com/theupdateframework/go-tuf/actions/runs/9470099707/job/26090235976#step:5:14) (and [revealed a bug](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L583-L586)) here due to `os.Rename` access denied on windows.

There is a [thread](https://github.com/golang/go/issues/22397#issuecomment-497874091) in https://github.com/google/renameio that goes into some detail on how atomic writes for windows is problematic.

It appears that an atomic write for windows is mostly not feasible, therefore the `persistMetadata` function would need to write directly to the target file for windows. Otherwise, the code will always [write a temp file and then read the temp file and write the real file](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L588-L611) (no rename/move) in windows, which is really inefficient and no benefit.

## Do we need atomic writes for persisted metadata?
After reviewing the code base and the caveats made for windows OS, I questioned whether or not I should add additional logic to work around the [windows bug discovered in the first attempt](https://github.com/theupdateframework/go-tuf/actions/runs/9470099707/job/26090235976#step:5:14) or eliminate the persisted metadata atomic write complexity completely.

Let's explore the case that there actually is a partially written or empty persisted metadata file:
- In order for this to occur the program must have been interrupted during file write, error handling, two or more clients are sharing the same metadata and running simultaneous updates
- A client attempts [online refresh](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L125-L143) with one or more partially written metadata files
  - the [updater loads the partial role metadata and attempts to verify it](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L353-L359) and it fails
  - the updater [downloads the remote metadata](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L375) due to invalid (or missing) local data
  -  the updater persists the remote copy of the metadata (overwriting the partial copy)
- (non-standard) A client uses [unsafeLocalRefresh](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L150) to load cached metadata
  - the [partial metadata fails verification ](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/trustedmetadata/trustedmetadata.go#L209) and the client returns an error   

### 2. Remove code complexity via windows workaround and rely on metadata integrity checks within go-tuf `updater` and the TUF framework to mitigate partially written or empty persisted metadata files

~~I am in favor of 2 but willing to refactor the PR to implement 1 if desired.~~
*edit: implemented option 1 in https://github.com/theupdateframework/go-tuf/pull/638*

mrjoelkamp:
## Summary
Using the current working directory (cwd) for temporary file storage has been a bit problematic for container based workflows and various operating systems.

This is especially true for k8s workloads that have `readOnlyRootFilesystem` set, which is a security best practice. The creation of the temporary file in the working directory is not permitted with a read-only root filesystem.

## Changes
* use `update.cfg.LocalMetadataDir` for temporary file storage, so that it can be configured to a writable volume.
* remove `moveFile` as it is no longer needed 
  * temp file and destination file are in the same directory
* fixes open file handle in `repository_similator_setup.go` leading to `os.Rename` error on windows

### fixes 
#639


mrjoelkamp:
> I can see where you're coming from, but before we remove it, do we know _why_ it is there in the first place? We wouldn't want to accidentally run into the risk of removing a [Chesterton fence](https://fs.blog/chestertons-fence/)...

@trishankatdatadog it appears that the temp file and rename logic is here to produce an atomic file write so that there is never a partially written or 0-byte file for persisted metadata. Although, the go-tuf `updater` logic and TUF framework is robust enough to recover from incomplete/corrupt persisted metadata.

## Problem
A go-tuf client is interrupted or errors when writing remote metadata to disk during `persistMetadata` in `updater.go` or a multi-client system shares `update.cfg.LocalMetadataDir` where metadata files are persisted and a client is writing persisted metadata while simultaneously another client reads it.

The risk of a client reading partially written files is low for single client systems, since the file writing error handler removes partially written files, but it can certainly still occur.

Incomplete, corrupted, or invalid persisted metadata can also occur via other means (disk read error, malicious code, bad actor)  

## Existing solution
Currently, the `persistMetadata` function writes to a temporary file in the current working directory (CWD). Once the file write is complete, then the temp file is either moved to the configured `LocalMetadataDir` or renamed if `LocalMetadataDir == CWD`. 

This mechanism prevents empty or partially written metadata files from ever existing in the persisted metadata directory. 

For windows OS, the file is read back into memory and written back to disk. Essentially adding an additional disk read/write without any added benefit.

## Issues
### Use of current working directory
Writing to the current working directory is problematic for containerized workloads (as described in the PR summary).

This should be an easy fix to either add a new configuration parameter for a temp file location or just use the `LocalMetadataDir` to store temporary files (`tuf_tmp`). But I ran into problems...

### Windows
I attempted to use `update.cfg.LocalMetadataDir` for the `tuf_tmp` file location but this [produced an error in windows tests](https://github.com/theupdateframework/go-tuf/actions/runs/9470099707/job/26090235976#step:5:14) (and [revealed a bug](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L583-L586)) here due to `os.Rename` access denied on windows.

There is a [thread](https://github.com/golang/go/issues/22397#issuecomment-497874091) in https://github.com/google/renameio that goes into some detail on how atomic writes for windows is problematic.

It appears that an atomic write for windows is mostly not feasible, therefore the `persistMetadata` function would need to write directly to the target file for windows. Otherwise, the code will always [write a temp file and then read the temp file and write the real file](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L588-L611) (no rename/move) in windows, which is really inefficient and no benefit.

## Do we need atomic writes for persisted metadata?
After reviewing the code base and the caveats made for windows OS, I questioned whether or not I should add additional logic to work around the [windows bug discovered in the first attempt](https://github.com/theupdateframework/go-tuf/actions/runs/9470099707/job/26090235976#step:5:14) or eliminate the persisted metadata atomic write complexity completely.

Let's explore the case that there actually is a partially written or empty persisted metadata file:
- In order for this to occur the program must have been interrupted during file write, error handling, two or more clients are sharing the same metadata and running simultaneous updates
- A client attempts [online refresh](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L125-L143) with one or more partially written metadata files
  - the [updater loads the partial role metadata and attempts to verify it](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L353-L359) and it fails
  - the updater [downloads the remote metadata](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L375) due to invalid (or missing) local data
  -  the updater persists the remote copy of the metadata (overwriting the partial copy)
- (non-standard) A client uses [unsafeLocalRefresh](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/updater/updater.go#L150) to load cached metadata
  - the [partial metadata fails verification ](https://github.com/theupdateframework/go-tuf/blob/626088b2ddcc65f9104ccd8c2bc4c9c8bff0cf0f/metadata/trustedmetadata/trustedmetadata.go#L209) and the client returns an error   

## Solutions

### 1. Refactor `persistMetadata` to add additional logic to the temp file rename/move to only provide atomic writes (via temp file) for POSIX systems
### 2. Remove code complexity via windows workaround and rely on metadata integrity checks within go-tuf `updater` and the TUF framework to mitigate partially written or empty persisted metadata files

I am in favor of 2 but willing to refactor the PR to implement 1 if desired.

trishankatdatadog:
> I am in favor of 2 but willing to refactor the PR to implement 1 if desired.

_Excellent_ summary, thanks very much, @mrjoelkamp! I suspect you are right that this is a reliability but not a security issue, and so I don't mind Option 2.

@JustinCappos do you see any security issues? If not, we should be able to simplify the code.

JustinCappos:
I'm more concerned about a file being verified, being overwritten, and then being re-read with the assumption it was verified.  Could this occur?

mrjoelkamp:
> _Excellent_ summary, thanks very much, @mrjoelkamp! I suspect you are right that this is a reliability but not a security issue, and so I don't mind Option 2.

I just looked to python-tuf for consistency between the two clients, which leads me to think that option 1 is probably best to keep them the same, see https://github.com/theupdateframework/go-tuf/issues/639#issuecomment-2163948854

mrjoelkamp:
> I'm more concerned about a file being verified, being overwritten, and then being re-read with the assumption it was verified. Could this occur?

No, whenever the metadata files are read from persisted storage in `updater` they are verified against the root to be "trusted" and the trusted metadata is kept in volatile memory.

JustinCappos:
> > I'm more concerned about a file being verified, being overwritten, and then being re-read with the assumption it was verified. Could this occur?
> 
> No, whenever the metadata files are read from persisted storage in `updater` they are verified against the root to be "trusted" and the trusted metadata is kept in volatile memory.

Okay, and one other related question.  Could one possibly use this to overwrite old, cached files in some way?  (I don't see how, but wanted to ask.)  

If not then I don't have any concerns...

metadata/updater/updater.go
```diff
@@ -620,12 +588,8 @@ func (update *Updater) persistMetadata(roleName string, data []byte) error {
 	}
 	// caching enabled, proceed with persisting the metadata locally
 	fileName := filepath.Join(update.cfg.LocalMetadataDir, fmt.Sprintf("%s.json", url.QueryEscape(roleName)))
-	cwd, err := os.Getwd()
-	if err != nil {
-		return err
-	}
 	// create a temporary file
-	file, err := os.CreateTemp(cwd, "tuf_tmp")
+	file, err := os.CreateTemp(update.cfg.LocalMetadataDir, "tuf_tmp")
```
trishankatdatadog, line 592: Just to be on the safe side, could we please use a new, random, temp filename every time?

whalelines: That is how `os.CreateTemp` works: https://pkg.go.dev/os#CreateTemp

mrjoelkamp:
> Okay, and one other related question. Could one possibly use this to overwrite old, cached files in some way? (I don't see how, but wanted to ask.)
> 
> If not then I don't have any concerns...

`persistMetadata` is only called here when the roles are updated from the remote server. From what I can tell, the role metadata has to pass verification before `persistMetadata` is called. No changes to the behavior here for that and it matches the `python-tuf` client implementation as well.

metadata/updater/updater.go
mrjoelkamp: yep, as an example [a previous test run](https://github.com/theupdateframework/go-tuf/actions/runs/9470099707/job/26090235976#step:5:14) generated `tuf_tmp1013966631` as the temp file name

trishankatdatadog: My bad, should have consulted the docs. (Knew this functionality almost certainly must exist in Go, but not exactly where.) Thanks!

trishankatdatadog:
This looks good to me. Do you think we need any regression test at least for this behaviour?

mrjoelkamp:
> This looks good to me. Do you think we need any regression test at least for this behaviour?

Before these changes I'd say yes, but now that this removes `moveFile` and sets the temp dir to the same as the local metadata dir we have test coverage in the existing tests.

trishankatdatadog:
> Before these changes I'd say yes, but now that this removes `moveFile` and sets the temp dir to the same as the local metadata dir we have test coverage in the existing tests.

Got it. I'll let @kommendorkapten or @rdimitrov approve this PR since they have more SITG than I do here 🙂

rdimitrov (approved):
Looks good! 🚀  

I have to check the code again but for future work I'm thinking that we can get this a step further and use memfs so we don't rely on having any disk permissions at all. 

In any case thank you for addressing this @mrjoelkamp 🙏

kommendorkapten:
Just to clarify, this already exists today: https://github.com/theupdateframework/go-tuf/blob/master/metadata/config/config.go#L42

`DisableLocalCache` allows for operating on a read only filesystem.

rdimitrov:
Funny that I added it and then forgot about its existence 😃

mrjoelkamp:
> Funny that I added it and then forgot about its existence 😃

😄 no problem! I did see that option, however we do need cache enabled for performance.

PR #641, opened 2024-06-18
export API to set RefTime of Updater

AdamKorcz:
This PR exports an API to set the `RefTime` of the trusted metadata of an `Updater`. This is useful for testing time-related cases in the client outside of go-tufs own repository.

### DIFF
--- a/metadata/updater/updater.go
+++ b/metadata/updater/updater.go
@@ -662,7 +662,7 @@ func (update *Updater) GetTrustedMetadataSet() trustedmetadata.TrustedMetadata {
 	return *update.trusted
 }
 
-func (update *Updater) SetRefTime(t time.Time) {
+func (update *Updater) UnsafeSetRefTime(t time.Time) {
 	update.trusted.RefTime = t
 }
 
--- a/metadata/updater/updater.go
+++ b/metadata/updater/updater.go
@@ -662,6 +662,9 @@ func (update *Updater) GetTrustedMetadataSet() trustedmetadata.TrustedMetadata {
 	return *update.trusted
 }
 
+// Sets the reference time that the updater uses.
+// This should only be done in tests.
+// UnsafeSetRefTime is useful when testing time-related behavior in go-tuf.
 func (update *Updater) UnsafeSetRefTime(t time.Time) {
 	update.trusted.RefTime = t
 }
--- a/metadata/updater/updater.go
+++ b/metadata/updater/updater.go
@@ -662,6 +662,10 @@ func (update *Updater) GetTrustedMetadataSet() trustedmetadata.TrustedMetadata {
 	return *update.trusted
 }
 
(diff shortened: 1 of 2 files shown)


metadata/updater/updater.go
```diff
@@ -662,6 +662,10 @@ func (update *Updater) GetTrustedMetadataSet() trustedmetadata.TrustedMetadata {
 	return *update.trusted
 }
 
+func (update *Updater) SetRefTime(t time.Time) {
```
kommendorkapten, line 665:
```suggestion
func (update *Updater) UnsafeSetRefTime(t time.Time) {
```

kommendorkapten: We have some other `Unsafe` methods like [this](https://github.com/theupdateframework/go-tuf/blob/41866142c66ab85ba57bc06fe01be1685fc2afd9/metadata/config/config.go#L46)

If would suggest prefixing this with Unsafe too, and add a comment on why this is needed, and the consequences of using it.

kommendorkapten:
Thanks for the contribution @AdamKorcz, would you mind also adding a comment describing this, why it may be useful and any consequences of using this in a real life application?

AdamKorcz:
> Thanks for the contribution @AdamKorcz, would you mind also adding a comment describing this, why it may be useful and any consequences of using this in a real life application?

https://github.com/theupdateframework/go-tuf/pull/641/commits/faf9792df89ebd60c16f13d49d22d6ea13868b1d - LMK if it needs more context. I have kept it short to align the description with other methods in the file.

metadata/updater/updater.go
```diff
@@ -662,6 +662,13 @@ func (update *Updater) GetTrustedMetadataSet() trustedmetadata.TrustedMetadata {
 	return *update.trusted
 }
 
+// Sets the reference time that the updater uses.
+// This should only be done in tests.
+// UnsafeSetRefTime is useful when testing time-related behavior in go-tuf.
```
kommendorkapten, line 667:
```suggestion
// UnsafeSetRefTime sets the reference time that the updater uses.
// This should only be done in tests.
// Using this function is useful when testing time-related behavior in go-tuf.
```

kommendorkapten: Just a minor change, comments in Go should start with the function name.

kommendorkapten (approved):
Thank you!

kommendorkapten:
@rdimitrov @trishankatdatadog any thoughts?

rdimitrov (approved):
> @rdimitrov @trishankatdatadog any thoughts?

I think it makes sense 👍 I don't think it's worth it that much but I was going to suggest that we can make the ReftTime field private and add set/get methods so setting it is more explicit.

Thanks for these changes, @AdamKorcz 🙌

kommendorkapten:
I think it's fine w/out get/set? The other fields are lacking this, and I think it's clear with the current design?

PR #647, opened 2024-07-17
Update MAINTAINERS.md

trishankatdatadog:
Stepping down to make way for others because unfortunately I have not been as active as I would like. It has been a great honour working with you all, and I hope to continue contributing as much as I can. Thank you all very much for your work!


kommendorkapten (approved):
Thank you for all good work @trishankatdatadog ❤️

PR #650, opened 2024-09-30
Update the staging TUF repo in the multi-repo example

rdimitrov:
The following PR:
* Fixes an issue where the 2nd repo from the multi-repo example no longer serves the necessary metadata files, so we default to using the same CDN URL. Even though it's not that explicit as before it still shows the capability of the multi-repo client feature.
* Regenerate the metadata files for the repo serving the `map.json` target.
* Updates the URLs pointing to the repo serving the `map.json` target.


rdimitrov:
Note that the CI will fail for the`multirepo` example until we merge this. The reasoning is the test relies on these files being already hosted. Perhaps an item we can address in the future.

kommendorkapten (approved):

Issue #654, opened 2024-10-10
Make repository.repositoryType public (repository.RepositoryType) 


eikenb:
The library uses `repository.repositoryType` as a way to manage the role metadata and interact with it while using the library. However the type is private and so cannot be used in any abstractions where you'd want to pass it as an argument.

Please consider making this type public so it can be used in abstractions.

Thanks.

rdimitrov:
hey, @eikenb, thanks for reaching out 🙏 

Do you have a specific use case for using this type? I'm asking this because so far it has been used only in the `examples` and we were discussing moving this package there to simplify things. We can make it public of course while doing this.

eikenb:
Moving the package to the examples area would be a great idea.

Right now, when working off the examples to help understand the package, using this type initially looks like a nice way to work with the roles as a group, when in fact it isn't, and that confuses things.

rdimitrov:
Opened a PR to make the move 👍

PR #656, opened 2024-10-30
Move the repository package under examples/repository

rdimitrov:
The following PR moves the repository package under examples/repository. 

The `repository` package is only used in the examples and there are no plans to build a repository functionality for the time being, so it makes more sense to live under the examples folder for now.

Fixes https://github.com/theupdateframework/go-tuf/issues/654


kommendorkapten (approved):
🎉

PR #658, opened 2024-11-20
fix: multirepo potential nil pointer dereference

MrDan4es:
This PR prevent potential panic on NIL pointer when we trying to download file (from user input repo name for example).

### DIFF
--- a/metadata/multirepo/multirepo.go
+++ b/metadata/multirepo/multirepo.go
@@ -311,8 +311,13 @@ func (client *MultiRepoClient) DownloadTarget(repos []string, targetFile *metada
 	log := metadata.GetLogger()
 
 	for _, repoName := range repos {
+		repoClient, ok := client.TUFClients[repoName]
+		if !ok {
+			return "", nil, fmt.Errorf("no client found for repo %q", repoName)
+		}
+
 		// see if the target is already present locally
-		targetPath, targetBytes, err := client.TUFClients[repoName].FindCachedTarget(targetFile, filePath)
+		targetPath, targetBytes, err := repoClient.FindCachedTarget(targetFile, filePath)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


metadata/multirepo/multirepo.go
```diff
@@ -311,6 +311,11 @@ func (client *MultiRepoClient) DownloadTarget(repos []string, targetFile *metada
 	log := metadata.GetLogger()
 
 	for _, repoName := range repos {
+		if _, ok := client.TUFClients[repoName]; !ok {
```
kommendorkapten, line 314: Would you mind storing the client as a variable, and then use it below on line 320 and 330 to avoid referencing the map again, would be slightly more readable.

kommendorkapten:
Thanks for fixing this. Would you mind amend the commit with a signoff?

MrDan4es:
Done

metadata/multirepo/multirepo.go
```diff
@@ -311,8 +311,14 @@ func (client *MultiRepoClient) DownloadTarget(repos []string, targetFile *metada
 	log := metadata.GetLogger()
 
 	for _, repoName := range repos {
+		repoClient, ok := client.TUFClients[repoName]
+		if !ok {
+			log.Error(fmt.Errorf("client not found"), "No client found for this repo", "repo", repoName)
+			continue
```
rdimitrov, line 317: Don't we want to return an error here instead of continuing?

rdimitrov (approved):
This looks nice! 🙌  Thank you 🚀

metadata/multirepo/multirepo.go
rdimitrov: Thank you! 💯

kommendorkapten (approved):
Thanks!

Issue #660, opened 2024-12-04
Hardcoded timeout in updater.DownloadTarget


eikenb:
https://github.com/theupdateframework/go-tuf/blob/990cdb83f6652d4ab430a9a8c9af520d029fd6d9/metadata/updater/updater.go#L248

We're hitting the timeout above when downloading files using the updater and want to raise it. Maybe this should be passed in as a parameter or added to the UpdaterConfig? Our current workaround is a re-implementation of the DownloadTarget() function that lets us set that value.

rdimitrov:
hey, @eikenb 👋 

Thanks for the raising this 👍 So far we haven't had anyone else raise that 15 seconds are not enough to download a target, but I do see how hardcoding it can become an issue (big targets, slow network, etc) 👍 

I've briefly checked the specification and there's no explicit mentioning of it, so it's up to the implementations to set this accordingly. That said I agree it's fair to make this configurable 👍 

We'll add it to the project planning and we'll try to include in the next release. Of course we'll be more than happy to welcome a PR in case you already have something on your side 🙌 That would make things move up much faster.

sureshkrishnan-v:
Hi, I’d like to work on this issue as a first-time contributor—could you confirm if it’s available and share any guidelines?

sureshkrishnan-v:
Ok,  the timeout hardcoded to 15s needs to be configurable else it defaults to 15s am I right? and need to know how can I get time from updaterconfig struct will it be set at initialisation time ?

rdimitrov:
Yes, it should be available through `config.UpdaterConfig` 👍 

But again, I would wait to see in case @eikenb has anything already prepared so we don't duplicate the same work twice 👍

sureshkrishnan-v:
ok i will wait

eikenb:
@rdimitrov @sureshkrishnan-v ... Sorry for the delayed response. I don't have time to write anything for this currently. So no blocker on my end to letting @sureshkrishnan-v work on it. 

Thanks for the consideration and great to hear it will be addressed.

rdimitrov:
Thanks for the reply, @eikenb 🙏 

@sureshkrishnan-v - I'll assign it to you 👍

sureshkrishnan-v:
I believe I have completed the task as specified. Can I proceed with a PR? #660

rdimitrov:
Hey @sureshkrishnan-v, apologies for replying a bit late 👋 

[...]

PR #665, opened 2024-12-17
Hardcoded timeout

sureshkrishnan-v:
"Fix: Made the timeout in the updater.DownloadTarget configurable with a default value of 15s."


rdimitrov:
Hey, added a comment here regarding the approach on how we might fix this - https://github.com/theupdateframework/go-tuf/issues/660#issuecomment-2605080467 😃

kommendorkapten:
As #682 and #684 are merged, I'm closing this.

Issue #667, opened 2025-01-20
updater.go - misuse of os.WriteFile ?


udf2457:
Given you have:

https://github.com/theupdateframework/go-tuf/blob/830edf88770482a58d7741473f553187a528e614/metadata/updater/updater.go#L595

Its not clear why you are doing this:

https://github.com/theupdateframework/go-tuf/blob/830edf88770482a58d7741473f553187a528e614/metadata/updater/updater.go#L601

`_, err = file.Write(data)` would surely be a far more sensible construct ?

I'm not going to open a PR on it, I'll leave that to your own coding preferences ...

rdimitrov:
hey, @udf2457, thanks for reaching out. 

I agree both can be used with the difference that one allows you to specify the file permissions explicitly so it might be the reason why we chose it. 

In any case as you mentioned it's a matter of preference and it doesn't affect performance nor behaviour in any way. The amount of imported packages would be the same too. That said, I think we can close the issue for now unless we see more reasons to change it. Wdyt?

udf2457:
Huh ? 🤷  Let me spell it out ?

You are already opening file in line 595 with `os.CreateTemp`.

You have `defer file.Close()` on the "next" line.

`os.WriteFile` is the "next" line after that.

Ergo you already have an open file handler. 

So ....

If you don't like the default permissions, don't use `CreateTemp` ?

Otherwise ... just use the already open file handler, no ?

rdimitrov:
Hey, @udf2457, thanks for the additional details. Can you open a PR addressing this per your suggestions? Thank you 🙏

udf2457:
Thx for merging @kommendorkapten, @rdimitrov ❤ . Will close this.

PR #669, opened 2025-01-21
updater.go: replace os.WriteFile with file.Write()

udf2457:
Ref https://github.com/theupdateframework/go-tuf/issues/667

I also added an explicit `file.Close()` on the error conditions here because otherwise `defer file.Close()` would be called after the `file.Remove()`, IIRC Windows errors if you attempt to remove an open file.

### DIFF
--- a/metadata/updater/updater.go
+++ b/metadata/updater/updater.go
@@ -597,10 +597,22 @@ func (update *Updater) persistMetadata(roleName string, data []byte) error {
 		return err
 	}
 	defer file.Close()
-	// write the data content to the temporary file
-	err = os.WriteFile(file.Name(), data, 0644)
+	// change the file permissions to our desired permissions
+	err = file.Chmod(0644)
 	if err != nil {
-		// delete the temporary file if there was an error while writing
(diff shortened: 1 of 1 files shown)


rdimitrov (approved):

kommendorkapten:
This looks good, but I'm thinking about the `defer file.Close()` on line 599, we can remove that too right? 
As there are no paths now when there is no `file.Close()`. It's closed explicitly on line 624 before `os.Rename()`.

metadata/updater/updater.go
```diff
@@ -597,10 +597,22 @@ func (update *Updater) persistMetadata(roleName string, data []byte) error {
 		return err
 	}
 	defer file.Close()
```
kommendorkapten, line 599:
```suggestion
```

kommendorkapten (approved):
Thank you!

kommendorkapten:
@rdimitrov would you take an an extra look?

rdimitrov (approved):
Looks good 🚀 I see we can also remove some of the err variables and put the statement in the if condition directly but that's just cosmetics and can be a separate PR.

PR #671, opened 2025-01-22
Remove readFile() and reverseSlice() in favour of stdlib

udf2457:
propose removing `readFile()` from `updater.go` as it appears to essentially be a re-implementation of stdlib's `os.ReadFile`

Also found a "hidden" re-implementation of `os.ReadFile` in metadata.go, so changed that to `os.ReadFile` too.

Also found the re-implementation of `os.Readfile` in `cli/tuf/cmd/root.go`, so changed that to `os.ReadFile` too.

Also found a re-implementation of `os.ReadFile` in `cli/tuf-client/cmd/root.go` where it appears completely unused ?

Also `reverseSlice()` in `metadata/updater/updater.go` seems to be a reimplementation of stdlib's `slices.Reverse()` so removed that too.

Phew... I think that's all of them. 😆

### DIFF
--- a/examples/cli/tuf-client/cmd/init.go
+++ b/examples/cli/tuf-client/cmd/init.go
@@ -81,7 +81,7 @@ func InitializeCmd() error {
 	}
 
 	// read the content of root.json
-	rootBytes, err := ReadFile(rootPath)
+	rootBytes, err := os.ReadFile(rootPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


udf2457:
I see the test failed on `examples/cli/tuf-client/cmd/root.go:21:2: "io" imported and not used`.

Easily fixed, I'll remove the `io` from that file.

udf2457:
Can you try re-running the test, `io` is gone now from `root.go`.

Apologies for missing that !

kommendorkapten (approved):
This is great, thank you!

PR #686, opened 2025-05-08
Set a default HTTP client for DefaultFetcher in DownloadFile method if none is set

malancas:
If someone instantiates a new DefaultFetcher directly with `DefaultFetcher{}` instead of with the new constructor , `DownloadFile` will fail because the `client` field has not been set. This pull request updates `DownloadFile` to check if the `client` field is `nil` and sets a default HTTP client if it is to avoid breaking current usage of `DefaultFetcher`.


rdimitrov (approved):
Nice catch 🚀

kommendorkapten (approved):
Nice work.

PR #698, opened 2025-10-30
Bump Go to 1.24.9

rdimitrov:
Bumps the version of Go to 1.24.9. 

This fixes several security vulnerabilities found in older versions of Go. For reference - https://github.com/theupdateframework/go-tuf/actions/runs/18939463805/job/54074306468?pr=697

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -24,9 +24,9 @@ jobs:
     steps:
       - id: govulncheck
         uses: golang/govulncheck-action@3a32958c2706f7048305d5a2e53633d7e37e97d0
-        continue-on-error: true
         with:
-          go-version-file: 'go.mod'
+          go-version-input: ''
+          go-version-file: go.mod
           go-package: ./...
 
   golangci:


rdimitrov:
Yeah, it seems we hit a requirement that got introduced in the newer crypto packages of Go.

I have temporarily disabled this since it's solely used in our tests (fixtures) but my assumption is that once #695 lands we'll have that sorted out.

**Details about the issue:**
-  Go 1.24 introduced a security enhancement that enforces a minimum RSA key size of 1024 bits
- The test repository's timestamp_key is only 1023 bits (1 bit short!), the other ones are okay (1024)
- When Go 1.24 tries to sign metadata with this undersized key, the crypto/rsa package rejects it and returns an error
- This causes tests to fail with: "unsigned metadata error: Verifying timestamp failed, not enough signatures, got 0, want 1"
- Added GODEBUG=rsa1024min=0 before running the unit tests (.github/workflows/tests.yml & Makefile test target) temporarily disables the 1024-bit minimum check and allows the existing 1023-bit key to work.

PR #717, opened 2026-01-21
Downgrade go mod version

Hayden-IO:
It's best practice to keep this at .0 so that clients don't have to depend on a patch release and instead can build with the latest patch release.


rdimitrov:
> It's best practice to keep this at .0 so that clients don't have to depend on a patch release and instead can build with the latest patch release.

hey, @Hayden-IO, thanks for reaching out! 

Usually we try to follow this but in this case there're a few vulnerabilities that we resolved by bumping to `.5`. 

You can check the govulncheck output from the CI run of your PR - https://github.com/theupdateframework/go-tuf/actions/runs/21230181654/job/61113707818?pr=717

PR #721, opened 2026-01-26
Do not allow empty hashes for the Target role

rdimitrov:


kommendorkapten (approved):

metadata/marshal.go
```diff
@@ -230,6 +230,10 @@ func (signed *TargetFiles) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &s); err != nil {
 		return err
 	}
+	// Per TUF spec, hashes are mandatory for target files
+	if len(s.Hashes) == 0 {
+		return fmt.Errorf("hashes must not be empty for target files")
+	}
```
cyphar, line 236: FYI, putting this check in the marshal path broke our use of this library to construct TUF extensions (we do a pre-flight check that the type supports `UnrecognizedFields` which needs to work with generic `any`s).

We could try to work around it but fundamentally I don't think these kinds of checks make sense in marshalling code -- the data is representable in JSON so validating it in `VerifyLengthHashes` is all that is really necessary IMHO.

rdimitrov: Thanks for flagging this. The spec requires at least one hash per target and Python TUF and Tough (other TUF implementations) also reject empty hashes during decoding so that was the reasoning behind the check.

That said, I see how this breaks your extension handling. Would moving the check from `UnmarshalJSON` to the updater’s metadata validation address your use case?

fyi: @kommendorkapten @jku

cyphar: Yeah that would help IIUC, we don't actually publish the output, it's just done in-memory to check if the type supports `UnrecognizedFields` (which is *incredibly* handy by the way -- thanks for keeping this feature from the Python implementation!).

I can send a PR if you like (and make sure it resolves the issue for us).

rdimitrov: Please, that would be great! Thanks! 🙏

PR #730, opened 2026-05-14
Fix log line to not be `fmt`-styled

abayer:
We've been seeing `Failed to verify %s with key ID %s` a bunch in the logs from a service of ours that's consuming a TUF root, and it's a bit confusing to see what appears to be a `fmt`-style line that isn't actually `fmt`ed. This PR changes that `log.Info` call to work like the one a bit below it for when a key ID is verified for the role.


rdimitrov (approved):
Thanks! 🙌

PR #733, opened 2026-05-19
Fix threshold counting for duplicate public keys

rdimitrov:
## Summary
- Count verified threshold contributions by resolved public key fingerprint instead of key ID.
- Prevent duplicate key records for the same public key from satisfying multi-key thresholds.
- Add regression coverage for ECDSA, Ed25519, and RSA duplicate-public-key cases.

## Testing
- go test ./metadata -run 'TestVerifyDelegate(Duplicate|Threshold|$)' -count=1 -v

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -24,6 +24,7 @@ import (
 	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/sha512"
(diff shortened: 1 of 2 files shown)


kommendorkapten (approved):
Thanks, looks good.

PR #736, opened 2026-06-02
Format expires with second precision per TUF spec

### LINKED ISSUE #596: bug: use the correct format for "expires" (should not include milliseconds)
Moved from https://github.com/rdimitrov/go-tuf-metadata/issues/22

arpitjain099:
## What this does

Fixes the `expires` serialization so it conforms to the TUF spec date/time format.

The four role `MarshalJSON` methods in `metadata/marshal.go` (`RootType`, `SnapshotType`, `TimestampType`, `TargetsType`) wrote the `expires` field as a raw `time.Time`:

```go
dict["expires"] = signed.Expires
```

`encoding/json` serializes a `time.Time` with `RFC3339Nano`, so any sub-second component leaks into the output, for example `2030-01-01T00:00:00.99008Z`. The TUF spec requires a whole-second UTC timestamp of the form `YYYY-MM-DDTHH:MM:SSZ`.

## Changes

- Add a small `formatExpires` helper plus an `expiresFormat` layout constant in `metadata/marshal.go`, and use the helper at all four marshal sites so they agree on a single format.
- Unmarshal is left untouched. It still accepts and round-trips sub-second input (parsing simply truncates to whole seconds on the next marshal).
- Add `metadata/expires_format_test.go`:
  - `TestExpiresMarshalSecondPrecision` marshals each of the four role types built from a `time.Time` that carries sub-second precision and asserts the emitted `expires` matches `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$`.
  - `TestExpiresMarshalUnmarshalRoundTrip` marshals, unmarshals, and confirms the value round-trips and re-marshals stably.
- Update the existing marshal-output golden assertions in `metadata/metadata_test.go` that hard-coded the old sub-second output (`...45.0000001Z`) to the corrected `...45Z`. The sub-second `testRootBytes` literal is kept as parse-direction input so the suite still exercises sub-second tolerance on unmarshal.
- Regenerate the on-disk signed fixtures (`root.json`, `snapshot.json`, `targets.json`, `timestamp.json` under `internal/testutils/repository_data/repository/metadata`) so their `rsassa-pss-sha256` signatures match the corrected second-precision `signed` body. Each fixture changes only two values: the `expires` string and the corresponding signature. The fixtures were re-signed with their existing keystore keys using the repo's own `internal/testutils/rsapss.LoadRSAPSSSignerFromPEMFile` helper, so the scheme is unchanged.

The legacy parse-only fixture `1.root.json` is intentionally left as-is: it is consumed by `TestFromFile`/`TestFromBytes` to assert sub-second parse tolerance, is never re-serialized or signature-verified in tests, and is signed by a key whose private half is not in the keystore.

## Testing

```
go test ./metadata/...
gofmt -l metadata/marshal.go metadata/metadata_test.go metadata/expires_format_test.go   # no output
go vet ./metadata/...                                                                      # clean
```

`go vet` and `gofmt` are clean. The new tests fail on the unpatched code (all four roles emit sub-second precision) and pass after the fix. `TestCompareFromBytesFromFileToBytes`, `TestToFromBytes`, `TestKeyVerifyFailures`, and `TestMetadataVerifyDelegate` continue to pass with the regenerated fixtures.

Fixes #596


arpitjain099:
Rebased onto master - the only conflict was the import block (master added `maps`, this PR added `time`), now both are present. The expires-format tests pass and the package vets clean.

arpitjain099:
Rebased onto master and, in doing so, found a problem with this branch that the old fixtures were hiding. CI will be red, deliberately.

Master regenerated the test data in #757. Rather than re-applying my `expires` reformat to the new fixtures, I took them as-is, and `metadata/trustedmetadata` now fails:

```text
unsigned metadata error: Verifying root failed, not enough signatures, got 0, want 1
```

The cause is this branch, not the merge. Verification re-marshals the parsed `signed` object to recover the canonical bytes. Doing the second-precision truncation inside `MarshalJSON` means metadata that was signed with a sub-second `expires` (master's fixtures still carry `2030-08-15T14:30:45.0000001Z`) re-marshals to different bytes than were signed, so every signature over it fails.

My original branch dodged this by regenerating the fixtures, which masked the incompatibility rather than fixing it. Any real repository with sub-second `expires` in already-signed metadata would break on upgrade.

Confirmed by testing: current master alone passes `go test ./metadata/trustedmetadata/...`; master plus this branch's `marshal.go` fails as above.

Two directions that avoid it, and I would rather have your steer than pick one:

1. Apply the format only where metadata is created or its expiry set, leaving `MarshalJSON` byte-preserving for parsed metadata.
2. Preserve the original `expires` string on unmarshal and emit it verbatim on re-marshal, applying the spec format only to newly set values.

Happy to implement either. I did not want to push one silently given it changes what the PR is.

arpitjain099:
Went ahead and implemented option 1 in 059fdff, since it is the smaller change and keeps `MarshalJSON` honest. CI should be green again.

The truncation now happens in `Root`, `Snapshot`, `Targets` and `Timestamp`, where the expiry is first set:

```go
func truncateExpires(t time.Time) time.Time {
	return t.UTC().Truncate(time.Second)
}
```

`MarshalJSON` is byte-preserving again. So newly constructed metadata gets the spec-required whole-second UTC form, and metadata parsed from bytes or from disk is written back exactly as it arrived, which is what verification depends on.

Tests cover both halves. `TestConstructorsTruncateExpiresToSeconds` and `TestConstructorsNormalizeExpiresToUTC` for the construction path, and `TestParsedExpiresRoundTripsUnchanged` as a guard that fails if anyone moves the formatting back into marshaling.

Two existing tests moved for good reasons rather than to make things pass. `TestToByte` builds its root through `Root(...)`, so its expected output is now whole seconds. `TestToFile` parses `testRootBytes` and writes them back, so its expected output keeps `.0000001Z`, which is the behaviour this whole change is protecting.

Full suite is green, `metadata/trustedmetadata` included, and `go vet ./metadata/...` is clean.

PR #740, opened 2026-06-12
Replace copyMapValues with maps.Copy from stdlib

udf2457:
maps.Copy is in stdlib since go1.21, removing the need for `func copyMapValues(src, dst map[string]any)`


kommendorkapten (commented):
Thanks!

kommendorkapten (approved):

PR #743, opened 2026-06-19
Nil-pointer panic: mapping can reference repos not in Repositories

udf2457:
A malformed or adversarially crafted map file could trigger a nil-pointer dereference.

If a mapping entry references a repo name that isn't in the `Repositories` map, `client.TUFClients[repoName]` returns `nil`, and the `.GetTargetInfo()` call would panic.

https://github.com/theupdateframework/go-tuf/blob/4b704cdac869b71e2fd67da76aaa09579b1df137/metadata/multirepo/multirepo.go#L257-L259


kommendorkapten (dismissed):
Thanks

rdimitrov (approved):

PR #744, opened 2026-06-19
Refactor delegator metadata check

udf2457:
If my understanding is correct `trusted.Root` is effectively always non-nil at this point because it is set unconditionally in `loadTrustedRoot` during construction.

So this is a suggestion to improve readability.


rdimitrov:
I think I agree with Copilot's comment here, we are not gaining too much in simplifying this, also I rather keep that part explicit for now so I'd close this one. Nevertheless thanks for raising it! 🙏

PR #749, opened 2026-06-19
Nil dereference on JSON null

udf2457:
if `json.Unmarshal` of the JSON value `null` into a pointer succeeds with a `nil`, `range mapFile.Repositories` would panic.


kommendorkapten (approved):
Thanks!

rdimitrov (approved):

Issue #747, opened 2026-06-19
Potential fragility in path-pattern matching


udf2457:
https://github.com/theupdateframework/go-tuf/blob/4b704cdac869b71e2fd67da76aaa09579b1df137/metadata/metadata.go#L585-L587

https://github.com/theupdateframework/go-tuf/blob/4b704cdac869b71e2fd67da76aaa09579b1df137/metadata/multirepo/multirepo.go#L248-L251

In theory there is some fragility in the above because `filepath.Match` is tied to the OS.  

In general it should have negligible impact, but you might hit an edge case on Windows with `\`.  So you might want to consider options for how you want to handle that ? Might need to check the TUF spec.

In addition `err` is discarded in the `metadata.go` version which means you have no way to distinguish between "no match" and "invalid pattern".  Surely an invalid pattern should error ?

PR #753, opened 2026-07-08
Surface invalid delegation path patterns and match OS-independently

arpitjain099:
As a supply-chain-security researcher I was looking at delegation path matching and hit the fragility flagged in #747. `isTargetInPathPattern` was throwing away the error from `filepath.Match`, so a malformed pattern like an unterminated character class (`targets/[`) got silently treated as a plain non-match, and a caller couldn't tell "this pattern doesn't match" apart from "this pattern is invalid". It also used `filepath.Match`, whose separator handling depends on the host OS, even though TUF target paths are always `/`-separated.

This switches to `path.Match` (always `/`-separated, `\` as a glob escape) and propagates the error up through `IsDelegatedPath`, which already returns `(bool, error)`. I kept the change scoped to `metadata.go`; the sibling `multirepo.go` code already captures and returns that error.

I verified it by adding a regression test: on the current code the malformed-pattern case returns `err == nil`, and after the fix it returns `path.ErrBadPattern`, while the existing spec-derived path cases still pass (`go test ./metadata -run TestIsDelegatedPath -count=1`). Thanks for taking a look.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -34,7 +34,7 @@ import (
 	"io"
 	"math"
 	"os"
-	"path/filepath"
+	"path"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


kommendorkapten:
Thank you for the PR. Maintainer availability is lower now during summer time, but a first skim this looks solid, but need to test this out more first, so will review may drag a few weeks.

kommendorkapten (dismissed):

kommendorkapten (approved):
This looks good, thanks!

PR #757, opened 2026-07-28
Improve performance of VerifyDelegate, regenerate test data

Hayden-IO:
When verifying metadata with multiple keys, VerifyDelegate would unnecessarily canonicalize the same JSON data within the loop for every key. This moves canonicalization to before signature validation for a performance improvement.

Running a local benchmark with 5000 target files and 20 signing keys used to sign the metadata, the benchmark showed around a 14x performance improvement.

go test -bench=BenchmarkVerifyDelegate ./metadata
BenchmarkVerifyDelegate-4          4    254893688 ns/op (254 ms / op)

go test -bench=BenchmarkVerifyDelegate ./metadata
BenchmarkVerifyDelegate-4         60     18850549 ns/op (18 ms / op)

This also adds a check that there actually is a signature to verify, skipping calling VerifySignature when it would have always failed.

When testing using Go 1.26.x, I also found that the tests would fail because of 1024 bit RSA keys. Regenerated all test data with RSA-2048.


Hayden-IO:
Benchmark script:

```
package metadata

import (
	"crypto"
	"crypto/ed25519"
	"fmt"
	"path/filepath"
	"testing"

	"github.com/sigstore/sigstore/pkg/signature"
	"github.com/theupdateframework/go-tuf/v2/internal/testutils"
)

// BenchmarkVerifyDelegate measures the performance of verifying a delegated role.
// It constructs a scenario with a large targets file and multiple valid signatures
// to amplify the impact of any redundant JSON serialization during verification.
func BenchmarkVerifyDelegate(b *testing.B) {
	repoPath := "../internal/testutils/repository_data/repository/metadata"
	targetsPath := "../internal/testutils/repository_data/repository/targets"
	keystorePath := "../internal/testutils/repository_data/keystore"
	err := testutils.SetupTestDirs(repoPath, targetsPath, keystorePath)
	if err != nil {
		b.Fatalf("failed to setup test dirs: %v", err)
	}
	defer testutils.Cleanup()

	root, err := Root().FromFile(filepath.Join(testutils.RepoDir, "root.json"))
	if err != nil {
		b.Fatalf("failed to load root metadata: %v", err)
	}

	targets, err := Targets().FromFile(filepath.Join(testutils.RepoDir, "targets.json"))
	if err != nil {
		b.Fatalf("failed to load targets metadata: %v", err)
	}

	// Make the targets file huge so EncodeCanonical is very slow.
	// This simulates a large repository where canonical JSON encoding is expensive.
	for i := 0; i < 5000; i++ {
		targets.Signed.Targets[fmt.Sprintf("dummy_target_%d", i)] = &TargetFiles{
			Length: 100,
			Hashes: Hashes{"sha256": []byte("dummy")},
		}
	}
	targets.Signatures = nil // Clear previous signatures

	// Make the TARGETS role have many keys and corresponding signatures
	role := root.Signed.Roles[TARGETS]
	role.KeyIDs = nil // Clear previous keys
	
	// Add 20 valid keys and signatures to force the verifier loop to execute
	// 20 successful verification iterations.
	for i := 0; i < 20; i++ {
		pub, priv, _ := ed25519.GenerateKey(nil)
		key, _ := KeyFromPublicKey(pub)
		id, _ := key.ID()
		root.Signed.Keys[id] = key
		role.KeyIDs = append(role.KeyIDs, id)
		
		signer, _ := signature.LoadSigner(priv, crypto.Hash(0))
		sig, _ := targets.Sign(signer)
		targets.Signatures = append(targets.Signatures, *sig)
	}
	// Require all 20 signatures to be verified
	role.Threshold = 20

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		err = root.VerifyDelegate(TARGETS, targets)
		if err != nil {
			b.Fatalf("VerifyDelegate failed: %v", err)
		}
	}
}
```

kommendorkapten:
Thanks! The test keys with < 1024 bits has been annoying, appreciated you fix that. Would you mind also removing the `rsa2014min=0` in https://github.com/theupdateframework/go-tuf/blob/913b64be42750e068314b2fe9e02766177b7d6d7/.github/workflows/tests.yml#L44 as that's no longer needed.

Hayden-IO:
Removed it in the workflow and in the Makefile

kommendorkapten (approved):

PR #758, opened 2026-07-28
Add support for ML-DSA (TAP 21)

Hayden-IO:
This commit introduces support for ML-DSA (FIPS 204) post-quantum
cryptographic signatures, as defined in TAP 21.

For now, the implementation uses filippo.io/mldsa for signing and
verification and filippo.io/mldsa/x509 for key parsing. These
implementations match what is currently in an internal package in the
std library. Once Go 1.27 lands, we'll be able to directly use
crypto/mldsa and crypto/x509. Additionally sigstore/sigstore will be
updated so that we can use the SignerVerifier interface.

Signed-off-by: Hayden <8418760+Hayden-IO@users.noreply.github.com>


Hayden-IO:
Currently rebased on https://github.com/theupdateframework/go-tuf/pull/757.

Hayden-IO:
If maintainers would prefer to not merge this until Go 1.27, that should be coming hopefully some time in August, at which point we'll merge https://github.com/sigstore/sigstore/pull/2353. This will mean less conditionals as we can reuse the cryptoutils interfaces. Note we'll still need a conditional to handle the different payload to sign as defined in TAP 21.

Hayden-IO:
Oh, and TAP 21 is not yet merged, though I pinged the PR - https://github.com/theupdateframework/taps/pull/195/

JustinCappos:
FYI: this should move quickly through the TAP process.  We're mostly
waiting for operational experience to ensure the details are all worked
out.  I would be looking for you all to give a thumbs up before we
approve.

I support merging the PR in and having the TAP status as a draft in the
repo now, if this is helpful.

jku:
I ran the go example cli against https://jku.github.io/pqc-test-repo2/metadata/: The metadata update is successful so it looks like python and go implementations agree on this.

metadata/metadata.go
```diff
@@ -231,6 +228,35 @@ func (meta *Metadata[T]) Sign(signer signature.Signer) (*Signature, error) {
...
+		prefix := append([]byte("tuf"), version)
+		payloadToSign = append(prefix, h.Sum(nil)...)
+	}
+
+	// sign the Signed part
+	var sb []byte
+	cryptoSigner, isCryptoSigner := signer.(crypto.Signer)
+	_, isMLDSA := publ.(*mldsa.PublicKey)
+
+	if isCryptoSigner && isMLDSA {
```
jku, line 250: what's the implication of checking `isCryptoSigner` -- is that just to keep a linter happy?

jku (approved):
seems correct to me.

I assume after https://github.com/sigstore/sigstore/pull/2353 this will become a bit simpler, right? The custom prefixing scheme still needs to be handled here but otherwise the special casing will go away.

Hayden-IO:
> seems correct to me.
> 
> I assume after https://github.com/sigstore/sigstore/pull/2353 this will become a bit simpler, right? The custom prefixing scheme still needs to be handled here but otherwise the special casing will go away.

Yea, along with Go 1.27, it'll simplify many of the conditionals, as we can reuse crypto/x509 and the signer interface from sigstore/sigstore.

We are likely a few weeks away from 1.27, so I'm fine either way with merging sooner or waiting.

metadata/metadata.go
Hayden-IO: The `signature.Signer` interface does not guarantee implementation of `crypto.Signer`. `kms.SignerVerifier` does however, as there is a `CryptoSigner` method, and so all `signature.Signer` implementations in sigstore/sigstore implement the `crypto.Signer` interface.

This is clunky, and it's awkward for the integrating application that signs as they would need to provide an ML-DSA implementation of `signature.Signer` and `crypto.Signer`, even though the latter isn't known until runtime.

Maybe this is the motivation to just wait until we land the sigstore/sigstore PR? Then you'd just call `signer.SignMessage`.

Hayden-IO: You'll see in the tests exactly this, a minimal implementation of an ML-DSA signer that implements both `signature.Signer` and `crypto.Signer`.

Orthogonal, but I'll mention that with ML-DSA and Ed25519, the need for the `signature.Signer` interface is less obvious because you always pass the message-to-sign. However, so that we have a unified interface between KMS and Go std lib signers, there's still value in it.

mnm678 (approved):

Hayden-IO:
https://github.com/sigstore/sigstore/pull/2353 will land shortly, so I'd suggest we hold off on merging.

jku:
I've updated this in #780

kommendorkapten:
Closing in favour of #780

PR #781, opened 2026-09-15
Treat metadata that does not deserialize as a repository error

arpitjain099:
A cached metadata file that does not parse stops the client updating, permanently, and deleting the same file fixes it.

`loadTimestamp`, `loadSnapshot` and `loadTargets` each read the local copy first and fall back to the remote when it turns out to be no good:

```go
if errors.Is(err, &metadata.ErrRepository{}) {
    // local timestamp is not valid, proceed downloading from remote; note that this error type includes several other subset errors
    log.Info("Local timestamp is not valid")
} else {
    // another error
    return err
}
```

The comment describes the intent, but the test does not cover the two ways a file can fail to be read at all. `fromBytes` returns a raw `*json.SyntaxError` when the JSON does not parse, and `checkType` returns an `*ErrValue` when `signed._type` is not the role being loaded. `ErrValue.Is` matches only `*ErrValue`, so both land in the `else` and are returned.

Driving this through the real `Refresh()` against the repository simulator, after one successful refresh, then damaging one cached file:

```
role=timestamp corruption=truncate  refresh err=unexpected end of JSON input
role=timestamp corruption=wrongtype refresh err=value error: expected metadata type timestamp, got - bogus
role=timestamp corruption=flipsig   refresh err=<nil>
role=snapshot  corruption=truncate  refresh err=unexpected end of JSON input
role=snapshot  corruption=wrongtype refresh err=value error: expected metadata type snapshot, got - bogus
role=snapshot  corruption=flipsig   refresh err=<nil>
role=targets   corruption=truncate  refresh err=unexpected end of JSON input
role=targets   corruption=wrongtype refresh err=value error: expected metadata type targets, got - bogus
role=targets   corruption=flipsig   refresh err=<nil>
```

The third line of each group is the part that makes this look unintended. A cached file whose signature does not verify recovers, because `ErrUnsignedMetadata` is an `ErrRepository`. So the adversarial case is handled and the benign one, a half-written file from a crash or a full disk, is not. Deleting the file also recovers, which means the repair is already available and the client simply will not perform it.

### The change

`ErrDeserialization`, returned by `checkType` and by the `json.Unmarshal` in `fromBytes`, covering exactly the two "cannot be read" cases. It is a subset of both `ErrRepository` and `ErrValue`, using the same double-subset form `ErrEqualVersionNumber` already uses for `ErrRepository` and `ErrBadVersionNumber`. `ErrRepository` is what makes the local fallback fire; `ErrValue` is what this path returned before, so callers and tests that check for one are unaffected.

Two things in the tree already say this is the intended classification:

- python-tuf, which this client is ported from, defines `DeserializationError` as a subclass of `RepositoryError`.
- `trustedmetadata.go:74`, `:121`, `:209`, `:309` and `:349` each return an `ErrRepository` when `Signed.Type` is wrong. All five are unreachable, because `checkType` has already rejected the same input with an `ErrValue` one layer down. The file disagrees with itself about what a type mismatch is, and the reachable half is the one that breaks the fallback.

### Scope

`errors.Is(err, &metadata.ErrRepository{})` appears in exactly three places in the tree, all of them the local-cache load. Metadata fetched from the remote is checked separately and those errors are still returned to the caller, so malformed remote metadata fails as before and nothing about verification changes.

### Tests

`updater_local_cache_test.go` runs a successful refresh, damages the cached timestamp, snapshot or targets file four ways (truncated, not JSON, wrong `_type`, empty), refreshes again, and asserts both that it succeeds and that the damaged file was overwritten with the original bytes. Twelve of the twelve fail on master with the errors above.

The unsigned case is kept as a separate control, since it passes either way: if a future change breaks the fallback as a whole rather than just the deserialization half, the two tests fail differently.

`go test ./...` is clean. `go vet` is clean. `gofmt -l` reports `metadata/metadata_test.go`, which it also reports on unmodified master.


metadata/updater/updater_local_cache_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2024 The Update Framework Authors
```
kommendorkapten, line 1:
```suggestion
// Copyright 2026 The Update Framework Authors
```

kommendorkapten:
Thanks, this looks good. One comment on the year for the copyright notice.

PR #783, opened 2026-09-22
metadata: correctly match `path_hash_prefixes` delegations

cyphar:
The specification clearly states that `path_hash_prefixes` are meant to
be interpreted as prefixes of the *hexadecimal representation* of the
SHA-256 hash of the target path.

However, go-tuf/v2 has never handled this correctly. Commit e0d9e3e2d67d
("feat: add support for succinct roles and hashbin delegations")
incorrectly tried to match against the raw byte values while commit
ea170bfe8108 ("Metadata tests followup (#54)") switched to base64
instead of hexadecimal (base16).

In both cases this meant that go-tuf/v2 would be unable to fetch the
target file that the repository publisher intended if they used
`path_hash_prefixes` (in most cases you would get a 404, in principle you
could have the wrong delegation provide the target). In practice, this
is not really possible for a malicious actor to exploit as SHA-256
hashes are (by design) unpredictable and `path_hash_prefixes` is
primarily intended as a fan-out mechanism rather than a trust boundary,
and so this is more a correctness fix than a security fix.

Fixes: ea170bfe8108 ("Metadata tests followup (#54)")
Fixes: e0d9e3e2d67d ("feat: add support for succinct roles and hashbin delegations")
Signed-off-by: Aleksa Sarai <aleksa@amutable.com>


kommendorkapten (approved):
Thanks!

PR #785, opened 2026-09-22
metadata: move number-of-hashes validation to trustedmetadata

cyphar:
This came from the discussion in https://github.com/theupdateframework/go-tuf/pull/721#discussion_r3986095027.

- - -

Commit 17b4808cf953 ("Do not allow empty hashes for the Target role
(#721)") added validation logic for the number of hashes in TargetFiles
to UnmarshalJSON. This broke use-cases where third-party extension code
wanted to be able to validate that UnrecognizedFields round-trips
properly.

The key property that is relevant from the perspective of the TUF spec
is that we do not permit TUF metadata with no hashes for a TargetFiles
structure to be used for updates -- this means this kind of check
belongs more in trustedmetadata than in the (un)marshal code.

Fixes: 17b4808cf953 ("Do not allow empty hashes for the Target role (#721)")
Signed-off-by: Aleksa Sarai <aleksa@amutable.com>


cyphar:
@kommendorkapten Hopefully I understood what you meant in #721? We can also make the error exit-on-first-hit instead of using `errors.Join`.

cyphar:
Ah, when testing this it seems the new `rejectNilMapValues` (added in 923b304b67ac8 #743) is now also problematic... :/

